PR #8, opened 2021-09-16
make div similar to other interfaces

mikelodder7:
Div methods were taking a copy instead of a reference where the other arithmetic methods were taking a reference. This makes them identifical.


tarcieri:
This is a breaking change to an API I already released, although in this case I think it's probably safe to yank that release and make the change as there's unlikely to be anyone using it considering it was just released 2 days ago.

PR #10, opened 2021-09-16
Use `Inner::leading_zeros` to compute `UInt::bits`

tarcieri:
Replaces a while loop with a bit-counting intrinsic


tarcieri:
cc @mikelodder7

mikelodder7 (approved):
LGTM

PR #12, opened 2021-09-17
Implement `ArrayDecoding`

daxpedda:
As discussed on Zulip. Any improvements on the documentation are welcome.

### DIFF
--- a/src/uint/array.rs
+++ b/src/uint/array.rs
@@ -40,7 +40,7 @@ macro_rules! impl_uint_array_encoding {
             impl ArrayDecoding for GenericArray<u8, $bytes> {
                 type Output = super::$uint;
 
-                fn to_bigint(self) -> Self::Output {
+                fn into_bigint(self) -> Self::Output {
                     Self::Output::from_le_byte_array(self)
                 }
             }
@@ -162,7 +162,7 @@ mod tests {
     #[cfg(target_pointer_width = "64")]
     fn to_bigint() {
         let expected_bytes = ByteArray::from(hex!("ffeeddccbbaa99887766554433221100"));
-        let actual_bytes = expected_bytes.to_bigint().to_le_byte_array();
+        let actual_bytes = expected_bytes.into_bigint().to_le_byte_array();
         assert_eq!(expected_bytes, actual_bytes);
     }
 }
--- a/src/uint/array.rs
+++ b/src/uint/array.rs
@@ -154,7 +154,7 @@ mod tests {
     #[cfg(target_pointer_width = "32")]
     fn to_bigint() {
         let expected_bytes = ByteArray::from(hex!("7766554433221100"));
-        let actual_bytes = expected_bytes.to_bigint().to_le_byte_array();
+        let actual_bytes = expected_bytes.into_bigint().to_le_byte_array();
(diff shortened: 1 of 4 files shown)


tarcieri:
On second thought, I'm not sure this is needed because I think what you're trying to do here can already be done by the existing `ArrayEncoding` trait.

You can write code that's generic with respect to the size of the input array like this:

```rust
use crypto_bigint::{ArrayEncoding, ByteArray,UInt, nlimbs};

fn parse_uint<const LIMBS: usize>(bytes: ByteArray<UInt<LIMBS>>) -> UInt<LIMBS>
where
    UInt<LIMBS>: ArrayEncoding
{
    UInt::from_be_byte_array(bytes)
}
```

tarcieri:
Okay, I see the problem though. It can't infer `LIMBS` from the size of the `GenericArray`, which is what you're trying to do.

This looks OK then but I'll leave some line notes.

src/uint/array.rs
```diff
@@ -35,6 +35,15 @@ macro_rules! impl_uint_array_encoding {
...
                 }
             }
+
+            #[cfg_attr(docsrs, doc(cfg(feature = "generic-array")))]
+            impl ArrayDecoding for GenericArray<u8, $bytes> {
+                type Output = super::$uint;
+
+                fn to_bigint(self) -> Self::Output {
+                    Self::Output::from_le_byte_array(self)
+                }
```
tarcieri: This only supports little endian. Can you add two methods and cover both cases, e.g. `to_bigint_be` and `to_bigint_le` maybe?

daxpedda: The trait is only supposed to be implemented for `GenericArray`, the conversion can be done either way, the result would be the same, I'm not sure what this would accomplish.

tarcieri: In its current form the trait only supports a little endian conversion...

daxpedda: Maybe I'm not understanding something here:
```rust
impl ArrayDecoding for GenericArray<u8, typenum::U32> {
    type Output = U384;

    fn into_bigint_le(self) -> Self::Output {
        Self::Output::from_le_byte_array(self)
    }

    fn into_bigint_be(self) -> Self::Output {
        // `GenericArray::as_be_slice` doesn't exist, I understand that's beside the point, but what should I do here?
        Self::Output::from_be_byte_array(self.as_be_slice())
    }
}
```
The outcome here would be exactly the same I assumed.

tarcieri: The methods just describe how the data in the `GenericArray` is to be interpreted. No additional conversion is required.

```rust
impl ArrayDecoding for GenericArray<u8, typenum::U32> {
    type Output = U384;

    fn into_bigint_le(self) -> Self::Output {
        Self::Output::from_le_byte_array(self)
    }

    fn into_bigint_be(self) -> Self::Output {
        Self::Output::from_be_byte_array(self)
    }
}
```

daxpedda: Ah ... right, I lost all context apparently :laughing:

daxpedda: Done.

tarcieri:
Thank you!

PR #14, opened 2021-09-21
Implement `Div` & `Rem` for `UInt`  with `NonZero`

daxpedda:
As discussed on Zulip.

### DIFF
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -279,8 +279,8 @@ where
     type Output = UInt<LIMBS>;
 
     fn rem(self, rhs: NonZero<UInt<LIMBS>>) -> Self::Output {
-        let (q, _, _) = self.ct_div_rem(&rhs);
-        q
+        let (r, _) = self.ct_reduce(&rhs);
+        r
     }
 }
 
@@ -289,8 +289,8 @@ where
     UInt<LIMBS>: Integer,
 {
     fn rem_assign(&mut self, rhs: &NonZero<UInt<LIMBS>>) {
-        let (q, _, _) = self.ct_div_rem(rhs);
-        *self = q
+        let (r, _) = self.ct_reduce(&rhs);
+        *self = r
     }
 }
 
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -2,7 +2,7 @@
 
 use super::UInt;
 use crate::limb::{Inner, SignedInner, BIT_SIZE};
-use crate::{Limb, Wrapping};
+use crate::{Integer, Limb, NonZero, Wrapping};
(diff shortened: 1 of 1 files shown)


src/uint/div.rs
```diff
@@ -131,6 +131,70 @@ impl<const LIMBS: usize> UInt<LIMBS> {
...
+
+impl<const LIMBS: usize> Div<NonZero<UInt<LIMBS>>> for UInt<LIMBS>
+where
+    UInt<LIMBS>: Integer,
+{
+    type Output = UInt<LIMBS>;
+
+    fn div(self, rhs: NonZero<UInt<LIMBS>>) -> Self::Output {
+        let (q, _, _) = self.ct_div_rem(&rhs);
+        q
```
daxpedda, line 175: Is there some more efficient way if we know `rhs` isn't zero?

mikelodder7: Unfortunately not

src/uint/div.rs
```diff
@@ -175,6 +239,70 @@ impl<const LIMBS: usize> DivAssign for Wrapping<UInt<LIMBS>> {
...
+}
+
+impl<const LIMBS: usize> Rem<NonZero<UInt<LIMBS>>> for UInt<LIMBS>
+where
+    UInt<LIMBS>: Integer,
+{
+    type Output = UInt<LIMBS>;
+
+    fn rem(self, rhs: NonZero<UInt<LIMBS>>) -> Self::Output {
+        let (q, _, _) = self.ct_div_rem(&rhs);
```
mikelodder7, line 282:
```suggestion
        let (r, _) = self.reduce(&rhs);
```

mikelodder7: You’re returning the quotient instead of the remainder. Reduce is more efficient if you just want the remainder

src/uint/div.rs
```diff
@@ -175,6 +239,70 @@ impl<const LIMBS: usize> DivAssign for Wrapping<UInt<LIMBS>> {
...
+        q
+    }
+}
+
+impl<const LIMBS: usize> RemAssign<&NonZero<UInt<LIMBS>>> for UInt<LIMBS>
+where
+    UInt<LIMBS>: Integer,
+{
+    fn rem_assign(&mut self, rhs: &NonZero<UInt<LIMBS>>) {
+        let (q, _, _) = self.ct_div_rem(rhs);
```
mikelodder7, line 292: Same as above

src/uint/div.rs
daxpedda: Right ... thanks.

mikelodder7 (approved):

PR #16, opened 2021-09-21
Add conversion `core::num::NonZeroUx`

daxpedda:
Adds convenience functions to convert `core::num::NonZeroUx` to `NonZero<UInt<LIMBS>>`.

### DIFF
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -206,3 +206,48 @@ where
         fmt::UpperHex::fmt(&self.0, f)
     }
 }
+
+impl<const LIMBS: usize> From<NonZeroU8> for NonZero<UInt<LIMBS>>
+where
+    UInt<LIMBS>: Integer,
+{
+    fn from(integer: NonZeroU8) -> Self {
+        Self::from_u8(integer)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


src/non_zero.rs
```diff
@@ -64,6 +68,41 @@ where
     }
 }
 
+impl<const LIMBS: usize> NonZero<UInt<LIMBS>>
+where
+    UInt<LIMBS>: Integer,
+{
+    /// Create a [`NonZero<UInt>`] from a [`NonZeroU8`] (const-friendly)
+    // TODO(tarcieri): replace with `const impl From<NonZeroU8>` when stable
```
tarcieri, line 76: For now you can add non-`const` impls of these as well

tarcieri (approved):
These look good in isolation, although it'd be good to have the `From` impls too

src/non_zero.rs
daxpedda: I left the TODOs because of `const`. Should I remove them?
Done otherwise.

tarcieri: TODOs are good. Eventually this can be unified as a `const impl`.

tarcieri:
Thanks!

PR #17, opened 2021-09-21
Fix build without `rand` crate feature

daxpedda:
The recent `NonZero` addition broke builds with `default-features = false`. I will add CI checks too.


tarcieri:
Thanks!

PR #18, opened 2021-09-21
Add traits to `ArrayDecoding`

daxpedda:
I could add more traits if this makes sense. Improves situations like https://github.com/RustCrypto/password-hashes/pull/232/files#r712984411.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -1,7 +1,7 @@
 //! Interop support for `generic-array`
 
-use crate::{Encoding, Integer, NonZero};
-use core::ops::{Add, Rem};
+use crate::{Encoding, Integer};
+use core::ops::Add;
 use generic_array::{typenum::Unsigned, ArrayLength, GenericArray};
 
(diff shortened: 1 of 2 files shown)


src/array.rs
```diff
@@ -31,7 +31,7 @@ pub trait ArrayEncoding: Encoding {
 #[cfg_attr(docsrs, doc(cfg(feature = "generic-array")))]
 pub trait ArrayDecoding {
     /// Big integer which decodes a `GenericArray`.
-    type Output: Integer;
+    type Output: ArrayEncoding + Integer + Rem<NonZero<Self::Output>, Output = Self::Output>;
```
tarcieri, line 34: Bounding on `ArrayEncoding` is fine, but something like `Rem<NonZero<Self::Output>, Output = Self::Output>` should probably get captured at a higher level than here.

daxpedda: Like `Integer`?

PR #21, opened 2021-09-21
Merge `proptests` crate into parent crate

tarcieri:
Splitting the proptests out into a separate crate was largely a workaround for the lack of `resolver = "2"`.

Now that this crate is in its own repo and we have switched to `resolver = "2"` we can use `proptest` directly.

This makes them part of the normal test workflow and thus hopefully provides better testing ergonomics.


tarcieri:
Seems I never actually set `resolver = "2"`. Need to do that first.

PR #25, opened 2021-10-15
feat: Bitwise And/Or operations

jasalper:


src/uint/and.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub const fn wrapping_and(&self, rhs: &Self) -> Self {
+        self.bitand(rhs)
+    }
+
+    /// Perform checked addition, returning a [`CtOption`] which `is_some` only
+    /// if the operation did not overflow.
+    pub fn checked_and(&self, rhs: &Self) -> CtOption<Self> {
+        let result = self.bitand(rhs);
+        CtOption::new(result, Choice::from(1))
+    }
```
tarcieri, line 35: We've generally followed the stdlib's pattern of impling these cases that don't matter anyway, although admittedly I find it a bit pointless and confusing.

I guess it's worth sticking with for consistency unless we decided to remove everything.

tarcieri (approved):
This looks fine other than the clippy errors which feel like a false positive to me.

You might also consider adding [proptests](https://github.com/RustCrypto/crypto-bigint/blob/master/tests/proptests.rs).

PR #27, opened 2021-10-17
Add bitwise xor and not

kaidokert:
Mostly a copy from @jperaire and/or implementation in https://github.com/RustCrypto/crypto-bigint/pull/25


tarcieri:
Needs rustfmt.

It seems I didn't notice the module names are inconsistent between `limb` and `uint` in #25. Anyway, I'd suggest making them consistent with the `bit*` prefix added in both cases. If you could do `and` and `or` as well for `uint` that'd be great.

Proptests would also be nice if you're interested in adding them. Otherwise I can merge without them and add them later.

kaidokert:
I think i covered most of the feedback, but i can't find `core::ops::Not` implementation in `num_bigint::BigUint`, unsure as to how to write a clean proptest for it.

tarcieri:
Thanks!

PR #36, opened 2021-11-13
Fix `random_mod` performance for small moduli; `NonZero` moduli

### LINKED ISSUE #3: `random_mod` can take a very long time
https://github.com/RustCrypto/utils/blob/710abc8c83a13def1c01958909405976f915e5a5/crypto-bigint/src/uint/rand.rs#L34-L42

If `modulus` is small compared to the full size ("number of limbs * size of limb") then it can take "forever".

The correct method is to mask off the top bits of the random number to match the size of `modulus`. This leads to a worst case of a 50% chance of failing and needing to generate another random number. The average for this worst case is needing to generate 2 random numbers.

tarcieri:
Changes the modulus for `random_mod` to be `NonZero`.

Changes the algorithm used by `random_mod` to generate a number that can be represented by the same number of bytes as the modulus.

Such a number can still be larger than the modulus, but is much more likely not to overflow than a "full-width" number provided the modulus is small relative to the width.

Closes #3 cc @Sc00bz

### DIFF
--- a/src/limb/rand.rs
+++ b/src/limb/rand.rs
@@ -1,20 +1,45 @@
 //! Random number generator support
 
-use super::Limb;
+use super::{Limb, BYTE_SIZE};
+use crate::{Encoding, NonZero, Random, RandomMod};
 use rand_core::{CryptoRng, RngCore};
+use subtle::ConstantTimeLess;
 
-impl Limb {
-    /// Generate a random limb
+impl Random for Limb {
     #[cfg(target_pointer_width = "32")]
     #[cfg_attr(docsrs, doc(cfg(feature = "rand")))]
-    pub fn random(mut rng: impl CryptoRng + RngCore) -> Self {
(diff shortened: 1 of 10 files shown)


src/limb/rand.rs
```diff
@@ -1,20 +1,45 @@
...
+
+        // TODO(tarcieri): use `div_ceil` when available
+        // See: https://github.com/rust-lang/rust/issues/88581
+        let mut n_bytes = modulus.bits() / 8;
+
+        // Ensure the randomly generated value can always be larger than
+        // the modulus in order to ensure a uniform distribution
+        if n_bytes < BYTE_SIZE {
+            n_bytes += 1;
+        }
```
tarcieri, line 34: I wasn't able to find a better way to calculate this until `div_ceil` is available: https://github.com/rust-lang/rust/issues/88581

PR #50, opened 2021-11-16
Faster random_mod

Sc00bz:
When the modulo is 2^(8*n) (or just above) the probability of success for each attempt is 1/256. This makes the worst case 50%.

### DIFF
--- a/src/limb/rand.rs
+++ b/src/limb/rand.rs
@@ -23,20 +23,15 @@ impl RandomMod for Limb {
     fn random_mod(mut rng: impl CryptoRng + RngCore, modulus: &NonZero<Self>) -> Self {
         let mut bytes = <Self as Encoding>::Repr::default();
 
-        // TODO(tarcieri): use `div_ceil` when available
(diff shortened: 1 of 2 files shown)


Sc00bz:
I don't know why I thought this project preferred `a = a & b` over the short hand. Also I just realized that doing `Limb::random_mod` on highest limb is better than the standard way. So the only real fix is for `Limb::random_mod` and `n_limbs = (n_bits + Limb::BIT_SIZE - 1) / Limb::BIT_SIZE`. Oh `n_limbs.saturating_sub(1)` can be `n_limbs - 1` since `n_limbs > 0` because `modulus` is `NonZero`.

Should I do a new pull request but minus all the stupid commits or fix it here?

tarcieri:
@Sc00bz fixing it here is fine

fjarri:
So, what's the status of this PR? I'd like to see this merged, can I help somehow?

tarcieri:
@fjarri looks like it needs a rebase at least

tarcieri:
Merged in #146

PR #56, opened 2022-01-06
Implement const-friendly `NonZero` from `UInt`

daxpedda:
The goal is to enable:
```rust
const P: NonZero<U256> = NonZero::from_uint(U256::from_be_hex(
    "ffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551",
));
```
Alternatively we can also provide `from_be_hex()` & co. methods on `NonZero` itself.


tarcieri:
Looks good, thanks!

PR #67, opened 2022-02-03
[WIP] fast multiplication 

### LINKED ISSUE #66: Improve multiplication
The current implementation of mul_wide uses schoolbook multiplication, which has complexity of the order of O(n*m) where n, m are the number of limbs in the operands. 

Perhaps we should switch to an asymptotically better algorithm like karatsuba multiplication. If so I would be happy to work on this.

pdogr:
Fixes #66 
This is a draft PR to add Toom-Cook multiplication with 3 evaluation points to improve the complexity of multiplication.

### DIFF
--- a/src/uint/mul.rs
+++ b/src/uint/mul.rs
@@ -17,42 +17,22 @@ impl<const LIMBS: usize> UInt<LIMBS> {
     ///
     /// For more info see: <https://github.com/RustCrypto/crypto-bigint/issues/4>
     // TODO(tarcieri): use `concat` (or similar) when const trait is stable
-    pub const fn mul_wide(&self, rhs: &Self) -> (Self, Self) {
-        let mut i = 0;
-        let mut lo = Self::ZERO;
-        let mut hi = Self::ZERO;
+    pub fn mul_wide(&self, rhs: &Self) -> (Self, Self) {
+        let zero = Self::ZERO;
+        let mut r = [zero.limbs, zero.limbs].concat();
+        let mut s = [zero.limbs, zero.limbs, zero.limbs, zero.limbs].concat();
 
-        // Schoolbook multiplication.
-        // TODO(tarcieri): use Karatsuba for better performance?
-        while i < LIMBS {
-            let mut j = 0;
-            let mut carry = Limb::ZERO;
+        toom22(&self.limbs, &rhs.limbs, &mut r, &mut s);
+        let (lo, hi) = r.split_at(LIMBS);
 
-            while j < LIMBS {
-                let k = i + j;
-
-                if k >= LIMBS {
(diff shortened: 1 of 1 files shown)


src/uint/mul.rs
```diff
@@ -17,42 +17,22 @@ impl<const LIMBS: usize> UInt<LIMBS> {
...
-
-                j += 1;
-            }
-
-            hi.limbs[i + j - LIMBS] = carry;
-            i += 1;
-        }
-
-        (lo, hi)
+    pub fn mul_wide(&self, rhs: &Self) -> (Self, Self) {
```
tarcieri, line 20: Please keep this function `const fn`.

You'll need to use loops instead of iterators, but otherwise it shouldn't be too difficult.

pdogr: Yes I switched to using while loops, but we can't use &mut T in const contexts and split_at, split_at_mut are not const fns. I'll have to redesign the impl if mul_wide is to be a const fn.

tarcieri: One way to move this forward would be to keep the current schoolbook multiplication method for `const fn mul_wide`, and then use the fast implementation in the `Mul` traits.

matheus23: Apparently there's a `konst` crate which implements [`split_at_mut` as a const fn](https://docs.rs/konst/latest/konst/slice/fn.split_at_mut.html).
IIUC, for it to be constant-time it needs rust version 1.64 at minimum, and it's using some unsafe code.
However, it seems like this crate's MSRV is 1.65, so perhaps that'd work?

tarcieri: Would prefer not to add additional dependencies. Curious how it works though... I thought all mut references were still unstable in `const fn`

matheus23: Regarding `&mut T` in `const fn`s it seems like this is blocking that: https://github.com/rust-lang/rust/issues/57349, although I may be wrong, I feel out of my depth here.

tarcieri: Yeah, I'm confused how the implementation in the `konst` crate could work on stable. It uses `from_raw_parts_mut` which is marked const: unstable, and the function doesn't compile on 1.70.0: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=79b84c558fb0ada4fa9e6980a797aa93

```
error[[E0658]](https://doc.rust-lang.org/stable/error_codes/E0658.html): mutable references are not allowed in constant functions
```

tarcieri:
Closing in favor of #402 which implements Karatsuba, which seems more appropriate for our applications.

`master` has diverged significantly anyway in regard to multiplication as we have added more functonality.

PR #73, opened 2022-03-25
Serde implementation

daxpedda:
Implemented `serde::Deserialize` and `serde::Serialize` for all types. This might be a bit overkill, implementing it for `UInt` might have been enough.

As serde currently doesn't support const generics (serde-rs/serde#1937), I used [serde-big-array](https://crates.io/crates/serde-big-array) for the `UInt` implementation. The dependency is quiet minimal, this PR also spawned est31/serde-big-array#14 to make sure that the dependency is as minimal as possible.

I hope the tests are appropriate enough.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,12 @@ der = { version = "=0.6.0-pre.3", optional = true, default-features = false }
 generic-array = { version = "0.14", optional = true }
 rand_core = { version = "0.6", optional = true }
 rlp = { version = "0.5", optional = true, default-features = false }
+serde_ = { version = "1", package = "serde", optional = true, default-features = false }
+serde-big-array = { version = "0.4", optional = true }
 zeroize = { version = "1", optional = true,  default-features = false }
 
 [dev-dependencies]
+bincode = "1"
 hex-literal = "0.3"
 num-bigint = "0.4"
 num-traits = "0.2"
@@ -38,6 +41,7 @@ rand_chacha = "0.3"
 default = ["rand"]
 alloc = []
 rand = ["rand_core/std"]
+serde = ["serde_", "serde-big-array"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,7 @@ der = { version = "=0.6.0-pre.3", optional = true, default-features = false }
 generic-array = { version = "0.14", optional = true }
 rand_core = { version = "0.6", optional = true }
 rlp = { version = "0.5", optional = true, default-features = false }
-serde_ = { version = "1", package = "serde", optional = true, default-features = false }
-serde-big-array = { version = "0.4", optional = true }
+serdect = { version = "0.1", optional = true, default-features = false }
 zeroize = { version = "1", optional = true,  default-features = false }
 
 [dev-dependencies]
@@ -41,7 +40,7 @@ rand_chacha = "0.3"
 default = ["rand"]
 alloc = []
 rand = ["rand_core/std"]
-serde = ["serde_", "serde-big-array"]
+serde = ["serdect"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/src/checked.rs
+++ b/src/checked.rs
@@ -87,13 +87,12 @@ impl<'de, T: Copy + Serialize> Serialize for Checked<T> {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "serde"))]
 mod tests {
     use crate::{Checked, U64};
     use subtle::{Choice, ConstantTimeEq, CtOption};
 
     #[test]
-    #[cfg(feature = "serde")]
     fn serde() {
         let test = Checked::new(U64::from_u64(0x0011223344556677));
 
@@ -114,7 +113,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg(feature = "serde")]
     fn serde_owned() {
         let test = Checked::new(U64::from_u64(0x0011223344556677));
 
--- a/src/checked.rs
+++ b/src/checked.rs
@@ -2,6 +2,9 @@
 
 use subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
 
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
 /// Provides intentionally-checked arithmetic on `T`.
 ///
 /// Internally this leverages the [`CtOption`] type from the [`subtle`] crate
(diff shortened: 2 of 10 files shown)


src/checked.rs
```diff
@@ -56,3 +59,78 @@ impl<T> From<Checked<T>> for Option<T> {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value = Option::<T>::deserialize(deserializer)?;
+
+        Ok(Self(match value {
+            Some(value) => CtOption::new(value, Choice::from(1)),
+            None => CtOption::new(T::default(), Choice::from(0)),
+        }))
```
daxpedda, line 75: I'm really not used to using subtle, any feedback is welcome.

daxpedda:
The reason why I was trying to avoid using slices is because when using [`DeserializeOwned`](https://docs.serde.rs/serde/de/trait.DeserializeOwned.html) it doesn't work. The correct way to do this is to use `Visitor`, which is exactly what `serde-big-array` does.

Alternatively, I could just copy the code of `serde-big-array` over, but it uses unsafe, which is maybe where a separate crate like `crypto-serde` could help. `serde-big-array` is really small, this is why I didn't worry about it, the maintainer is est31 (I want to avoid pinging him), he also hosts [`rcgen`](https://crates.io/crates/rcgen).

The main reason I made this PR is because currently `ScalarCore` is using slices to de/serialize and is failing when used with `DeserializeOwned` as pointed out above. My thinking was that if we implement the Serde traits on `UInt` directly we can do it correctly here instead of having to maintain things all over the place.

Preferably I would like to have this merged until we have `crypto-serde`. Alternatively we can scratch this and I can make PRs to fix this in `elliptic-curve` and other crates directly. I could also help jump-start the work on `crypto-serde`, but I have no idea about the vision, so more information would be helpful, you can also ping me on Zulip for that.

daxpedda:
> I'm a bit surprised to hear that `DeserializeOwned` bounds aren't working when using borrowed values which are immediately consumed and turned into owned values. That should work? There's nothing magic about `DeserializeOwned`: it's just a marker trait that uses HRTBs.

I honestly don't know enough about Serde to argue with you, but just to clarify, compilation isn't the problem, it just throws an error during runtime. I will make a small test to exemplify this.

tarcieri:
@daxpedda are you sure the issue wasn't fixed by https://github.com/RustCrypto/elliptic-curves/issues/536 ?

daxpedda:
```toml
[dependencies]
bincode = "1"
# Can't compile without `pem`, this is a bug.
elliptic-curve = { version = "0.12.0-pre.1", features = ["pem"] }
p256 = { version = "0.11.0-pre.0", features = ["serde"] }
serde = "1"

[patch.crates-io]
elliptic-curve = { git = "https://github.com/RustCrypto/traits" }
```

```rust
use p256::Scalar;

fn main() {
    let scalar = Scalar::ZERO;

    let serialized = bincode::serialize(&scalar).unwrap();
    let deserialized: Scalar = bincode::deserialize_from(serialized.as_slice()).unwrap();

    assert_eq!(scalar, deserialized);
}
```

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("invalid type: byte array, expected a borrowed byte array")', src/main.rs:7:81
```

Same error on the latest version on crates.io and master.

daxpedda:
> I think the most likely issue, especially if it's occurring at runtime, is the input format decoder doesn't allow borrowing from the input.
```rust
<&[u8]>::deserialize(deserializer)?;
```
I think so too, this here just doesn't work. As far as I'm aware there is no way to avoid implementing a `Visitor` by hand to solve this problem.

> That's the sort of concern it'd be really nice to solve in a single place like `crypto-serde`.

I'm happy to contribute :smiley:

tarcieri:
Initial implementation of `crypto-serde`: https://github.com/RustCrypto/formats/pull/531

tarcieri:
@daxpedda you can use this now: https://crates.io/crates/serdect/0.1.0

src/uint/macros.rs
```diff
@@ -20,7 +20,7 @@ macro_rules! impl_uint_aliases {
                 }
 
                 fn from_le_bytes(bytes: Self::Repr) -> Self {
-                    Self::from_be_slice(&bytes)
```
daxpedda, line 23: Bug I found on the way. We need to test `Encoding`.

src/uint.rs
```diff
@@ -166,6 +169,37 @@ impl<const LIMBS: usize> fmt::UpperHex for UInt<LIMBS> {
...
+#[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
+impl<'de, const LIMBS: usize> Deserialize<'de> for UInt<LIMBS>
+where
+    UInt<LIMBS>: Encoding,
+{
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let mut buffer = Self::ZERO.to_le_bytes();
```
daxpedda, line 182: Couldn't come up with a cleaner way.

src/checked.rs
```diff
@@ -56,3 +59,76 @@ impl<T> From<Checked<T>> for Option<T> {
...
```
tarcieri, line 75: Perhaps this (provided I got my parens right):

```suggestion
        let value = Option::<T>::deserialize(deserializer)?;
        Self(CtOption::new(value.unwrap_or_default(), Choice::from(value.is_some() as u8)))
```

Cargo.toml
```diff
@@ -24,9 +24,12 @@ der = { version = "=0.6.0-pre.3", optional = true, default-features = false }
 generic-array = { version = "0.14", optional = true }
 rand_core = { version = "0.6", optional = true }
 rlp = { version = "0.5", optional = true, default-features = false }
+serde_ = { version = "1", package = "serde", optional = true, default-features = false }
+serdect = { version = "0.1", optional = true, default-features = false }
```
tarcieri, line 28: `serdect` re-exports `serde` as `serdect::serde`.

Is there some other reason you're pulling it in as an explicit crate dependency?

daxpedda: Yeah: I forgot :laughing:.
I feel very out-of-touch after such a long break.

src/uint.rs
tarcieri: Yeah, it's kind of tough. I think `GenericArray::default()` would work if you use `Self::from_le_byte_array` and the `ArrayEncoding` trait.

Otherwise there's `UInt::<Limbs>::Repr::default()` but unfortunately those impls only go up to 32-bytes, so not only would you need a `Default` bound to use them but also it would limit the serializable size to `U256` unnecessarily.

Cargo.toml
daxpedda: Done.

src/checked.rs
daxpedda: Done, had to make a temporary variable for `Choice` to avoid cloning `value`.

tarcieri (approved):

PR #74, opened 2022-04-28
Pad limbs with zeros when displaying hexadecimal representation.

PopcornPaws:
Limbs use the default `u64`/`u32` `LowerHex` and `UpperHex` implementations when displaying hexadecimal output. This is problematic because when someone creates a number from big-endian hex, e.g.:
```
let hex = "AAAAAAAABBBBBBBB0000000CDDDDDDDD";
let n = U128::from_be_hex(hex);
```
then `n.to_string()` ouputs `AAAAAAAABBBBBBBBCDDDDDDDD` instead of the original `hex`. If the second limb is simply `0`, then the output is `AAAAAAAABBBBBBBB0`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -120,7 +120,7 @@
 //! [`Sub`]: core::ops::Sub
 //! [`CryptoRng`]: rand_core::CryptoRng
 
-#![no_std]
+//#![no_std]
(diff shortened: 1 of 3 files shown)


tarcieri:
Nice catch, thanks!

PR #93, opened 2022-06-11
Add `Encoding` tests

daxpedda:
Follow-up on #73.


tarcieri:
As an additional test, you could [`reverse`](https://doc.rust-lang.org/std/primitive.slice.html#method.reverse) the bytes of one endianness and check it's decodable by the other, and vice versa

daxpedda:
Done.

PR #105, opened 2022-07-28
uint: Implement from_word and from_wide_word

haslersn:
This is useful for clients who write generic code targeting both 32-bit and 64-bit.


tarcieri:
Thanks!

PR #106, opened 2022-07-28
uint: Optimize neg_mod by simply calling sub_mod

haslersn:
The new implementation uses 2 instead of 3 iterations over the limbs.


tarcieri:
Thank you!

PR #108, opened 2022-08-07
uint: Implement modulo operations for special moduli

haslersn:
For a project of mine where the modulus can be chosen to be close to `UInt::MAX`, I created optimized implementations of modular operations. I thought maybe others can benefit from them, too, so I ported my implementation to the crypto-bigint crate.

This commit implements modulo operations (`neg`, `add`, `sub`, `mul`) for special moduli that are so close to `MAX` that the difference to overflow fits in a single `Limb`. For such moduli, these new implementations are much faster than the existing generic modulus implementations. (For `mul` there's no comparison since there's no corresponding generic modulus implementation, yet.)

For `U256`, I benchmarked the generic against the specialized implementations using criterion-rs on Intel Core i7-8565U @ 1.80GHz and obtained the following average times. Note that I used a `const` modulus known at compile-time, which enables some compiler optimizations after inlining. With a modulus known only at runtime, times might differ.

| | `U256::add_mod` | `U256::sub_mod` | `U256::mul_mod` |
|-|-|-|-|
| generic (after #109 got merged) | 10.857 ns | 9.6262 ns | not implemented |
| `_special` | 3.8276 ns | 4.1339 ns | 20.188 ns |

### DIFF
--- a/src/uint.rs
+++ b/src/uint.rs
@@ -22,6 +22,7 @@ mod encoding;
 mod from;
 mod inv_mod;
 mod mul;
+mod mul_mod;
 mod neg_mod;
 mod shl;
 mod shr;
--- a/src/uint/add_mod.rs
+++ b/src/uint/add_mod.rs
@@ -30,6 +30,22 @@ impl<const LIMBS: usize> UInt<LIMBS> {
 
         res
     }
+
+    /// Computes `self + rhs mod p` in constant time for the special modulus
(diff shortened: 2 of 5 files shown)


tarcieri:
This looks reasonable enough at first glance. Give me a little more time to more thoroughly review it.

haslersn:
I had an error in my benchmarks, namely I forgot to pass the inputs through `black_box`. I corrected the original post with updated benchmark results.

tarcieri:
Thank you!

PR #110, opened 2022-08-10
Faster decoding for UInt

andrewwhitehead:
On this machine these changes take `Encoding::from_{le|be}_bytes` from around 200ns to around 3ns for a U256. It adds non-const (but still constant-time) methods for decoding UInts from a byte array.

Restricting the input to an array allows the assert to be optimized out, as is the `unreachable!()` panicking according to godbolt. But there may be a cleaner way to obtain the fixed-size chunks (for example, copying into an intermediate buffer), and in future `slice::as_chunks` could be applicable.

### DIFF
--- a/src/uint/encoding.rs
+++ b/src/uint/encoding.rs
@@ -154,8 +154,45 @@ const fn decode_hex_byte(bytes: [u8; 2]) -> u8 {
     result
 }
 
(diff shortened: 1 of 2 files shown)


tarcieri:
Thanks!

PR #112, opened 2022-08-23
Add left, right bitshift for Limb

andrewwhitehead:
Convenience methods to avoid wrapping and unwrapping the Limb.


tarcieri (approved):

tarcieri:
Thanks!

PR #113, opened 2022-08-23
Faster const decoding of UInt

andrewwhitehead:
As a follow-up to #110, this makes the const `from_be_*` and `from_le_*` methods equally fast as the non-const implementation. Removes the Decoder struct and alternative versions.


tarcieri:
Nice, thank you!

PR #114, opened 2022-08-24
Add a public LIMBS constant to UInt

andrewwhitehead:
This is something I've found myself reaching for. It allows for `U256::LIMBS` (f.ex) to be used in a const context, instead of having to import the `nlimbs!` macro.


tarcieri:
Sure, seems like a good idea

PR #115, opened 2022-08-25
Fix constant-time behaviour for ct_reduce/ct_div_rem

andrewwhitehead:
When using self.bits() instead of the full field width, run time could vary drastically depending on the input, for a modulus significantly smaller than the UInt size.

### DIFF
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -17,7 +17,7 @@ impl<const LIMBS: usize> UInt<LIMBS> {
     /// When used with a fixed `rhs`, this function is constant-time with respect
     /// to `self`.
     pub(crate) const fn ct_div_rem(&self, rhs: &Self) -> (Self, Self, u8) {
-        let mut bd = self.bits().saturating_sub(rhs.bits());
+        let mut bd = (LIMBS * Limb::BIT_SIZE).saturating_sub(rhs.bits());
         let mut rem = *self;
         let mut quo = Self::ZERO;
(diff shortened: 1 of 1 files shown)


tarcieri:
Those names are arguably inconsistent with some of the other naming in the crate, most notably `UInt::shl_vartime`/`::shr_vartime` which are vartime for a non-fixed `rhs`

andrewwhitehead:
The methods state "When used with a fixed `rhs`, this function is constant-time with respect to `self`" but that doesn't appear to be the case. Also I thought I had run tests against this fix, but apparently not. :|

tarcieri:
> ...but that doesn't appear to be the case.

You're saying it's non-constant-time with fixed `rhs` or something else?

If so, I don't see what the issue is.

andrewwhitehead:
For example benchmarking:

```rust
use crypto_bigint::U256;
let u1 = U256::from_be_hex("0123456780123456780123456780123456780123456780123456780123456780");
let u2 = U256::from_be_hex("0000000000000000000000000000000000000000000000000000000000000012");
let modulus = U256::from_be_hex("0000000000000000000000000000000000000000000000000000000012345678");
```

`u1.reduce(&modulus)` is about 6000% slower than `u2.reduce(&modulus)`

tarcieri:
Aah, of course, because a varying parameter is subsequently passed to `shl_vartime`. Good catch!

PR #117, opened 2022-08-25
Fix constant-time behaviour for ct_reduce/ct_div_rem

### LINKED ISSUE #116: `UInt::{ct_div_rem, ct_reduce}` are not constant-time
As noted in #115, both of these functions compute values and then pass them as the `rhs` to `UInt::shl_vartime`.

- https://github.com/RustCrypto/crypto-bigint/blob/ed8d96b/src/uint/div.rs#L24
- https://github.com/RustCrypto/crypto-bigint/blob/ed8d96b/src/uint/div.rs#L63

cc @andrewwhitehead @mikelodder7

andrewwhitehead:
Fixes #116

### DIFF
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -36,11 +36,9 @@ impl<const LIMBS: usize> UInt<LIMBS> {
             quo = quo.shl_vartime(1);
         }
 
-        let is_some = if mb == 0 { 0 } else { 1 };
-        if is_some == 0 {
-            quo = Self::ZERO;
-        }
-        (quo, rem, is_some as u8)
+        let is_some = Limb(mb as Word).is_nonzero();
+        quo = Self::ct_select(Self::ZERO, quo, is_some);
+        (quo, rem, (is_some & 1) as u8)
     }
 
     /// Computes `self` % `rhs`, returns the remainder and
@@ -66,8 +64,8 @@ impl<const LIMBS: usize> UInt<LIMBS> {
             c = c.shr_vartime(1);
         }
 
-        let is_some = if mb == 0 { 0 } else { 1 };
-        (rem, is_some)
+        let is_some = Limb(mb as Word).is_nonzero();
+        (rem, (is_some & 1) as u8)
     }
 
     /// Computes `self` % 2^k. Faster than reduce since its a power of 2.
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -1,7 +1,7 @@
 //! [`UInt`] division operations.
 
 use super::UInt;
-use crate::limb::{SignedWord, Word, HI_BIT};
+use crate::limb::Word;
 use crate::{Integer, Limb, NonZero, Wrapping};
 use core::ops::{Div, DivAssign, Rem, RemAssign};
 use subtle::{Choice, CtOption};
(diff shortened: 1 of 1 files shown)


src/uint/div.rs
```diff
@@ -17,35 +17,29 @@ impl<const LIMBS: usize> UInt<LIMBS> {
...
             if bd == 0 {
                 break;
             }
             bd -= 1;
             c = c.shr_vartime(1);
-            e = e.shr_vartime(1);
+            quo = quo.shl_vartime(1);
+        }
+
+        let is_some = if mb == 0 { 0 } else { 1 };
```
andrewwhitehead, line 39: I believe this is sufficient since there are no constant-time guarantees around `rhs`.

src/uint/div.rs
```diff
@@ -17,35 +17,29 @@ impl<const LIMBS: usize> UInt<LIMBS> {
     /// When used with a fixed `rhs`, this function is constant-time with respect
     /// to `self`.
     pub(crate) const fn ct_div_rem(&self, rhs: &Self) -> (Self, Self, u8) {
-        let mut bd = self.bits().saturating_sub(rhs.bits());
+        let mb = rhs.bits();
+        let mut bd = (LIMBS * Limb::BIT_SIZE) - mb;
         let mut rem = *self;
         let mut quo = Self::ZERO;
-
         let mut c = rhs.shl_vartime(bd);
```
tarcieri, line 24: Doesn't this still shift by a variable number of bits?

andrewwhitehead: It depends only on the value of `rhs`, so it should be constant for a given divisor. Previously it depended on the value of `self` as well.

tarcieri: Aha!

src/uint/div.rs
mikelodder7, line 39:
```suggestion
        let is_some = ((mb | -mb) >> 63) + 1;
```

mikelodder7: Or

```suggestion
        let is_some = Limb::is_zero(mb);
```

tarcieri: Might be good to leave a comment to that effect

tarcieri: Wouldn't `63` need to be `(Limb::BIT_SIZE - 1)`?

This seems ok but I'm also ok with a branch on a value computed from `rhs` where we already assume timing is variable in respect to its value

andrewwhitehead: I threw in a `ct_select` since it doesn't make a significant difference to the runtime.

mikelodder7: Can you add your test case that triggered this PR?

mikelodder7: Want to make sure it’s actually fixed

tarcieri: Just saw this comment.

Perhaps it would make sense to use something like [`sidefuzz`](https://github.com/phayes/sidefuzz) to test constant-time behvaior?

andrewwhitehead: My test was a criterion benchmark, I suspected it from reading the code.

```rs
#[macro_use]
extern crate criterion;
use criterion::{black_box, Criterion};
fn criterion_benchmark(c: &mut Criterion) {
    use crypto_bigint::U256;
    let u1 = U256::from_be_hex("0123456780123456780123456780123456780123456780123456780123456780");
    let u2 = U256::from_be_hex("0000000000000000000000000000000000000000000000000000000000000012");
    let md = U256::from_be_hex("0000000000000000000000000000000000000000000000000000000012345678");
    c.bench_function("reduce large", move |b| {
        b.iter(|| black_box(u1).reduce(&md))
    });
    c.bench_function("reduce small", move |b| {
        b.iter(|| black_box(u2).reduce(&md))
    });
}
criterion_group!(benches, criterion_benchmark);
criterion_main!(benches);
```

Old timing:

```
reduce large            time:   [948.59 ns 949.00 ns 949.46 ns]                           
  6 (6.00%) high mild
  2 (2.00%) high severe

reduce small            time:   [15.520 ns 15.535 ns 15.549 ns]                           
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  2 (2.00%) high severe
```

And patched:

```
reduce large            time:   [955.12 ns 955.84 ns 956.81 ns]                          
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) high mild
  7 (7.00%) high severe

reduce small            time:   [955.58 ns 956.55 ns 957.99 ns]                          
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe
```

andrewwhitehead: Also if somebody feels like implementing a limb-based method this looks interesting: https://www.bearssl.org/bigint.html#modred

PR #118, opened 2022-08-25
Fix ct_select documentation

andrewwhitehead:


tarcieri (approved):
Looks good!

Sidebar: maybe we should add a `debug_assert!` which ensures `c` is either `0` or `Word::MAX`

PR #119, opened 2022-08-26
dudect: add constant-time test for `UInt::reduce`

tarcieri:
Uses the `dudect-bencher` crate, a wrapper for the `dudect` C library:

https://github.com/oreparaz/dudect

Example output:

```
running 1 bench
bench reduce seeded with 0x2311ed8a8b8e6544
bench reduce ... : n == +0.100M, max t = +1.98296, max tau = +0.00628, (5/tau)^2 = 634328

dudect benches complete
```

The important part here is `max t = +1.98296`. If `t` were to exceed 10, then the code is likely *not* constant time. Since it's well below 10, `reduce` appears to be OK.

cc @andrewwhitehead @mikelodder7 @rozbb


dudect/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+//!
+//! These use the `dudect_bencher` crate to check certain operations for
+//! constant-time behavior.
+
+use crypto_bigint::{Limb, UInt, U256};
+use dudect_bencher::{ctbench_main, ctbench_main_with_seeds, BenchRng, Class, CtRunner};
+use rand::{Rng, RngCore};
+
+/// Generate a random `UInt`.
+// TODO(tarcieri): use `UInt::random` when `dudect_bencher` updates `rand`
```
tarcieri, line 11: @rozbb if you have a moment, it'd be nice to get the `rand` version bumped

rozbb: Just published v0.5. Let me know if it works for you.

tarcieri: Seems good!

dudect/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+    // Make 100,000 inputs on each run
+    for _ in 0..100_000 {
+        inputs.push(rand_uint(rng));
+
+        // Randomly pick which distribution this example belongs to
+        if rng.gen::<bool>() {
+            classes.push(Class::Left);
+        } else {
+            classes.push(Class::Right);
+        }
```
tarcieri, line 42: @rozbb is this the best way to do this?

Or should I perhaps make one of the two `Class`es always the same?

tarcieri:
@rozbb I'm running this in CI with a random seed.

Will it actually exit with a nonzero status if `t > 10`? Or is there some way I can make that happen?

rozbb:
I was gonna mention something about this. You need a specific hypothesis of what might cause non-constant-time-ness. So you might want to do experiments where the Left class has small modulus and moduland (if that's a word), and the Right class has large modulus and moduland. Maybe make 4 benches if you wanna check all the combos.

rozbb:
Re t > 10: that threshold is kinda arbitrary, so the crate doesn't do anything special when it's exceeded. You'd probably be able to get away with piping to an awk script here.

tarcieri:
@rozbb at least initially, I'd like to check if timing varies with respect to `self`.

So would examples where `self` is a fixed value versus randomly generated values help in that respect?

tarcieri:
> You'd probably be able to get away with piping to an awk script here.

I noticed upstream dudect prints messages like "For the moment, maybe constant time" and "Probably not constant time".

Is there any way to get `dudect-bencher` to print something like that which is easy to scan for?

tarcieri:
> So you might want to do experiments where the Left class has small modulus and moduland (if that's a word), and the Right class has large modulus and moduland.

Timing is already expected to vary with respect to the modulus, so we can't check that.

I went ahead and pushed up a version which uses fixed vs random values.

rozbb:
I'm not sure if that test would pick up on a flaw, since it's comparing the runtime of a randomly chosen integer with the runtimes of randomly chosen integers. One experiment that comes to mind is to simply pick a large integer and a small integer, measure their reduce runtimes 100,000 times, and check those dists. The most likely source of non-constant-time-ness is probably that, right?

rozbb:
Re CI: I think it's a faux pas to include subjective evaluations with statistical tests. But I know a bit of awk so I'll give it a shot over the weekend

tarcieri:
@rozbb I pushed up a new version that generates two random values `a` and `b`, where `a` is within `0..2^128`, and `b` is within `0..2^256`, and for each iteration selects either `a` or `b` at random

rozbb:
Nice, looks good. Another thing I don't know is whether it helps to randomize the Left/Right choice. Intuitively though it'd probably be less noisy if you tested 50,000 Left followed by 50,000 right.

tarcieri:
@rozbb if I do that, can I just use for loops rather than preallocating a vec? (which I assume was to eliminate potential noise from the RNG)

rozbb:
Oh good point. It could just be two for loops (Left then Right) without allocation

tarcieri:
@rozbb I removed the vecs and random generation and just did simple loops for each example.

For whatever reason I'm now seeing higher variability in max t. It's usually a large negative value, usually `-10..-40`, however I've also seen outliers like `-193` and once `+20`.

rozbb:
It's actually the value of |t| that tells you statistical distance, so large negative values are bad.

I've been playing around with this and it seems really noisy, even with a large-sample random-class vector-based approach (saw t = +0.08 up to t = +2.3, which is a wild range). I don't know why almost everything produces such large |t| values, even for identical functions (just testing `a.reduce(&mod)` versus `a.reduce(&mod)`. Probably something I'm not understanding about microbenchmarks.

The one thing I was able to do to get it to display small consistent |t| values is to do the vec-allocation, and have each `run_one` invocation actually run the function 1000 times. This yields |t| values around 2. I was also able to verify that this detects a timing difference between the classes `a.reduce(mod)` and `a.reduce(smallmod)` where `smallmod` is a 128-bit modulus.

rozbb:
Re CI: the following bash will exit with 1 iff the max |t| > 10:
```bash
cargo run --release \
    | tee /dev/tty \
    | perl -n -e '/max t = [+-](\d*\.?\d*)/; if ($1 > 10) { exit 1 }'
```

tarcieri:
@rozbb ugh, this is fun (actions/runner#241):

https://github.com/RustCrypto/crypto-bigint/runs/8053794965?check_suite_focus=true#step:4:8

```
tee: /dev/tty: No such device or address
```

I also wasn't able to get `max t` as stable as you even with an inner loop. I did at least get it to the point that it's consistently <100 with what's in the PR, and >100 if I use a random modulus.

rozbb:
Ugh this is so odd. I'm starting to wonder if there's a bug in my crate, but it all seems correct from a cursory check. A bug would probably stand out more.

For the TTY issue, `tee /dev/stderr` would be a (slightly less readable) fix

tarcieri:
@rozbb ok, `tee /dev/stderr` did it and we can at least see the results now.

I'll say this at least, when I change the code to be obviously non-constant-time locally, I'm getting `max t` > 100. So I think we can start there in hopes of catching something egregious.

PR #122, opened 2022-09-20
Add bit accessor methods

andrewwhitehead:
This adds `UInt::bit_vartime(index)` for accessing the value of the bit at a given index. It's trivial to cast the result to u8 or bool so I leave that up to the caller, to avoid casting the value twice.

For consistency and accuracy, `bits` is renamed to `bits_vartime`.

### DIFF
--- a/src/uint/bits.rs
+++ b/src/uint/bits.rs
@@ -2,15 +2,19 @@ use crate::{Limb, UInt, Word};
 
 impl<const LIMBS: usize> UInt<LIMBS> {
     /// Get the value of the bit at position `index`, as a 0- or 1-valued Word.
-    #[inline]
-    pub const fn bit(self, index: usize) -> Word {
-        assert!(index < LIMBS * Limb::BIT_SIZE);
-        (self.limbs[index / Limb::BIT_SIZE].0 >> (index % Limb::BIT_SIZE)) & 1
+    /// Returns 0 for indices out of range.
+    #[inline(always)]
+    pub const fn bit_vartime(self, index: usize) -> Word {
+        if index >= LIMBS * Limb::BIT_SIZE {
+            0
+        } else {
+            (self.limbs[index / Limb::BIT_SIZE].0 >> (index % Limb::BIT_SIZE)) & 1
+        }
     }
 
     /// Calculate the number of bits needed to represent this number.
     #[allow(trivial_numeric_casts)]
-    pub const fn bits(self) -> usize {
+    pub const fn bits_vartime(self) -> usize {
         let mut i = LIMBS - 1;
         while i > 0 && self.limbs[i].0 == 0 {
             i -= 1;
@@ -30,13 +34,15 @@ impl<const LIMBS: usize> UInt<LIMBS> {
 
 #[cfg(test)]
 mod tests {
-    use crate::U64;
+    use crate::U128;
 
     #[test]
-    fn bit_get_ok() {
-        let u = U64::from_be_hex("d201000000010000");
-        assert_eq!(u.bit(0), 0);
-        assert_eq!(u.bit(1), 0);
-        assert_eq!(u.bit(16), 1);
+    fn bit_vartime_ok() {
+        let u = U128::from_be_hex("f0010000000000000001000000010000");
+        assert_eq!(u.bit_vartime(0), 0);
+        assert_eq!(u.bit_vartime(1), 0);
+        assert_eq!(u.bit_vartime(16), 1);
+        assert_eq!(u.bit_vartime(127), 1);
+        assert_eq!(u.bit_vartime(130), 0);
     }
 }
--- a/src/uint/bits.rs
+++ b/src/uint/bits.rs
@@ -1,6 +1,13 @@
 use crate::{Limb, UInt, Word};
 
 impl<const LIMBS: usize> UInt<LIMBS> {
+    /// Get the value of the bit at position `index`, as a 0- or 1-valued Word.
+    #[inline]
+    pub const fn bit(self, index: usize) -> Word {
+        assert!(index < LIMBS * Limb::BIT_SIZE);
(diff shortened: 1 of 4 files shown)


src/uint/bits.rs
```diff
@@ -1,6 +1,13 @@
 use crate::{Limb, UInt, Word};
 
 impl<const LIMBS: usize> UInt<LIMBS> {
+    /// Get the value of the bit at position `index`, as a 0- or 1-valued Word.
+    #[inline]
+    pub const fn bit(self, index: usize) -> Word {
```
tarcieri, line 6: It might be good to name this `bit_vartime` to note that the access is variable time with respect to `index`, since it effectively computes a pointer based on its value.

src/uint/bits.rs
```diff
@@ -1,6 +1,13 @@
...
+        assert!(index < LIMBS * Limb::BIT_SIZE);
```
tarcieri, line 7: I think there are a few different ways to avoid panicking in this case:

- Return `0`. This is the [behavior of `num-bigint`](https://docs.rs/num-bigint/0.4.3/src/num_bigint/biguint.rs.html#922-931).
- Return `Option` or `Result`, which would still enable detecting the index is out of range in `const fn` contexts

andrewwhitehead: I prefer the first option if panicking is not desired. I think that generally the index will be known to be in a valid range (for example when used in `for idx in 0..UInt::BIT_SIZE`) and so the assert just gets optimized out.

src/uint/bits.rs
tarcieri: Variable-index table lookups are a classical source of microarchitectural sidechannel attacks, e.g. cache timing attacks

src/uint/bits.rs
tarcieri: First option seems good to me too

src/uint/bits.rs
tarcieri: I should probably also note a fully constant-time version is also possible but it would be significantly slower: you'd need to scan over every word every time, accumulating the result as a `Word`, and masking the value with bitwise arithmetic if the index is not within a given word

andrewwhitehead: Yes I considered that, but I can't think of a use case right now. I suppose `bit_vartime` it is. In my case I'm just iterating through all the bits so there's no security concern.

I should note that the `bits` method below does not appear to be constant time either due to the initial iteration over the zero-valued limbs.

src/uint/bits.rs
```diff
@@ -1,6 +1,34 @@
...
+    /// Unlike `bit_vartime`, this method is constant time in regards to the index.
+    /// Returns 0 for indices out of range.
+    #[inline(always)]
+    pub const fn bit(self, index: usize) -> Word {
+        let mut limb = Limb::ZERO;
+        let mut i = 0;
+
+        while i < LIMBS {
+            let c = i.wrapping_sub(index / Limb::BIT_SIZE) as Word;
+            limb = Limb::ct_select(self.limbs[i], limb, Limb(c).is_nonzero());
```
andrewwhitehead, line 14: I haven't manually verified that this is constant time.

src/uint/bits.rs
tarcieri: Yeah good observation re: `bits` cc @mikelodder7

mikelodder7: It’s not supposed to be constant time. Renaming is fine.

src/uint/bits.rs
mikelodder7: Yep agreed

src/uint/bits.rs
mikelodder7: Is there a reason to make this constant time? To make this constant time requires touching every bit which doesn’t seem necessary

tarcieri: @mikelodder7 nah, I'm just more concerned with naming and making it clear when functions have non-constant-time behavior

mikelodder7: Got it. Fair enough

andrewwhitehead:
I've updated this PR to only add `bit_vartime` and skip `bit` for now.

src/uint/sqrt.rs
```diff
@@ -25,7 +25,7 @@ impl<const LIMBS: usize> UInt<LIMBS> {
             // Sometimes an increase is too far, especially with large
             // powers, and then takes a long time to walk back.  The upper
             // bound is based on bit size, so saturate on that.
-            let res = Limb::ct_cmp(Limb(xn.bits() as Word), Limb(max_bits as Word)) - 1;
+            let res = Limb::ct_cmp(Limb(xn.bits_vartime() as Word), Limb(max_bits as Word)) - 1;
```
andrewwhitehead, line 28: This use of `bits_vartime` seems a little concerning after the renaming? @tarcieri

tarcieri: cc @mikelodder7

src/uint/bits.rs
```diff
@@ -1,9 +1,20 @@
...
+        if index >= LIMBS * Limb::BIT_SIZE {
+            0
+        } else {
+            (self.limbs[index / Limb::BIT_SIZE].0 >> (index % Limb::BIT_SIZE)) & 1
+        }
+    }
+
     /// Calculate the number of bits needed to represent this number.
     #[allow(trivial_numeric_casts)]
-    pub const fn bits(self) -> usize {
```
tarcieri, line 6: Since `master` is still v0.4.x, can you keep this around with a `#[deprecated]` attribute, calling `bits_vartime`?

We can remove it in the next breaking release.

src/uint/sqrt.rs
mikelodder7: Seems okay for now. Not a clear direction yet going forward

tarcieri:
Looks good now, thanks!

PR #123, opened 2022-09-22
ci: update actions used in GitHub Actions workflows

striezel:
Updates the actions [`actions/checkout`](https://github.com/actions/checkout) and [`actions/cache`](https://github.com/actions/cache) to v3, their latest major release.

This should basically be backwards compatible, so I expect no breakage.


tarcieri:
Thanks!

PR #127, opened 2022-09-25
Optimize neg_mod

andrewwhitehead:
This change optimizes `neg_mod` (25% faster here) by performing a bit mask instead of `adc`.


tarcieri:
Thank you!

Issue #70, opened 2022-03-09
MulMod not implemented and not properly documented


haslersn:
It would be nice if the trait `MulMod` could be implemented.

Furthermore, the [current documentation (v0.3.2)](https://docs.rs/crypto-bigint/0.3.2/crypto_bigint/trait.MulMod.html) states that `mul_mod`

> Requires `p_inv = -(p^{-1} mod 2^{BITS}) mod 2^{BITS}` to be provided for efficiency.

However, any even `p` is not invertible. An even `p` might be used e.g. when computing over a ring rather than a prime field. So, how should `p_inv` be set if `p` is even?

haslersn:
For an even modulus `m` one can probably do the following: Decompose `m =: 2^k * n` such that `n` is odd. Then perform the multiplication separately mod `2^k` and mod `n`, and combine the results using CRT.

haslersn:
> The trait in its current form is specialized to a Montgomery reduction

@tarcieri [Wikipedia says](https://en.wikipedia.org/wiki/Montgomery_modular_multiplication):

> The need to convert a and b into Montgomery form and their product out of Montgomery form means that computing a single product by Montgomery multiplication is slower than the conventional or [Barrett reduction](https://en.wikipedia.org/wiki/Barrett_reduction) algorithms.

Did you take this into account when chosing Montgomery multiplication?

haslersn:
Looking at [the old implementation](https://github.com/RustCrypto/utils/pull/510/files#diff-414a3020881fe024f6738c3b212796e285e04e106cdc496d5aee5964c52ec20fR81-R152), it looks like it doesn't bring the number into Montgomery form before calling `mont_reduce`.

tarcieri:
Yes, that is definitely why it didn't work.

We use both Montgomery and Barrett reductions in the https://github.com/rustcrypto/elliptic-curves crates (for base and scalar field elements respectively, although the latter only for `p256`) and the fact we have modular multiplication implemented already on a base/scalar field-by-field basis means we haven't been particularly motivated to implement a generic solution in `crypto-bigint` yet, although it would hypothetically be useful for migrating the `dysa` and `rsa` crates off of [`num-bigint-dig`](https://crates.io/crates/num-bigint-dig).

For the elliptic curve crates specifically we generally perform a sequence of operations, and thus the overhead of converting to/from Montgomery form is ammortized.

[...]

Issue #58, opened 2022-01-16
Heap-backed fixed-width integers


tarcieri:
Somewhat in the spirit of #29, I have been wondering if it would make sense to support a heap-backed integer type (cc @elichi)

The idea would be that integers are still fixed-width and would otherwise work the same as the const generic counterparts, but the width could be chosen at runtime/initialization time, rather than at compile time. So they're not really "arbitrary precision" in that they don't grow/shrink, they're still fixed-width, just fixed at a width determined at runtime.

One place where this would be applicable is the `rsa` crate. While it'd be nice for microcontroller users to be able to instantiate RSA with e.g. 2048-bit or 4096-bit keys using only the stack, other use cases like OpenPGP implementations may need to support arbitrary key sizes running upwards of 16384-bit. For these use cases where key sizes can potentially vary wildly, it'd be nice to choose the integer size at runtime. The `Integer` and other traits in this crate would permit accepting either. (cc @dignifiedquire)

Internally this would involve implementing core algorithms that operate over slices instead of fixed-size arrays. This could have other benefits: it could reduce the monomorphization penalty for different const generic sizes by having them pass a slice to a function which can operate over many sizes.

There are some potential drawbacks: we don't presently unroll loops, but in some cases it'd be nice and generate better code, and this could potentially negatively impact such code. I think the solution there would be to duplicate some of that loop unrolled code between the stack and heap-backed implementations, so the stack-allocated version always unrolls the loops ahead-of-time, and the heap-backed version would need to loop over a slice.

tarcieri:
Another downside I forgot to mention:

The current approach of defining all operations on a newtype for `[Limb; LIMBS]` allows the compiler to optimize away bounds checks in certain cases because it knows a priori that accesses on literal indexes or indexes whose range can be deduced at compile time will always be in-range.

Implementing operations in terms of `&[Limb]` could preclude these optimizations, although using `#[inline(always)]` may be sufficient to ensure they're always applied if possible.

dignifiedquire:
one trick I have seen is delegsting to the const versions on the dynamic versions as soon as the size is known.

newpavlov:
[...]

Issue #28, opened 2021-10-31
[RFC] Being generic over a modulo instead of a size


elichai:
If we think from an algebraic eyes, it would be very nice to enforce typing over `F_q` for some `q`, meaning that the type will be `FieldElement<q>`, then `q` can be use to choose the size of the underlying array and generate arithmetic that work on that field specifically (and then using specialization we could provide efficient implementation for common fields).

This allows you to easily separate variables from different fields and makes field conversions very explicit and obvious, which should increase resilience in complex algorithms that do convert between fields (ie in a MPC protocol where you use Paillier encryption over a field element in an EC group, when you convert between the fields you'll realize you probably need to attach some range proof and dlog proof).

Would love to hear any thoughts on this :)

newpavlov:
I [think](https://github.com/RustCrypto/crypto-bigint/issues/3#issuecomment-915618162) that we should build such types on top of "simple" bigints as wrappers. But, unfortunately, we probably will be blocked on insufficient power of const generics for the near future.

tarcieri:
Yes, it'd be great if we could make the modulus a const generic parameter, unfortunately you can only use `const N: usize` on `stable`

elichai:
Even on nightly, I tried using a string to represent the moduli and very fast gotten into a compiler bug: https://github.com/rust-lang/rust/issues/90455

But maybe being generic over a constant Uint<N> could work in the future, having a const generic that is generic over a const generic seem currently impossible even in nightly (`struct FieldElement<const N: usize, const P: Uint<{N}>>`)

tarcieri:
I think we can close this as being implemented by what the latest prereleases call `ConstMontyForm`:

https://docs.rs/crypto-bigint/0.6.0-pre.12/crypto_bigint/modular/struct.ConstMontyForm.html

Though it doesn't use `adt_const_params` to be const generic around a modulus, it supports an `impl_modulus!` macro which defines a set of `ConstMontyParams` at compile time, and the `ConstMontyForm` type is generic around a modulus specified by those params.

So it's still a type-safe way of expressing field elements whose modulus is known at compile time.

PR #130, opened 2022-10-11
Implement modular arithmetic

jellevos:
Hi @tarcieri and the other contributors! For the `scicrypt` crate I was working on constant-time big integers by wrapping GMP, but the interface was extremely clumsy. I decided to switch to this crate (which looks awesome!) and added the necessary methods for modular arithmetic.

I implemented the following functionality:
- Shifting left and right by 1 and returning the overflow as a `Choice`
- `Neg` for `Wrapping<Uint<LIMBS>>`
- Modular multiplicative inverse
- Computing parameters for the Montgomery form (MF)
- Going to and from MF
- Modular addition (in MF)
- Modular multiplication (in MF)
- Modular exponentiation (in MF)

In my opinion, there are at least two large issues right now that I would like to hear your opinion on.
- [x] The `Modular` interface is not satisfying to me because it copies the `MontgomeryParameters` and does not assert that two `Modular` integers indeed share the same modulus.
- [x] None of the functions are currently `const fn`.

Still, I thought it was good to already make a PR to bring this to your attention. I am happy to put in the effort to bring this to a point where it is mergeable! :)

**Edit:**
The modulus is now a constant and the struct is called `Residue` rather than `Modular`.


jellevos:
I now see issues #70 and #28. I think `adt_const_params` will take quite a while before it is stabilized, but in the meantime one option is not to use a constant modulus and instead use debug asserts to check that e.g. the two moduli in a multiplication are the same. Based on these two issues I will change `Modular` to `Residue` and change as many functions as possible to be `const`.

tarcieri:
@jellevos the workaround that @haslersn suggested which uses a ZST + trait with an associated constant seems OK as a workaround until `adt_const_params` is stabilized:

https://github.com/RustCrypto/crypto-bigint/issues/70#issuecomment-1193319917

```rust
pub trait ResidueParameters<const L: usize>: PartialEq + Debug {
    const MODULUS: Option<UInt<L>>;
}

#[derive(PartialEq)]
pub struct Residue<P, const L: usize>
where
    P: ResidueParameters<L>,
{
    repr: UInt<L>,
    phantom: PhantomData<P>,
}
```

It could be wrapped up in a macro which defines a ZST and impls the trait for it like so to reduce the boilerplate:

```rust
impl_modulus!(Modulus, "ffffffff00000001...");
```

tarcieri:
Also thanks for submitting this! I'll try to review it in depth when I have some time.

jellevos:
> It could be wrapped up in a macro which defines a ZST and impls the trait for it like so to reduce the boilerplate:
> 
> ```rust
> impl_modulus!(Modulus, "ffffffff00000001...");
> ```

This is a really good idea! One thing that is not immediately clear to me is how ergonomic this will be for non-hardcoded moduli, but I will investigate.

> Also thanks for submitting this! I'll try to review it in depth when I have some time.

Feel free to wait a bit until I have incorporated the changes above. :) If you like, I can ping you when they are ready.

tarcieri:
> This is a really good idea! One thing that is not immediately clear to me is how ergonomic this will be for non-hardcoded moduli, but I will investigate.

The philosophy of this crate is to largely focus on compile-time optimizations, and perhaps do a secondary pass for things that are dependent on runtime values (e.g. #58).

Having a `const MODULUS` unlocks some potential interesting future optimizations when used in conjunction with other `const fn` methods of `UInt`.

It would enable checking special properties of the modulus, and using those properties to select particular implementations in a way this could all be known at compile time. This would allow the compiler to optimize away the branches, so you could provide several implementations of e.g. modular sqrt, picking the best one for a particular modulus at compile time.

jellevos:
@tarcieri The implementation is mostly finished, but I am running into a seemingly fundamental issue. Specifically, I get the following error when running the tests in `uint/modular/pow.rs` on stable:
```
error: any use of this value will cause an error
  --> src/uint/shr.rs:67:20
   |
67 |                   if i < (LIMBS - 1) - full_shifts {
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |                      |
   |                      exceeded interpreter step limit (see `#[const_eval_limit]`)
   |                      inside `uint::shr::<impl uint::UInt<16_usize>>::shr_vartime` at src/uint/shr.rs:67:20
   |                      inside `uint::shr::<impl uint::UInt<16_usize>>::shr_vartime_wide` at src/uint/shr.rs:89:17
   |                      inside `uint::div::<impl uint::UInt<16_usize>>::ct_reduce_wide` at src/uint/div.rs:100:17
   |                      inside `<modular::pow::tests::Modulus as ResidueParams<{ nlimbs!(<$uint_type>::BIT_SIZE) }>>::R2` at src/uint/modular/macros.rs:24:17
   |
  ::: src/uint/modular/macros.rs:23:13
   |
23 | /             const R2: $crate::UInt<{ nlimbs!(<$uint_type>::BIT_SIZE) }> =
24 | |                 $crate::UInt::ct_reduce_wide(Self::R.square_wide(), &Self::MODULUS).0;
   | |______________________________________________________________________________________-
   |
   = note: `#[deny(const_err)]` on by default
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #71800 <https://github.com/rust-lang/rust/issues/71800>

error[E0080]: evaluation of `uint::modular::Residue::<uint::modular::pow::tests::Modulus, 16_usize>::new` failed
  --> src/uint/modular/mod.rs:48:41
   |
48 |         let product = integer.mul_wide(&MOD::R2);
   |                                         ^^^^^^^ referenced constant has errors

note: the above error was encountered while instantiating `fn uint::modular::Residue::<uint::modular::pow::tests::Modulus, 16_usize>::new`
  --> src/uint/modular/macros.rs:36:9
   |
36 |         $crate::uint::modular::Residue::<$modulus, { $modulus::LIMBS }>::new($variable)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
  ::: src/uint/modular/pow.rs:56:24
   |
56 |         let base_mod = residue!(base, Modulus);
   |                        ----------------------- in this macro invocation
   |
   = note: this note originates in the macro `residue` (in Nightly builds, run with -Z macro-backtrace for more info)
```

In other words, we are hitting the compiler's `const_eval_limit`, and I do not immediately see a way to fix this. ☹️ On nightly, the fix is simple: I have added the following snippet in `lib.rs`, which basically turns off this check (we could also just put in a larger number than the default, rather than completely omitting the check):
```
#![feature(const_eval_limit)]
#![const_eval_limit = "0"]
```

What do you think? Do you have any clever ideas on how to get around this? Apart from this, I only need to add a few docstrings before the PR should be ready for review.

jellevos:
I suppose we could also shrink the modulus in those tests, which would make the tests and library work on stable but would force dependencies that use somewhat large moduli to be nightly.

haslersn:
> the workaround that @haslersn suggested which uses a ZST + trait with an associated constant seems OK as a workaround

@tarcieri In practice this was very cumbersome, because one has to drag the `const L` parameter through everything. So instead I ended up creating a trait `GenericUInt` that abstracts over `UInt<L>` and using it like this:

```rust
pub trait ResidueParameters: PartialEq + Debug + Send + 'static {
    type UInt: GenericUInt;
    const MODULUS: Option<Self::UInt>;
}

#[derive(PartialEq)]
pub struct Residue<P>
where
    P: ResidueParameters,
{
    repr: P::UInt,
    phantom: PhantomData<P>,
}
```

tarcieri:
@jellevos wow, that is really unfortunate. I was unaware `const_eval_limit` was set so low as to preclude these sorts of computations from working.

I see there's a tracking issue for the low limit precluding other real-world usages: https://github.com/rust-lang/rust/issues/93481 (The main tracking issue for `const_eval_limit` is https://github.com/rust-lang/rust/issues/67217).

Tough to say what to do about this. It would be good to determine that the caller can change this limit, or otherwise if `crypto-bigint` needs to set it, we'd need to add something like a `nightly` crate feature which bumps the limit.

tarcieri:
> So instead I ended up creating a trait `GenericUInt` that abstracts over `UInt<L>`

@haslersn FWIW the existing `Integer` trait largely provides that.

haslersn:
@tarcieri yes, but the `Integer` trait lacks access to modular arithmetic (`wrapping_add`, ...), direct access to limbs, and conversion from primitive types such as `u32`.

tarcieri:
@haslersn you can add those bounds to the existing `Integer` trait, rather than making a new one

haslersn:
@tarcieri yes, when adding the functionality directly in crypto-bigint, that would be the way to go.

jellevos:
> Tough to say what to do about this. It would be good to determine that the caller can change this limit, or otherwise if `crypto-bigint` needs to set it, we'd need to add something like a `nightly` crate feature which bumps the limit.

I think the best solution in the mean time would be to make sure the tests in this library pass on stable and then leave the problem with the user (i.e. a user must switch to nightly if the const computations exceed the current limit).

I can finish these changes tomorrow, then I would say this is ready for review. I will add a clear statement in the docs that one may need to up `const_eval_limit`.

tarcieri:
@jellevos sounds good!

FWIW I left a comment here: https://github.com/rust-lang/rust/issues/93481#issuecomment-1279997537

jellevos:
> > the workaround that @haslersn suggested which uses a ZST + trait with an associated constant seems OK as a workaround
> 
> @tarcieri The initial attempt at this was very cumbersome, because one has to drag the `const L` parameter through everything.

Currently, I use a second macro `residue!` that at least prevents the end user from being exposed to this (i.e. having to set `const LIMBS`). It takes a `UInt` and such a modulus trait and returns a `Residue`.

jellevos:
@tarcieri Ready for review 😄 I did have to bump the minimum rustc version to 1.61, as the previous versions did not stabilize trait bounds on const functions.

tarcieri:
@jellevos just did a quick scan and it looks good at first glance.

I was hoping to do a more in-depth review this weekend but didn't have time, and also you were still making changes.

I'll try to take a look at this soon, hopefully next weekend.

tarcieri:
I was hoping to get around to look at this over the weekend but was too busy. Will look at it soon.

jellevos:
> The philosophy of this crate is to largely focus on compile-time optimizations, and perhaps do a secondary pass for things that are dependent on runtime values (e.g. #58).
> 
> Having a `const MODULUS` unlocks some potential interesting future optimizations when used in conjunction with other `const fn` methods of `UInt`.
> 
> It would enable checking special properties of the modulus, and using those properties to select particular implementations in a way this could all be known at compile time. This would allow the compiler to optimize away the branches, so you could provide several implementations of e.g. modular sqrt, picking the best one for a particular modulus at compile time.

I did some thinking about this, and from my perspective, it would seem that supporting variable moduli would not directly contradict this crate's philosophy. Those heap-backed integers to me are a different beast: They involve types for which we do not know their concrete size at compile time, but for a variable modulus we can generally assume to know. While the current const moduli enable implementing (a compile-time optimized!) version of e.g. integer-based ElGamal, it would not allow us to implement e.g. RSA or Paillier without knowing the public key ahead of time.

For the crate that I maintain (scicrypt, definitely still WIP) supporting variable moduli would allow me to completely switch away from other crates/libraries and use `crypto-bigint` to implement the cryptosystems, so they are constant-time. I also believe it might be useful for the RSA crate, which then may be able to move away from the not necessarily constant-time `num-bigint` crate.

If you are okay with exploring this direction, I am happy to implement this in one of the following weeks. The API would require some thinking (safety vs ergonomics).

tarcieri:
@jellevos given the limitations we've encountered, backing things out to runtime-based moduli seems fine

jellevos:
@tarcieri Just for completeness: I only hit the const limit when testing with 2048-bit moduli so the code as it is (without runtime-based moduli) should already be useful for applications with smaller moduli. 😄 Considering this, I will address runtime-based moduli in a second PR if you agree.

src/uint/inv_mod.rs
```diff
@@ -25,11 +27,72 @@ impl<const LIMBS: usize> UInt<LIMBS> {
         }
         x
     }
+
+    /// Computes the multiplicative inverse of `self` mod `modulus`. In other words `self^-1 mod modulus`. Returns `(inverse, 1...1)` if an inverse exists, otherwise `(undefined, 0...0)`. The algorithm is the same as in GMP 6.2.1's `mpn_sec_invert`.
+    pub const fn ct_inv_mod(self, modulus: &UInt<LIMBS>) -> (Self, Word) {
```
tarcieri, line 32: This crate has a "constant time by default unless explicitly marked otherwise" policy, so the `ct_*` prefixes on the functions are redundant.

We explicitly mark variable-time functions as `*_vartime`.

I think there were a handful of `pub(crate)` functions with `ct_*` names before which should probably be renamed, but this PR is greatly expanding that and exposing them as part of the public API.

```suggestion
    pub const fn inv_mod(self, modulus: &UInt<LIMBS>) -> (Self, Word) {
```

src/uint/inv_mod.rs
```diff
@@ -25,11 +27,72 @@ impl<const LIMBS: usize> UInt<LIMBS> {
...
+        debug_assert!(modulus.ct_is_odd() == Word::MAX);
+
+        let mut a = self;
+
+        let mut u = UInt::ONE;
+        let mut v = UInt::ZERO;
+
+        let mut b = *modulus;
```
tarcieri, line 40: If you're derefing anyway, perhaps consider making `modulus` owned instead of borrowed?

src/uint/inv_mod.rs
tarcieri: Oh, scrolling down I now see there's a separate definition of `inv_mod`.

Not sure how better to name the const versions other than `const_*`, but having one named `ct_*` and the other not would generally imply to me one is constant time and the other is not, which is not the case here.

src/uint/modular/macros.rs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            const LIMBS: usize = { nlimbs!(<$uint_type>::BIT_SIZE) };
+            const MODULUS: $crate::UInt<{ nlimbs!(<$uint_type>::BIT_SIZE) }> =
+                <$uint_type>::from_be_hex($value);
+            const R: $crate::UInt<{ nlimbs!(<$uint_type>::BIT_SIZE) }> = $crate::UInt::MAX
+                .ct_reduce(&Self::MODULUS)
+                .0
+                .wrapping_add(&$crate::UInt::ONE);
+            const R2: $crate::UInt<{ nlimbs!(<$uint_type>::BIT_SIZE) }> =
+                $crate::UInt::ct_reduce_wide(Self::R.square_wide(), &Self::MODULUS).0;
```
tarcieri, line 23: If this is the only usage of `square_wide`, it seems like it can be trivially replaced with `Self::R.mul_wide(&Self::R)`

src/uint/pow_mod.rs
```diff
@@ -0,0 +1,42 @@
+// use crate::{Concat, Split, UInt};
```
tarcieri, line 1: Perhaps remove this file and PR it separately?

src/uint/inv_mod.rs
jellevos: My mistake, this function should have been `pub(crate)`. Note that this function is in fact (supposed to be) constant-time. I took the `ct_*` prefix to mean that these functions cannot work with `subtle` because they are `const`, but I see that this is indeed not always consistent. Perhaps this naming issue could be addressed in a future PR?

src/uint/modular/macros.rs
jellevos: Is it okay if I leave it like this so I can implement a faster square method later today? Modular exponentiation would benefit greatly from an optimized squaring method. 😄

src/uint/pow_mod.rs
jellevos: Oops, my bad. This is old code that was not supposed to be pushed. 😅

jellevos:
@tarcieri Ready for review! Later today I expect to finish the efficient squaring method in a separate PR.

jellevos:
**Summary of new changes:**
- Changed folder structure to accommodate runtime-based moduli neatly.
- I forgot to implement the inverse for residues before; that's done now.
- Renamed `Residue` to `ConstResidue`, I reserved `Residue` for residues with a modulus decided at runtime.
- Added traits to provide a consistent API between `ConstResidue`s and `Residue`s.

I want to raise an issue about the multiplicative inverses as they are currently very unsafe to use. Currently, I only check that an inverse exists using `debug_assert!`. Release builds, however, do not panic when an inverse does not exist. Instead, they actually continue on with computations on a residue with an undefined value.

@tarcieri Do you have any clever ideas on how to handle this? One idea is to return a `CtOption`, but then the inverse cannot be a `const fn`. Thanks for your help!

tarcieri:
> Renamed `Residue` to `ConstResidue`, I reserved `Residue` for residues with a modulus decided at runtime.

I'd personally prefer (and it would be more consistent with the other RustCrypto projects) if it were the reverse of that, with `Residue` being const-friendly, and something like `DynResidue` with a runtime-dependent modulus.

jellevos:
I agree, I reverted it. 😄 I am still thinking about how to best handle the problem with multiplicative inverses:

> I want to raise an issue about the multiplicative inverses as they are currently very unsafe to use. Currently, I only check that an inverse exists using debug_assert!. Release builds, however, do not panic when an inverse does not exist. Instead, they actually continue on with computations on a residue with an undefined value.
>
> @tarcieri Do you have any clever ideas on how to handle this? One idea is to return a CtOption, but then the inverse cannot be a const fn. Thanks for your help!

tarcieri:
@jellevos yeah, that's a tricky and annoying tradeoff.

You could have a `const fn` version that panics, and a non-`const fn` version that returns a `CtOption`, which are both wrappers around a third internal (`const fn` version that does something like return the `Choice` selector as a `u8`, but that's messy.

tarcieri:
Cool, will take another look this weekend

tarcieri (approved):

haslersn:
When using this much appreciated feature in my project, I needed to do some changes to the API surface around `Residue`. Some changes were necessary to make it usable at all (from different crates) and some of them were meant to simplify usage of Residues, especially in generic contexts.

I'd like to PR these changes, discuss them, and reach a consensus. However, I'll only get to that after the paper deadline, which is in about a week.

So it would be nice if you could wait a bit with the next release.

tarcieri:
No worries there, it's going to be at least a few months before we're ready to cut another release

PR #134, opened 2022-11-11
Runtime moduli

jellevos:
This PR extends #130. It adds residues for moduli decided at run-time.


src/uint/modular/mod.rs
```diff
@@ -1,17 +1,62 @@
-use crate::UInt;
```
tarcieri, line 1: This PR looks good aside from the use of the 2015 edition module style (i.e. `mod.rs`)

We've adopted the 2018 edition module style, where this file would be modular.rs.

Unfortunately I don't think they ever added a lint for mixed module styles.

tarcieri:
Will fix it up retroactively

src/uint/modular/mod.rs
daxpedda: Is this what you mean?:
https://rust-lang.github.io/rust-clippy/master/index.html#mod_module_files

tarcieri: Oh ho, went looking for that the other day and didn't find it... weird

PR #141, opened 2022-12-08
Add constant-time division by a single limb

fjarri:
Based on https://gmplib.org/~tege/division-paper.pdf . This works ~100x faster than `div_rem()` for `U256`.

For reviewers:
- Suggestions on how exactly to expose the API (including the method naming) are welcome. 
- Should `div_rem()` and related methods take `NonZero<Self>` too, and this way avoid returning a `CtOption`?

Changes:
- Added public `Uint` methods: `ct_div_rem_limb_with_reciprocal()`, `div_rem_limb_with_reciprocal()`, `div_rem_limb()`.
- Added `Div`, `Rem`, `DivAssign`, and `RemAssign` implementations for the division by `Limb`.
- Added public `Reciprocal` type (exported at the root level).
- Added `Limb::ct_lt()` and `ct_le()` (constant time < and <=), with the return value applicable in `ct_select()`.
- Separated `Uint::shl_vartime()` into `shl_vartime()` proper (that only shifts by whole limbs and calls `shl_limb()`) and `shl_limb()` (shifts within one limb, constant-time).
- Added benchmarks (division only for now).
- Removed `Uint<LIMBS>: Integer` bound from `Div`, `Rem`, `DivAssign`, and `RemAssign` of `Uint` by `Uint`.

TODO:
- There are suggestions for implementing schoolbook division in the paper, so I think it may be worth it to try to compare that to the currently used algorithm. But that's better left for another PR.


tarcieri:
@fjarri going to need a rebase, but that should be it for breaking changes from me for now

Cargo.toml
```diff
@@ -45,3 +47,10 @@ serde = ["serdect"]
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
+
+[[bench]]
+name = "bench"
+harness = false
+
+[profile.bench]
+opt-level = 3
```
tarcieri, line 56: FWIW I think this is the default for the bench profile (which is the same as the `release` profile)

tarcieri: Confirmed: https://doc.rust-lang.org/cargo/reference/profiles.html#bench

> The `bench` profile inherits the settings from the `release` profile.

```
[profile.release]
opt-level = 3
```

src/uint/div_limb.rs
```diff
@@ -0,0 +1,323 @@
...
+#[inline(always)]
+const fn ct_lookup<const N: usize>(arr: &[u16; N], idx: u16) -> u16 {
+    let mut t = arr[0];
+    let mut j: u16 = 1;
+    while j < N as u16 {
+        t = ct_select(t, arr[j as usize], ct_eq(idx, j));
+        j += 1;
+    }
+    t
+}
```
tarcieri, line 38: Curious if anyone has made a more efficient bitsliced implementation of those reciprocal tables

tarcieri:
Perhaps you could `impl<...> Div<NonZero<Limb>> for Uint<...>`? (also `DivAssign`)

Cargo.toml
fjarri: Fixed.

src/uint/div_limb.rs
fjarri: That was a great suggestion, I reimplemented it via shift-and-subtract, and it's actually much faster this way.

fjarri:
@tarcieri : should I change the regular division API to taking `NonZero<Uint>` in this PR?

tarcieri:
@fjarri yeah, if there are remaining places where it doesn't (the `const fn`s?) it'd be good to use it everywhere for consistency

It should be used on all the `Div` trait impls at least

tarcieri (approved):
Up to you if you want to change all `div` methods to use `NonZero` divisors

tarcieri:
Thank you!

PR #145, opened 2022-12-11
Fix benchmarks compilation

fjarri:
- Fixed some compilation errors forgotten in #141 
- Added benchmark building to CI to prevent that from happening again (just building, not running, to save time)


tarcieri:
Thanks!

PR #146, opened 2022-12-11
Faster random_mod

fjarri:
Superseding #50 

- Rebased
- Fixed a bug in `Limb::random_mod()`
- Small style fixes


tarcieri:
Thanks both of you!

PR #149, opened 2022-12-15
Add `SubResidue` trait and impls for `Residue` and `DynResidue`

fjarri:
Not sure about adding `neg()`:
- Should there be a separate `NegResidue` trait, or should it be a part of `SubResidue`?
- The default implementation (via `SubResidue`) would require an access to zero, and therefore an additional trait bound, which `DynResidue` does not implement.

### DIFF
--- a/src/uint/modular.rs
+++ b/src/uint/modular.rs
@@ -13,11 +13,18 @@ mod add;
 mod inv;
 mod mul;
 mod pow;
+mod sub;
 
 /// Provides a consistent interface to add two residues of the same type together.
 pub trait AddResidue {
     /// Computes the (reduced) sum of two residues.
     fn add(&self, rhs: &Self) -> Self;
+}
+
+/// Provides a consistent interface to subtract two residues of the same type.
+pub trait SubResidue {
+    /// Computes the (reduced) difference of two residues.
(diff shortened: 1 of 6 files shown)


tarcieri:
> Should there be a separate NegResidue trait, or should it be a part of SubResidue?

A separate `NegResidue` trait seems fine to me

fjarri:
Do you want me to add `NegResidue` in this PR, or leave it for another one?

tarcieri:
Seems fine in this PR

fjarri:
Speaking of, why are there separate `*Residue` traits at all? Wouldn't the `core::ops` ones suffice?

tarcieri:
Yeah, I've had similar thoughts but never broached them. We should probably switch to the `core::ops` traits.

cc @jellevos

tarcieri:
@fjarri perhaps you can follow up on switching to the `core::ops` traits and impl'ing `Neg` in that PR?

jellevos:
> Yeah, I've had similar thoughts but never broached them. We should probably switch to the `core::ops` traits.
> 
> cc @jellevos

Yes, in hindsight, that makes total sense. I saw the great progress you two made in the past weeks! Happy to help out if anything still needs to be taken care of regarding modular arithmetic.

PR #153, opened 2022-12-19
Add constant-time `leading_zeros()`, `trailing_zeros()`, `bits()`, and `bit()` for `Uint`

fjarri:
Is there some reason `bits_vartime()` returns `usize` and not `u32` like the corresponding methods of primitive types?


tarcieri:
Not sure... switching to `u32` seems fine to me.

cc @mikelodder7

mikelodder7:
Seems fine to me

fjarri:
Ok, I don't think it's worth it to do it in this PR, since there will be quite a few changes across the codebase. Filed #154

tarcieri (approved):

PR #155, opened 2022-12-21
Use built-in traits for `(Dyn)Resiude`

fjarri:
- Remove `AddResidue`, `SubResidue`, `MulResidue`, `InvResidue`, `PowResidue`
- Impl `Add`, `Mul`, `Sub`, and `Neg` for `Residue` and `DynResidue` (also the combinations of by-ref and by-value, and the corresponding `const fn`s)
- Add `Pow`, `Invert` and `Square` traits, and impl them for `Residue` and `DynResidue`
- Add `PowBoundedExp` (hosting the former `pow_specific()` functionality) and a generic impl of `Pow` for `Bounded` exponents via this trait
- Remove `GenericResidue` trait
- Add `Retrieve` trait
- Add `Bounded` trait hosting `BITS` and `BYTES` constants which were formerly in `Encoding`

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -227,10 +227,14 @@ pub trait Encoding: Sized {
 }
 
 /// Support for optimized squaring
-pub trait Square {
+pub trait Square: Sized
+where
+    for<'a> &'a Self: core::ops::Mul<&'a Self, Output = Self>,
+{
     /// Computes the same as `self.mul(self)`, but may be more efficient.
-    fn square(&self) -> Self;
-    // TODO: can we provide a default implementation by depending on `Mul`?
+    fn square(&self) -> Self {
+        self * self
+    }
 }
 
 /// Constant-time exponentiation.
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -202,14 +202,17 @@ pub trait Split<Rhs = Self> {
     fn split(&self) -> (Self::Output, Self::Output);
 }
 
-/// Encoding support.
-pub trait Encoding: Sized {
+/// Integers whose representation takes a bounded amount of space.
+pub trait Bounded {
     /// Size of this integer in bits.
     const BITS: usize;
 
     /// Size of this integer in bytes.
     const BYTES: usize;
+}
 
+/// Encoding support.
+pub trait Encoding: Sized {
     /// Byte array representation.
     type Repr: AsRef<[u8]> + AsMut<[u8]> + Copy + Clone + Sized;
 
@@ -241,14 +244,22 @@ where
 pub trait Pow<Exponent> {
     /// Raises to the `exponent` power.
     fn pow(&self, exponent: &Exponent) -> Self;
-    // TODO: can express via `pow_specific()` if there is a trait containing the integer's bit size
+}
 
+impl<T: PowBoundedExp<Exponent>, Exponent: Bounded> Pow<Exponent> for T {
+    fn pow(&self, exponent: &Exponent) -> Self {
+        self.pow_bounded_exp(exponent, Exponent::BITS)
+    }
+}
+
+/// Constant-time exponentiation with exponent of a bounded bit size.
+pub trait PowBoundedExp<Exponent> {
     /// Raises to the `exponent` power,
     /// with `exponent_bits` representing the number of (least significant) bits
     /// to take into account for the exponent.
     ///
     /// NOTE: `exponent_bits` may be leaked in the time pattern.
-    fn pow_specific(&self, exponent: &Exponent, exponent_bits: usize) -> Self;
+    fn pow_bounded_exp(&self, exponent: &Exponent, exponent_bits: usize) -> Self;
 }
 
 /// Constant-time inversion.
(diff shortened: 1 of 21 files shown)


jellevos:
Hey @fjarri, nice work! I just thought I would explain why I originally added `pow_specific`. One example use case is RSA, where the public exponent $e$ can be expressed in only a few bits (e.g. $65537 = 10001_2$). Perhaps there is a different way to address this particular use case.

fjarri:
@jellevos fair enough, that's a very good reason to have this functionality. I renamed `pow_specific` into `pow_bounded_exp` to be more specific (sorry for the pun), and rearranged the traits a little.

src/traits.rs
```diff
@@ -202,14 +202,17 @@ pub trait Split<Rhs = Self> {
     fn split(&self) -> (Self::Output, Self::Output);
 }
 
-/// Encoding support.
-pub trait Encoding: Sized {
+/// Integers whose representation takes a bounded amount of space.
+pub trait Bounded {
```
tarcieri, line 206: Separating out a trait for the sizes from `Encoding` seems good although I'm not sure about the name itself, but of course `Sized` is already taken by `core`.

It seems `num-traits` uses [`Bounded` to represent min and maximum values](https://docs.rs/num-traits/latest/num_traits/bounds/trait.Bounded.html) which seems like a better fit for that name.

If I can't think of anything better, `Bounded` seems OK.

src/traits.rs
```diff
@@ -225,3 +228,43 @@ pub trait Encoding: Sized {
...
+    /// NOTE: `exponent_bits` may be leaked in the time pattern.
+    fn pow_bounded_exp(&self, exponent: &Exponent, exponent_bits: usize) -> Self;
+}
+
+/// Constant-time inversion.
+pub trait Inv: Sized {
+    /// Computes the inverse. Returns [`CtOption`] that evaluates to `None`
+    /// if the number is not invertible.
+    fn inv(&self) -> CtOption<Self>;
+}
```
tarcieri, line 270: You might look to the trait design in `elliptic-curve` for this:

https://docs.rs/elliptic-curve/0.12.3/elliptic_curve/ops/trait.Invert.html

Notably that design has an associated `Output` type which avoids the `CtOption` when the value being inverted is non-zero:

https://docs.rs/elliptic-curve/0.12.3/src/elliptic_curve/scalar/nonzero.rs.html#169-181

src/traits.rs
fjarri: > It seems `num-traits` uses `Bounded` to represent min and maximum values which seems like a better fit for that name.

Yes, that's where I took the name from. Those values can be potentially added to the trait too, but I didn't want to do it in this PR. For unsigned integers `BITS` is just a different representation of `MAX`, anyway. I think `BITS` should be separate from the `Encoding` trait (since we already have `.bits()` outside of it, which is connected to `BITS`), but as for the name of the new trait, I'm open to suggestions. 

The conceptual problem here is that `pow_bounded_exp()` makes sense only for unsigned integer exponents; so technically "Bounded" is too loose of a restriction and will have to be changed if signed integers are introduced. The "theoretical" bound would be `Unsigned + Integer + Bits", with the last two possibly implied by `Unsigned`.

src/traits.rs
fjarri: So, are you suggesting to rename `Inv` to `Invert` and add the `Output` associated type? I can do that.

tarcieri: I'm fine with either name but I think it's nice to have the associated `Output` to support infallible inversions

fjarri: Done. Can't implement for `NonZero<DynResidue>` though since it does not implement `Zero`. 

Also, as a side note: would it make sense to have a `pub(crate) NonZero::new_unchecked()`, to skip the `unwrap()` in the cases like `impl Invert for NonZero<Residue>`? Or is having `unwrap()` easier to audit?

tarcieri: > Can't implement for NonZero<DynResidue> though since it does not implement Zero.

Seems like it should be able to?

> Also, as a side note: would it make sense to have a pub(crate) NonZero::new_unchecked(),

Sure

fjarri: `Zero` needs a `ZERO` constant, and we can't make that for `DynResidue` since it can have different parameters

tarcieri: Aah yeah, it would need valtrees-capable const generics at the very least, to allow passing in `DynResidueParams` at compile time. So no go for now anyway.

fjarri: > Also, as a side note: would it make sense to have a pub(crate) NonZero::new_unchecked(),

Just for certainty, do you want me to add it in this PR?

tarcieri: Yeah that's fine

fjarri: It's fine that I add it or it's fine that I don't? :) I don't think it's worth doing in this PR

tarcieri: Yeah sure, either way

src/traits.rs
tarcieri: Alright, let's go with `Bounded` for now and we can revisit it later if need be

PR #159, opened 2023-01-01
Some fixes for modular inversion

fjarri:
- Break long lines in comments and tests and fix some hardcoded bit sizes
- Fix the hardcoded `Limb` size in `inv_odd_mod()` - was set to 64 (did not cause errors, just made the inversion twice as slow on 32-bit targets)
- Added `inv_odd_mod_bounded()` for cases of argument/modulus known to be small
- Removed `inv_odd_mod_option()` - we do not provide such interface for other constant functions

Additionally:
- Introduced a `CtChoice` newtype for constant-time const fns
- Replaced some multiplications by `Word::MAX` with negations
- Normalized constant-time comparisons API in `Limb` and `Uint`: removed `ct_cmp()` (we never need it in constant-time context, `ct_gt()`/`ct_lt()`/`ct_eq()` are enough), matched const fns with `subtle` trait methods, matched methods between `Limb` and `Uint`
- Removed `SignedWord` and `SignedWideWord`
- `Uint` objects are taken by reference where they were previously taken by value.

### DIFF
--- a/src/limb.rs
+++ b/src/limb.rs
@@ -20,7 +20,7 @@ mod sub;
 #[cfg(feature = "rand_core")]
 mod rand;
 
-use crate::{Bounded, Zero};
+use crate::{Bounded, CtChoice, Zero};
 use core::fmt;
 use subtle::{Choice, ConditionallySelectable};
 
@@ -56,12 +56,6 @@ pub type WideWord = u128;
 
 /// Highest bit in a [`Limb`].
 pub(crate) const HI_BIT: usize = Limb::BITS - 1;
-
-/// A boolean value returned by constant-time `const fn`s.
-/// `Word::MAX` signifies `true`, and `0` signifies `false`.
-// TODO: should be replaced by `subtle::Choice` or `CtOption`
-// when `subtle` starts supporting const fns.
-pub type CtChoice = Word;
 
 /// Big integers are represented as an array of smaller CPU word-size integers
 /// called "limbs".
@@ -100,7 +94,7 @@ impl Limb {
     /// Return `b` if `c` is truthy, otherwise return `a`.
     #[inline]
     pub(crate) const fn ct_select(a: Self, b: Self, c: CtChoice) -> Self {
-        Self(a.0 ^ (c & (a.0 ^ b.0)))
+        Self(c.select(a.0, b.0))
     }
 }
 
--- a/src/limb.rs
+++ b/src/limb.rs
@@ -20,7 +20,7 @@ mod sub;
 #[cfg(feature = "rand_core")]
 mod rand;
 
-use crate::{Bounded, CtChoice, Zero};
+use crate::{Bounded, Zero};
 use core::fmt;
 use subtle::{Choice, ConditionallySelectable};
 
@@ -90,12 +90,6 @@ impl Limb {
     /// Size of the inner integer in bytes.
     #[cfg(target_pointer_width = "64")]
     pub const BYTES: usize = 8;
-
-    /// Return `b` if `c` is truthy, otherwise return `a`.
-    #[inline]
-    pub(crate) const fn ct_select(a: Self, b: Self, c: CtChoice) -> Self {
-        Self(c.select(a.0, b.0))
-    }
 }
 
 impl Bounded for Limb {
--- a/src/limb.rs
+++ b/src/limb.rs
@@ -73,6 +73,12 @@ pub(crate) type WideSignedWord = i128;
 /// Highest bit in a [`Limb`].
 pub(crate) const HI_BIT: usize = Limb::BITS - 1;
 
+/// A boolean value returned by constant-time `const fn`s.
+/// `Word::MAX` signifies `true`, and `0` signifies `false`.
+// TODO: should be replaced by `subtle::Choice` or `CtOption`
+// when `subtle` starts supporting const fns.
+pub type CtChoice = Word;
+
 /// Big integers are represented as an array of smaller CPU word-size integers
 /// called "limbs".
 #[derive(Copy, Clone, Debug, Default, Hash)]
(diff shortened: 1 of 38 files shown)


tarcieri:
> Tangentially related: what should we return as is_some in constant functions that can fail - Limb or Word? There are cases of both in the codebase.

It should probably always be `Word`, as `Limb` is mainly intended for use inside the inner array of a `Uint`

fjarri:
Hm, in some places we also return `u8` in {0, 1}. Which makes sense if it's converted to `Choice`, but inconvenient if it has to be chained with const fn analogues of `subtle`.

tarcieri:
Yeah, subtle's use of `u8` is a bit odd as I imagine most of the time LLVM is implicitly widening them to word-sized values sitting in a register.

I don't really have a better suggestion for `Word` -> `Choice` than casting to an intermediate `u8`.

fjarri:
> I don't really have a better suggestion for Word -> Choice than casting to an intermediate u8.

It's more of a question, whether we should return `Word` everywhere, or `u8` everywhere. Currently both are used in different places.

tarcieri:
`Word` seems like it had the most mechanical sympathy to me

fjarri:
Ok, updated to `Word` everywhere.

fjarri:
I wonder if it even makes sense to define some `type IsSome = Word` to make it more distinguishable.

tarcieri:
@fjarri sounds like a great idea!

fjarri:
More things that could be done here:
- Match the `self` usage in comparison API between `Limb` and `Uint`: `Uint` uses `self` and `rhs`, `Limb` uses `lhs` and `rhs`. The former seems more natural, the latter does not conflict with `subtle` traits.
- Make `CtChoice` an actual type, with `TRUE` and `FALSE` constants, conversion to `Choice` etc?
- Where do I put `CtChoice`? Currently it lives in `limb.rs`.
- Normalize taking `Uint` by ref or by value (I think it should be taken by ref everywhere, but there are instances where it's passed by value)

fjarri:
Just in case there's an ambiguity: those things could be done, but not necessarily in this PR since it has already overgrown its initial purpose.

src/uint/cmp.rs
```diff
@@ -25,103 +23,97 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
-            }
+        let is_lt = self.ct_lt(other);
+        let is_eq = self.ct_eq(other);
+
+        if is_lt == Word::MAX {
+            Ordering::Less
+        } else if is_eq == Word::MAX {
+            Ordering::Equal
+        } else {
+            Ordering::Greater
```
tarcieri, line 116: Seems like you could use `subtle` here (although the mixture of `ct_eq` might get a little confusing)

```suggestion
        let is_eq = self.ct_eq(other).ct_eq(Word::MAX);
        let is_gt = self.ct_gt(other).ct_eq(Word::MAX);
        
        let mut ordering = Ordering::Less;
        ordering.conditional_assign(&Ordering::Equal, is_eq);
        ordering.conditional_assign(&Ordering::Greater, is_gt);
        ordering
```

src/limb.rs
```diff
@@ -58,21 +50,19 @@ pub(crate) type WideSignedWord = i64;
...
-pub(crate) type WideSignedWord = i128;
-
 /// Highest bit in a [`Limb`].
 pub(crate) const HI_BIT: usize = Limb::BITS - 1;
 
+/// A boolean value returned by constant-time `const fn`s.
+/// `Word::MAX` signifies `true`, and `0` signifies `false`.
+// TODO: should be replaced by `subtle::Choice` or `CtOption`
+// when `subtle` starts supporting const fns.
+pub type CtChoice = Word;
```
tarcieri, line 64: Seems like this could benefit from a few constants to codify the contract?

```suggestion
pub type CtChoice = Word;

pub const CT_FALSE: CtChoice = 0;
pub const CT_TRUE: CtChoice = CtChoice::MAX;
```

src/uint/cmp.rs
fjarri: This method wasn't constant-time before the change, so I didn't think of that, but sure, this can be done

src/limb.rs
fjarri: Yes, I guess it's an intermediate solution between the current one and making `CtChoice` its own type.

src/uint/rand.rs
```diff
@@ -45,7 +45,7 @@ impl<const LIMBS: usize> RandomMod for Uint<LIMBS> {
             }
             n.limbs[n_limbs - 1] = n.limbs[n_limbs - 1] & mask;
 
-            if n.ct_lt(modulus).into() {
+            if ConstantTimeLess::ct_lt(&n, modulus).into() {
```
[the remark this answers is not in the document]
fjarri: Yes, that's what I mentioned in my last comment. We can switch to non-self methods for Uint, like Limb does it.

src/limb.rs
tarcieri: A newtype for `Word` might make sense, although maybe as a followup

src/uint/rand.rs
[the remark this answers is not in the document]
fjarri: Fixed.

src/limb.rs
fjarri: Added a newtype.

src/uint/cmp.rs
fjarri: After some thought, I think the current implementation is fine. This method is not supposed to be constant-time, so we probably shouldn't complicate it (plus its return value can't be used in a constant-time context anyway).

tarcieri (approved):

src/uint/cmp.rs
tarcieri: Turns out that `Ordering` doesn't impl `ConditionallySelectable` either: https://github.com/dalek-cryptography/subtle/pull/102

PR #163, opened 2023-01-16
Improvements to modular arithmetic support

haslersn:
- Fix the macro `impl_modulus` so that it's usable from external crates. More specifically:
  - Make `UInt::ct_reduce`, `UInt::ct_reduce_wide` and `montgomery_reduction` public, because they are used in the macro `impl_modulus`. Without this change, the macro is unusable from external crates.
  - Fix qualification of re-exported items.
- Build the `GenericResidue` type around standard Rust traits such as `Add`, `Sub`, and `Mul`, instead of inventing our own traits. This way we can use overloaded operators in code that is generic over any `GenericResidue`.
- For any `GenericResidue`, also require `Sub`, `SubAssign`, `Neg`, and `ConstantTimeEq`.
- For `Residue` (i.e. with compile-time modulus) implement `Zero` and `Random`. (For `DynResidue` we cannot implement these traits, because we'd need to pass the desired modulus.)
- Remove the `Residue` suffix from the traits `SquareResidue`, `PowResidue`, as well as `InvResidue` and move them to `src/traits.rs`, because they are generic, i.e., they can in principle be applied to more than just residues.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1,8 +1,8 @@
 //! Traits provided by this crate
 
-use crate::{Limb, NonZero};
+use crate::{Limb, NonZero, UInt, Word};
 use core::fmt::Debug;
-use core::ops::{BitAnd, BitOr, BitXor, Div, Not, Rem, Shl, Shr};
+use core::ops::{BitAnd, BitOr, BitXor, Div, Mul, Not, Rem, Shl, Shr};
 use subtle::{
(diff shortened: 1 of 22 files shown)


tarcieri:
@haslersn look like it needs a rebase

tarcieri:
Fixing the macros and removing `*Residue` from traits that aren't residue-specific

haslersn:
Superseded by #166, #167, #168, #169, #170, #171, #172

haslersn:
Superseded by #166, #167, #168, #169, #170, #171, #172

PR #167, opened 2023-01-17
Impl `ConditionallySelectable` and `ConstantTimeEq` for `DynResidue`

haslersn:
They were already implemented for `Residue`.


src/uint/modular/runtime_mod.rs
```diff
@@ -99,6 +101,27 @@ impl<const LIMBS: usize> DynResidue<LIMBS> {
     }
 }
 
+impl<const LIMBS: usize> ConditionallySelectable for DynResidue<LIMBS> {
+    fn conditional_select(a: &Self, b: &Self, choice: Choice) -> Self {
+        debug_assert_eq!(a.residue_params, b.residue_params);
```
tarcieri, line 106: Hmm, this seems tricky to deal with but also not a great solution...

src/uint/modular/runtime_mod.rs
```diff
@@ -99,6 +101,27 @@ impl<const LIMBS: usize> DynResidue<LIMBS> {
...
+        Self {
+            montgomery_form: Uint::conditional_select(
+                &a.montgomery_form,
+                &b.montgomery_form,
+                choice,
+            ),
+            residue_params: a.residue_params,
```
tarcieri, line 113: Namely if residue params aren't the same, in a release build this may result in selecting `b`'s montgomery_form with `a`'s residue_params

haslersn: I actually just created this PR for consistency, but I don't use DynResidue in my application. So feel free to close this PR if the solution is not acceptable.

tarcieri: Okay, I think some more work is needed here.

Namely this probably needs a fallible inherent method for conditional selection which returns an error in the event the residue params mismatch.

PR #172, opened 2023-01-23
Make `impl_modulus!` usable from external crates

haslersn:
Previously it was impossible to use `impl_modulus!` in crates that use crypto-bigint, because certain functions were only available within the crypto-bigint crate.

### DIFF
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -44,7 +44,7 @@ impl<const LIMBS: usize> Uint<LIMBS> {
     /// Computes `self` / `rhs`, returns the quotient (q), remainder (r)
     /// and the truthy value for is_some or the falsy value for is_none.
     ///
-    /// NOTE: Use only if you need to access const fn. Otherwise use [`div_rem`] because
+    /// NOTE: Use only if you need to access const fn. Otherwise use [`Self::div_rem`] because
     /// the value for is_some needs to be checked before using `q` and `r`.
     ///
     /// This is variable only with respect to `rhs`.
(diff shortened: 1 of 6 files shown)


src/uint/div.rs
```diff
@@ -79,12 +79,12 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
     /// and the truthy value for is_some or the falsy value for is_none.
     ///
-    /// NOTE: Use only if you need to access const fn. Otherwise use [`rem`].
+    /// NOTE: Use only if you need to access const fn. Otherwise use [`Self::rem`].
     /// This is variable only with respect to `rhs`.
     ///
     /// When used with a fixed `rhs`, this function is constant-time with respect
     /// to `self`.
-    pub(crate) const fn ct_rem(&self, rhs: &Self) -> (Self, CtChoice) {
+    pub const fn ct_rem(&self, rhs: &Self) -> (Self, CtChoice) {
```
tarcieri, line 87: Ugh, we've been trying to keep these functions out of the public API but I see they're needed to define constants.

I haven't been wild about their name. They're really "const_" as opposed to "ct_", and the wrapper functions which return `subtle`-based types like `CtOption` and `Choice` are also constant-time, which makes the `ct_` prefix confusing in the public API.

I guess we can make them public in this PR and circle back on better names.

src/uint/modular.rs
```diff
@@ -1,7 +1,7 @@
-mod reduction;
-
 /// Implements `Residue`s, supporting modular arithmetic with a constant modulus.
 pub mod constant_mod;
+/// Implements Montgomery reduction
+pub mod reduction;
```
tarcieri, line 4: Rather than making this module `pub` I think it would make more sense to re-export `montgomery_reduction` from the `modular` module

mikelodder7 (approved):
Makes sense. Nothing else better we could do than define macros to create constants?

tarcieri:
Not for now. When const generics improve we can get rid of the macros.

fjarri:
> I haven't been wild about their name. They're really "const_" as opposed to "ct_", and the wrapper functions which return subtle-based types like CtOption and Choice are also constant-time, which makes the ct_ prefix confusing in the public API.

I would vote for making the `const fn`s have `const_` prefix. They won't intersect with the `subtle`-based functions since those can't be used in the `const` context anyway. We don't need a special `ct_` prefix for constant-timeness since the library already postulates that it's constant-time by default, and anything not constant-time is specifically marked as `_vartime`. The only problem are the `const fn` analogues of `subtle` methods which have `ct_` prefixes for consistency reasons, but I think they're all currently crate-private.

tarcieri:
@haslersn can you add a test to the `tests/` directory to ensure that `impl_modulus!` works via the crate's public API?

Otherwise I'd say this is fine and we can circle back on the `ct_` prefix separately.

haslersn:
Should now be all done including the `const_` prefix:

- Added test
- Added `$crate` qualification in the `impl_modulus!` implementation so that the macro's user doesn't need additional imports
- Renamed `ct_rem{_wide}` to `const_rem{_wide}`

tarcieri:
@haslersn hmm, I don't see the test?

haslersn:
No, they were pushed just before my previous comment

tarcieri (approved):

tarcieri:
Oh sorry, didn’t refresh

PR #173, opened 2023-01-23
Support to concat `U320` and `U640`

haslersn:
- Add concat support for the `U320` and `U640` types
- Add split support for the resulting `U640` and `U1280` types
- While we're at it, remove (anyway broken) split support for types with odd number of limbs


haslersn:
I needed this change for a project of mine.

PR #193, opened 2023-03-06
Fix #191 by using a module path available from the outside of the create

### LINKED ISSUE #191: `const_residue` doesn't seem to work outside of crate
(Forgive me if this is an obvious problem based on my lack of Rust experience).

I'm trying to use the modular exp support recently added, but I'm running in to a problem with const_residue. Specifically, I do this:

```rust
impl_modulus!(PMod, U1536, P_STR);
```
where `P_STR` contains the hex string for the modulus. Then, in a test, I do:

```rust
let base = const_residue!(G1, PMod);
```

where `G1` is a big int of the same size as the modulus. However, this leads to a failure saying "module `uint` is private". Specifically:

```
error[E0603]: module `uint` is private
   --> /home/amnesia/Persistent/.cargo/registry/src/github.com-1ecc6299db9ec823/crypto-bigint-0.5.0/src/uint/modular/constant_mod/macros.rs:48:17
    |
48  |         $crate::uint::modular::constant_mod::Residue::<$modulus, { $modulus::LIMBS }>::new(
    |                 ^^^^ private module
    |
note: the module `uint` is defined here
   --> /home/amnesia/Persistent/.cargo/registry/src/github.com-1ecc6299db9ec823/crypto-bigint-0.5.0/src/lib.rs:167:1
    |
167 | mod uint;
    | ^^^^^^^^
```

olabiniV2:
(Since the old PR had bad formatting, I'm opening this instead)

### DIFF
--- a/src/uint/modular/constant_mod/macros.rs
+++ b/src/uint/modular/constant_mod/macros.rs
@@ -45,7 +45,7 @@ macro_rules! impl_modulus {
 /// For example, `residue!(U256::from(105u64), MyModulus);` creates a `Residue` for 105 mod `MyModulus`.
 macro_rules! const_residue {
     ($variable:ident, $modulus:ident) => {
-        $crate::uint::modular::constant_mod::Residue::<$modulus, { $modulus::LIMBS }>::new(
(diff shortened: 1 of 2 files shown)


tarcieri:
Thank you!

PR #203, opened 2023-03-29
Montgomery multiplication improvements

fjarri:
- Added benchmarks for Montgomery multiplication and DynResidueParams creation
- Replaced the final reduction in `montgomery_reduction()` with `sub_mod_with_hi()`. Speeds it up by ~20% (for `U256`), I guess due to better vectorization?
- Extracted `muladdcarry()` from `montgomery_reduction()`
- Simplified `Uint::sub_mod()`, `sub_mod_special()`, `add_mod()`, `add_mod_special()` by reusing existing methods.
- Made `DynResidueParams::new()` a `const fn`
- Optimized `mod_neg_inv` calculation in `DynResidueParams::new()`, speeds it up by ~10% (for `U256`).

Originally I wanted to implement Montgomery multiplication by simultaneous multiplication + reduction instead of `mul_wide`, but it showed exactly the same performance, with the exception of the last reduction part - that's how I discovered the performance improvement. Still not quite sure what causes it.

(By the way, I measured the performance on arm64 - would be interesting to see the results for x64)

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -104,10 +104,45 @@ fn bench_modpow<'a, M: Measurement>(group: &mut BenchmarkGroup<'a, M>) {
     });
 }
 
+fn bench_montgomery<'a, M: Measurement>(group: &mut BenchmarkGroup<'a, M>) {
+    const TEST_SET: usize = 10;
+    let xs = (0..TEST_SET)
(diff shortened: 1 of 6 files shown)


tarcieri:
Seems like a nice simplification. Thanks!

Issue #158, opened 2022-12-27
`num-traits` support (optional)


tarcieri:
Currently `num-traits` is only in `dev-dependencies`.

As [originally suggested here](https://github.com/RustCrypto/crypto-bigint/issues/1#issuecomment-945146204) it would be useful to have some support for traits for abstracting across bignum libraries, which would make it possible for e.g. the `rsa` crate to use either `crypto-bigint` or `num-bigint-dig`.

Though `num-traits` probably lacks all of the requisite functionality to do that, it would get us quite a bit of the way there.

ycscaly:
> Currently `num-traits` is only in `dev-dependencies`.
> 
> As [originally suggested here](https://github.com/RustCrypto/crypto-bigint/issues/1#issuecomment-945146204) it would be useful to have some support for traits for abstracting across bignum libraries, which would make it possible for e.g. the `rsa` crate to use either `crypto-bigint` or `num-bigint-dig`.
> 
> Though `num-traits` probably lacks all of the requisite functionality to do that, it would get us quite a bit of the way there.

Is the absence of `num-traits` the reason for not using crypto-bigint in `rsa`, `k256` etc.?

I am now compiling a list of all dependencies for the MPC project I am building, and obviously a bigint library is a prerequisite for many higher-level primitives (like the actual group code, for both Paillier and the elliptic curves `k256`, `ed25519` etc. that I need).

I am very careful in my analysis, as for me it is important to know exactly the building blocks both theory and implementation wise. It is important for me to not have duplication, so e.g. relying on both crypto-bigint and num-bigint is something I'd go a long way to avoid. There is enough attack surface as it is, I don't want to be exposed to double the attack surface.

Now, regarding choosing which bigint library to use, crypto-bigint seemed the obvious choice for me:
- because it is constant-time (num-bigint is not, afaik)
- it is no-std pure-rust (which is important for me for WASM, but num-bigint seems to offer it)
- it is developed by RustCrypto, which means the same developers of many other building blocks I plan to use anyway (like `k256`, and the hash functions, and maybe other tools). 

But finding out `crypto-bigint` isn't even used internally within RustCrypto made me wonder how production-ready it is, and what was the reasons behind you choosing to work with `num-bigint` (is it purely for historical reasons? num-bigint was ready before crypto-bigint was, and you need to do the switch now? will supporting `num-traits` allow that transition?)

Thanks

tarcieri:
[...]

PR #214, opened 2023-04-24
num_traits::Num, AsNaturalNumber, AsRingElement

ycscaly:
WIP: for discussion purposes only


Cargo.toml
```diff
@@ -26,6 +26,7 @@ rand_core = { version = "0.6.4", optional = true }
 rlp = { version = "0.5", optional = true, default-features = false }
 serdect = { version = "0.2", optional = true, default-features = false }
 zeroize = { version = "1", optional = true,  default-features = false }
+num-traits = "0.2"
```
tarcieri, line 29: As previously discussed in #158

```suggestion
num-traits = { version = "0.2", optional = true }
```

rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly
```
tarcieri, line 1: This crate supports `stable` rust, so making this `nightly` does not make sense

ycscaly: This was just for intellij, shouldn't have been added tomgit, ill remove it but in any case this doesn't effect the PR I just wanted to put some code here to test the idea and have a discussion around it

ycscaly:
Closing after #215  was closed

PR #218, opened 2023-04-30
Num traits

### LINKED ISSUE #158: `num-traits` support (optional)

ycscaly:
Resolves #158


src/wrapping.rs
```diff
@@ -90,6 +159,24 @@ impl<T: Serialize> Serialize for Wrapping<T> {
...
+
+    use crate::{Wrapping, U64};
+
+    fn assures_num<T: Num>(x: T) {}
+
+    #[test]
+    fn num_traits() {
+        const TEST: Wrapping<U64> = Wrapping(U64::from_u64(0x0011223344556677));
+
+        assures_num(TEST)
```
ycscaly, line 176: My logic here is: `Num` is implemented for `Wrapping<T> where ...` so that we're implementing it for the minimally restricted type, but we want to assure that it works for `Wrapping<Uint<Limbs>>`, so this code assures it.

I'm not sure if there's a better way of checking this.

src/wrapping.rs
```diff
@@ -63,6 +70,68 @@ impl<T: ConstantTimeEq> ConstantTimeEq for Wrapping<T> {
...
+    where Wrapping<T>:
+    Add<Wrapping<T>, Output = Wrapping<T>>
+    + Sub<Wrapping<T>, Output = Wrapping<T>>
+    + Mul<Wrapping<T>, Output = Wrapping<T>>
+    + Div<NonZero<T>, Output = Wrapping<T>>
+    + Rem<NonZero<T>, Output = Wrapping<T>>
+{
+    type FromStrRadixErr = ();
+
+    fn from_str_radix(str: &str, radix: u32) -> Result<Self, Self::FromStrRadixErr> {
```
ycscaly, line 130: How do we want to handle this?

Obviously this is unimplemented logic for this crate, and I'm not sure whether we want to support this (or, for example, return an error for any radix other than 2, 10 or 16.)

`num-bigint`'s [implementation](https://github.com/rust-num/num-bigint/blob/6f2b8e0fc218dbd0f49bebb8db2d1a771fe6bafa/src/biguint/convert.rs#LL220C75-L220C75) is complex, but I could copy it if that's the way to go.

My thoughts are to actually to use `from_str_radix()` implementation of [`u32`](https://doc.rust-lang.org/std/primitive.u32.html) and [`u64`](https://doc.rust-lang.org/std/primitive.u64.html), that is, to parse limb-by-limb and sum them up.

src/wrapping.rs
```diff
@@ -63,6 +70,68 @@ impl<T: ConstantTimeEq> ConstantTimeEq for Wrapping<T> {
...
+    }
+}
+
+#[cfg(feature = "num-traits")]
+impl<T: Zero> Div<Wrapping<T>> for Wrapping<T>
+    where Wrapping<T>: Div<NonZero<T>, Output = Wrapping<T>>
+{
+    type Output = Wrapping<T>;
+
+    fn div(self, rhs: Wrapping<T>) -> Self::Output {
```
ycscaly, line 103: This and `rem` panics in case of division by zero.
This is the intended behavior of this crate, but I'm not sure that "this is the right panic" - i.e. now we'll panic on `unwrap()`ping of `NonZero` on zero, and the panic message might not be ideal; should I instead `assert_ne(rhs, T::Zero)` instead?

ycscaly:
@tarcieri please see and respond to my comments so I can proceed

src/wrapping.rs
```diff
@@ -63,6 +70,70 @@ impl<T: ConstantTimeEq> ConstantTimeEq for Wrapping<T> {
     }
 }
 
+#[cfg(feature = "num-traits")]
+impl<T: Zero> num_traits::Zero for Wrapping<T>
```
tarcieri, line 74: Seems like this (and several of the other impls) can just be blanket impls like:

```
impl<T: Zero> num_traits::Zero for T
```

...size `Wrapping<T>` already impls `Zero` using this same pattern

tarcieri: Never mind, I suppose such a blanket impl wouldn't be permitted.

tarcieri:
#433 covers quite a bit of this

src/wrapping.rs
tarcieri: #246 and/or #247 are some options, although I took issue with them not being constant-time

tarcieri:
Closing in favor of #433. Please open a new PR if you would like additional support which wasn't covered there.

PR #219, opened 2023-04-30
Conversions

ycscaly:
Implements conversion across different `Uxxx` types. This allows what was previously e.g.

```rust
let u: U128 = transcript.challenge(b"u");

let u: U256 = U128::ZERO.concat(&u);
let u: U512 = U256::ZERO.concat(&u);
let u: U1024 = U512::ZERO.concat(&u);
let u: U2048 = U1024::ZERO.concat(&u);
let u: U4096 = U2048::ZERO.concat(&u);
```

To become:
```rust
let u: U128 = transcript.challenge(b"u");
let u: U4096 = u.to_u4096;
```


ycscaly:
Resolves #220

Cargo.toml
```diff
@@ -18,6 +18,7 @@ rust-version = "1.65"
 
 [dependencies]
 subtle = { version = "2.4", default-features = false }
+paste = "1.0.12"
```
tarcieri, line 21: I would prefer not to add any additional mandatory dependencies.

Also, this use case doesn't seem to warrant an optional dependency either.

tarcieri:
It seems like this would be better implemented in terms of the `From` trait.

ycscaly:
> It seems like this would be better implemented in terms of the `From` trait.

Agreed

> It seems like this would be better implemented in terms of the `From` trait.

Agreed, this also allowed me to drop the `paste` dependency.

Should be good now

ycscaly:
Replaced by #230

PR #221, opened 2023-04-30
Add `UintVec`: heap-allocated fixed-precision integers

tarcieri:
Use cases like DSA and RSA benefit from having heap-allocated integers whose precision can be chosen at runtime rather than compile time.

These algorithms are used in conjunction with varying key sizes, and when dealing with a large number of potential key sizes it's helpful not to have to monomorphize the algorithm implementation for each one.

Ideally we can define traits so that code can be written generically but used with either `Uint` or `UintVec`, allowing heapless `no_std` users to use e.g. `U2048`, while users with a heap can use `UintVec`.

It should also be possible to make some of the existing algorithm implementations generic over `Uint` vs `UintVec`.

This initial implementation only provides addition and comparison support, but also the initial scaffolding for adding more operations.

### DIFF
--- a/src/uint_vec/cmp.rs
+++ b/src/uint_vec/cmp.rs
@@ -1,0 +1,47 @@
+//! [`UintVec`] comparisons.
+//!
+//! By default these are all constant-time and use the `subtle` crate.
+
+use super::UintVec;
+use crate::Limb;
+use subtle::{Choice, ConstantTimeEq};
+
+impl ConstantTimeEq for UintVec {
+    #[inline]
+    fn ct_eq(&self, other: &Self) -> Choice {
+        let (shorter, longer) = Self::sort_by_precision(self, other);
+        let mut ret = Choice::from(1u8);
+
+        for i in 0..longer.limbs.len() {
+            let a = shorter.limbs.get(i).unwrap_or(&Limb::ZERO);
+            let b = longer.limbs.get(i).unwrap_or(&Limb::ZERO);
+            ret &= a.ct_eq(b);
+        }
+
+        ret
+    }
+}
+
+impl Eq for UintVec {}
+impl PartialEq for UintVec {
+    fn eq(&self, other: &Self) -> bool {
+        self.ct_eq(other).into()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::UintVec;
+    use subtle::ConstantTimeEq;
+
+    #[test]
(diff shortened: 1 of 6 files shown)


tarcieri:
While `UintVec` may at least communicate it's heap-allocated, I'm not sure it's a good name.

Maybe `DynUint` would be better?

src/uint_vec/cmp.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl ConstantTimeEq for UintVec {
+    #[inline]
+    fn ct_eq(&self, other: &Self) -> Choice {
+        let (shorter, longer) = Self::sort_by_precision(self, other);
+        let mut ret = Choice::from(1u8);
+
+        for i in 0..longer.limbs.len() {
+            let a = shorter.limbs.get(i).unwrap_or(&Limb::ZERO);
+            let b = longer.limbs.get(i).unwrap_or(&Limb::ZERO);
```
tarcieri, line 17: This could probably use a comment that it isn't constant-time with respect to the precision of integers, only their values.

dignifiedquire:
would be really useful to have benchmarks against gmp and num-bigint-dig for various ops and sizes, even if slower, just so we can orientate things

tarcieri:
Definitely want to add benchmarks, especially to play around with optimizing the carry chain function.

@dignifiedquire I didn't think about competitive benchmarks with other crates yet though... that's a good idea!

fjarri:
> Maybe DynUint would be better?

Or `HeapUint`?

andrewwhitehead:
`AllocUint`? `HeapUint` wouldn't make sense if custom allocator support was added

tarcieri:
I guess a problem with `DynUint` is what would we call the corresponding `Residue`/`DynResidue` types?

So something other than `Dyn` seems good. `Heap` is explicit but eventual custom allocator support is a good point.

In interest of keeping it short and familiar, perhaps `BigUint`, although I'm used to `Big` meaning arbitrary precision instead of dynamic fixed precision.

Naming is hard!

tarcieri:
Some other options: `BoxUint` or `BoxedUint`

tarcieri:
I went with `BoxedUint`, which I think is sufficiently clear. `BoxedResidue` is nicely distinct from `DynResidue`.

Going to merge this despite the limited functionality to have a basic scaffold in place. Additional functionality can be added in independent followup PRs.

PR #224, opened 2023-05-03
ConcatOther<Rhs> for cross-sizes concatenation of Uint<>s

### LINKED ISSUE #225: Concatenation of different sizes Uint<>s
Another missing cross-size functionality is concatenation. It is also a requirement to implement cross-sizes multiplication

ycscaly:
Resolves #225


ycscaly:
Resolves #225

ycscaly:
@tarcieri

src/uint.rs
```diff
@@ -388,6 +388,265 @@ impl_concat! {
...
+    (U1536, 1536),
+    (U1792, 1792),
+    (U2048, 2048),
+    (U3072, 3072),
+    (U3584, 3584),
+    (U4096, 4096),
+    (U6144, 6144),
+    (U8192, 8192)
+    )
+}
```
tarcieri, line 413: Can you move these out of `uint.rs` into their own submodule? It makes the uint.rs file huge, and if you move them to their own module, you can feature-gate them in one place instead of every time the macro is invoked

ycscaly: Yes, I thought it is better this way anyways, but the original code had it here so I followed through.

Can I put them in concat?

ycscaly: Or do you want a seperate "cross-size" module

tarcieri: `concat` works, although maybe put them in an inline submodule so you can gate them in a single place, e.g. `mod cross_size`

Cargo.toml
```diff
@@ -43,6 +43,7 @@ default = ["rand"]
 alloc = ["serdect?/alloc"]
 rand = ["rand_core/std"]
 serde = ["dep:serdect"]
+cross-size = []
```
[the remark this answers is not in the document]
ycscaly: as you wish. I just wanted to reflect the nature of rhe feature, i.e. what's missing with your suggestion is what trait are implemented in this feature.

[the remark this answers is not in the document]
tarcieri: `cross-size` without context doesn't mean anything. It doesn't even reflect what's changed by enabling the feature is additional trait impls.

This also leaves the door open to using it for other trait impls which would impact compile times.

src/uint.rs
tarcieri: I'm confused what you can't do? Regardless this is hundreds and hundreds of lines of feature-gated macro invocations which should find a better home than the toplevel module.

ycscaly: I can create a separate module for all of this then. Will be easier.

Cargo.toml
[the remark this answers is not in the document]
ycscaly: Ok, did this for #224, #219, and #226. Do you want it to be kept as three separate PRs or merged into a single one?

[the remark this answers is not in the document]
tarcieri: A single PR would probably be best

ycscaly:
Replaced by #230

PR #226, opened 2023-05-03
Cross-sizes wide multiplication & Mul trait implementation

### LINKED ISSUE #209: Multiplying U2048 by U2048 to get U4096
I am preparing to write code of (threshold) Paillier using `crypto-bigint`, and was wondering how to go about operations between different bit sizes. I understand that in `crypto-bigint` the size of the integer is fixed at compilation, however it should be possible to transition between types - so that multiplying two numbers of a given size gives you a number double that size.

However, I have not found any way to do that currently. This is a sketch of what I figured out was possible with the current API:
```
let P = crypto_primes::prime::<16>(1024);
let Q = crypto_primes::prime::<16>(1024);
let (N_low, N_high) = P.mul_wide(&Q);
let N_limbs: [Limb; 32] = concat_arrays!(N_low.to_limbs(), N_high.to_limbs());
let N = U2048::from(N_limbs);

let N2_low, N2_high = N.mul_wide(&N);
let N2_limbs: [Limb; 64] = concat_arrays!(N2_low.to_limbs(), N2_high.to_limbs());
let N2 = U4096::from(N2_limbs);
```
Is this the right approach? can we think about adding appropriate functions?

ycscaly:
This implements `Mul<Self, Output = Self::Concat>` for all types as well as `Mul<Rhs, Output = Uint<{nlimbs!($self_bits) + nlimbs!($rhs_bits)}>>` for all types.

This is possible because `mul_wide` now takes a second generic paramater:
```rust
pub const fn mul_wide<const RHS_LIMBS: usize>(
        &self,
        rhs: &Uint<RHS_LIMBS>,
    ) -> (Uint<RHS_LIMBS>, Self);
```

One could debate whether this is a breaking change, for this changes the signature of a public function, but I'd argue against that, for Rust allows type inference in a way that simplifies down to the exact same signature for the special case where `Rhs = Self` which was the case for `mul_wide()` before this change. 

This means that I did not need to change any code that used `mul_wide`, and for example, `U64::ONE.mul_wide(&U64::ZERO)` compiles without needing to specify the generic parameters.

In order to implement this, I needed to also implement cross-sizes concatenation, which is why I already merged #224 into this PR, so it should be reviewed & approved before this one.

Resolves #209


tarcieri:
This feels like a combinatorial explosion. Are you checking how it impacts compile times?

tarcieri:
> On my PC it's an increase from 1.47s to 4.98s

Oof, so a ~350% increase.

This really feels like the sort of problem it would be worth waiting for const generics to properly address, rather than providing a "brute force" solution that adds hundreds (thousands?) or trait impls.

ycscaly:
> > On my PC it's an increase from 1.47s to 4.98s
> 
> Oof, so a ~350% increase.
> 
> This really feels like the sort of problem it would be worth waiting for const generics to properly address, rather than providing a "brute force" solution that adds hundreds (thousands?) or trait impls.

How would const-generics properly address this? 

I am writing my library now, and if I won't have these features I'd have to fork which I really am trying my best to avoid here. I prefer having a temporary solution that will be replaced in the future rather than having nothing.

Would feature-gating this be an acceptable compromise?

tarcieri:
I don't see how exporting the macro would help. It's adding `From` impls on types defined in `crypto-bigint`. Those aren't useful outside the crate.

src/traits.rs
```diff
@@ -198,6 +198,18 @@ pub trait Concat<Rhs = Self> {
...
+/// Concatenate two numbers into a value of a size equaling the sum of their sizes, using the `rhs`
+/// value as the least significant value.
+pub trait ConcatOther<Rhs> {
+    /// Concatenated output: sum of the widths of `Self` and `Rhs`.
+    type Output;
+
+    /// Concatenate the two values, with `self` as most significant and `rhs`
+    /// as the least significant.
+    fn concat(&self, rhs: &Rhs) -> Self::Output;
+}
```
tarcieri, line 211: Why are you defining a separate trait for this? Especially one with the exact same shape as the existing `Concat` trait

ycscaly: Because I can't use Concat as it specifies `Rhs = Self`.

tarcieri: The correct syntax is `Concat<U128>`

ycscaly: No, but `Concat` defines `rhs` to be `Self`:
```rust
pub trait Concat<Rhs = Self> {
    /// Concatenated output: twice the width of `Self`.
    type Output;

    /// Concatenate the two values, with `self` as most significant and `rhs`
    /// as the least significant.
    fn concat(&self, rhs: &Self) -> Self::Output;
}
```

Changing the method signature to take `rhs: &Rhs` does allow me to dipose of `ConcatOther`, but I'm not sure whether this is a breaking change or not. I pushed a commit that does this, let me know whether you want this or should I revert.

tarcieri: Okay, that's a bug in the definition of `Concat`.

Changing it should be fine.

ycscaly:
Replaced by #230

PR #229, opened 2023-05-07
extra_sizes implemented

### LINKED ISSUE #228: Add missing Uint<> types/sizes
In my code, I am using types that represent the multiplication of two different already existing types (e.g. `U4096` and `U256`). However, their concatenated type is missing (e.g. `U4352`).

I could open a PR that adds just the types that are missing for my specific use-case; however, I thought the fact I needed to add these types would mean sometime in the future another consumer would need to add their own missing types for similar reasons etc. and it would be better to solve it once-and-for-all, if possible.

So my actual suggestion is to simply add all sizes between 64-bit and 8192-bit that are multiplication of the limb size, i.e. 64-bit (U64, U128, ... U8128 , U8192). @tarcieri thoughts?

ycscaly:
Resolves #228


src/uint.rs
```diff
@@ -335,6 +335,139 @@ macro_rules! impl_uint_aliases {
      };
 }
 
+#[cfg(feature = "extra-sizes")]
```
ycscaly, line 338: This is getting excessive, and I wouldn't want this in `uint.rs`. However, I'm debating between two options:
1. create an `extra-sizes` module and put it there, and keep the original sizes in `uint.rs`
2. create an e.g. `sizes` module for all of these definitions

tarcieri: I think it's fine for now but there's another problem which I'll add separate line notes for

src/uint.rs
```diff
@@ -335,6 +335,139 @@ macro_rules! impl_uint_aliases {
...
+    (U7808, 7808, "7808-bit"),
+    (U7872, 7872, "7872-bit"),
+    (U7936, 7936, "7936-bit"),
+    (U8000, 8000, "8000-bit"),
+    (U8064, 8064, "8064-bit"),
+    (U8128, 8128, "8128-bit"),
+    (U8192, 8192, "8192-bit")
+}
+
+#[cfg(not(feature = "extra-sizes"))]
```
tarcieri, line 470: Instead of having a `not`, `extra-sizes` could be purely additive.

That would also take care of the `doc_auto_cfg` documentation, so the extra sizes are appropriately tagged as needing the feature.

ycscaly: Problem is, the original set of supported sizes is large and looks abit "random" IMO. If, e.g., we'd have `U64, U128, U256, U512, U1024, U2048, U4096, U8192` as the types supported by default and everything else requiring `extra_sizes`, this would make sense. However, I don't think we can do that can we? (I'm not sure if feature-gating a non-feature gated type is considered a break of SemVer or not)

Otherwise, with the way things are currently, this becomes a little grotesque. However, if you'd still prefer that, I'll make the change. Remember however, that these distinctions will become harder with the `extra_trait_impls` macros.

ycscaly: (Your suggestion was my original line of thought and I chose against it because of the aforementioned reasoning)

tarcieri: I think one way or another the types which require a feature need to be annotated with the feature name in the rustdoc.

If you simply define them in a module which is feature-gated, `doc_auto_cfg` should take care of that automatically.

ycscaly: Can I keep only `U64, U128, U256, U512, U1024, U2048, U4096, U8192` outside the feature-gate, or should I stick with the current set?

ycscaly: Should be good now

tarcieri (approved):
Looks better now, thanks.

Still a lot in `uint.rs`, and I'm not sure the feature-gated sizes will be properly reflected in the rustdoc, but I can try fixing that in a followup.

ycscaly:
> Looks better now, thanks.
> 
> Still a lot in `uint.rs`, and I'm not sure the feature-gated sizes will be properly reflected in the rustdoc, but I can try fixing that in a followup.

Thanks!

PR #238, opened 2023-05-18
Add cmp_vartime, faster Eq for Uint

andrewwhitehead:
This optimizes Eq for Uint to avoid comparing the two values twice. `cmp_vartime` is added for cases where a short-circuiting comparison is sufficient.


tarcieri (approved):
Looks good, thanks!

Issue #232, opened 2023-05-08
Residue has improper behavior with even moduli


kayabaNerve:
```rs
        const limbs: usize = 256 / Limb::BITS;
        let params: DynResidueParams<limbs> = DynResidueParams::new(&U256::from_u8(10u8));
        for i in 0..10 {
            let x = DynResidue::new(&U256::from_u8(i as u8), params);
            println!("{:?}", x.retrieve());
        }
```

This should print 0 ..= 9. Instead,

```
Uint(0x0000000000000000000000000000000000000000000000000000000000000000)
Uint(0x0000000000000000000000000000000000000000000000000000000000000004)
Uint(0x0000000000000000000000000000000000000000000000000000000000000000)
Uint(0x0000000000000000000000000000000000000000000000000000000000000007)
Uint(0x0000000000000000000000000000000000000000000000000000000000000003)
Uint(0x0000000000000000000000000000000000000000000000000000000000000000)
Uint(0x0000000000000000000000000000000000000000000000000000000000000006)
Uint(0x0000000000000000000000000000000000000000000000000000000000000002)
Uint(0x0000000000000000000000000000000000000000000000000000000000000009)
Uint(0x0000000000000000000000000000000000000000000000000000000000000000)
```

This behavior also exists with the non-dyn form.

https://docs.rs/crypto-bigint/latest/crypto_bigint/macro.impl_modulus.html does not document any bounds on the modulus.
https://docs.rs/crypto-bigint/latest/crypto_bigint/modular/constant_mod/struct.Residue.html does not either.
https://docs.rs/crypto-bigint/latest/crypto_bigint/modular/constant_mod/trait.ResidueParams.html does.
https://docs.rs/crypto-bigint/latest/crypto_bigint/modular/runtime_mod/struct.DynResidue.html does.
https://docs.rs/crypto-bigint/latest/crypto_bigint/modular/runtime_mod/struct.DynResidueParams.html doesn't.

All of these should document all bounds on the moduli. Additionally, DynResidueParams::new should return a Result, or at least panic on an invalid moduli to prevent the above behavior.

ycscaly:
I also get similarly weird behaviors with `DynResidue`:
```rust
[...]

PR #240, opened 2023-05-22
Enforce valid modulus for `DynResidueParams`

AaronFeickert:
As noted in #232, it is possible to create `DynResidueParams` with an invalid modulus. If this happens, arithmetic operations will quietly produce invalid results.

At the [suggestion](https://github.com/RustCrypto/crypto-bigint/issues/232#issuecomment-1557420665) of @tarcieri, this PR takes steps toward enforcing modulus validity. It has `DynResidueParams::new` panic on an invalid modulus, which keeps its signature intact while safely failing, and deprecates it. It also introduces `DynResidueParams::new_checked`, which returns a `CtOption`. It also adds a few simple tests and updates the documentation.

At the next breaking release, the checked behavior can move to `DynResidueParams::new` if desired.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -74,7 +74,7 @@ fn bench_division<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
 }
 
 fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
-    let params = DynResidueParams::new_checked(&(U256::random(&mut OsRng) | U256::ONE)).unwrap();
+    let params = DynResidueParams::new(&(U256::random(&mut OsRng) | U256::ONE));
     group.bench_function("multiplication, U256*U256", |b| {
         b.iter_batched(
             || {
@@ -88,7 +88,7 @@ fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
     });
 
     let m = U256::random(&mut OsRng) | U256::ONE;
-    let params = DynResidueParams::new_checked(&m).unwrap();
+    let params = DynResidueParams::new(&m);
     group.bench_function("modpow, U256^U256", |b| {
         b.iter_batched(
             || {
@@ -107,12 +107,12 @@ fn bench_montgomery_conversion<M: Measurement>(group: &mut BenchmarkGroup<'_, M>
     group.bench_function("DynResidueParams creation", |b| {
         b.iter_batched(
             || U256::random(&mut OsRng) | U256::ONE,
-            |modulus| DynResidueParams::new_checked(&modulus).unwrap(),
+            |modulus| DynResidueParams::new(&modulus),
             BatchSize::SmallInput,
         )
     });
 
-    let params = DynResidueParams::new_checked(&(U256::random(&mut OsRng) | U256::ONE)).unwrap();
+    let params = DynResidueParams::new(&(U256::random(&mut OsRng) | U256::ONE));
     group.bench_function("DynResidue creation", |b| {
         b.iter_batched(
             || U256::random(&mut OsRng),
@@ -121,7 +121,7 @@ fn bench_montgomery_conversion<M: Measurement>(group: &mut BenchmarkGroup<'_, M>
         )
     });
 
-    let params = DynResidueParams::new_checked(&(U256::random(&mut OsRng) | U256::ONE)).unwrap();
+    let params = DynResidueParams::new(&(U256::random(&mut OsRng) | U256::ONE));
     group.bench_function("DynResidue retrieve", |b| {
         b.iter_batched(
             || DynResidue::new(&U256::random(&mut OsRng), params),
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -73,6 +73,7 @@ fn bench_division<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
     });
 }
 
+#[allow(deprecated)]
 fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
     let params = DynResidueParams::new(&(U256::random(&mut OsRng) | U256::ONE));
     group.bench_function("multiplication, U256*U256", |b| {
@@ -103,6 +104,7 @@ fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
     });
 }
 
+#[allow(deprecated)]
 fn bench_montgomery_conversion<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
     group.bench_function("DynResidueParams creation", |b| {
         b.iter_batched(
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -73,9 +73,8 @@ fn bench_division<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
     });
 }
 
-#[allow(deprecated)]
 fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
-    let params = DynResidueParams::new(&(U256::random(&mut OsRng) | U256::ONE));
+    let params = DynResidueParams::new_checked(&(U256::random(&mut OsRng) | U256::ONE)).unwrap();
     group.bench_function("multiplication, U256*U256", |b| {
         b.iter_batched(
             || {
@@ -89,7 +88,7 @@ fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
     });
 
     let m = U256::random(&mut OsRng) | U256::ONE;
-    let params = DynResidueParams::new(&m);
+    let params = DynResidueParams::new_checked(&m).unwrap();
     group.bench_function("modpow, U256^U256", |b| {
         b.iter_batched(
             || {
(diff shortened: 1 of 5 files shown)


src/uint/modular/runtime_mod.rs
```diff
@@ -37,8 +37,18 @@ pub struct DynResidueParams<const LIMBS: usize> {
...
+        note = "This will return an `Option` in a future version to account for an invalid modulus, but for now will panic if this happens. Consider using `new_checked` until then."
+    )]
+    /// Instantiates a new set of `ResidueParams` representing the given `modulus`, which _must_ be odd.
+    /// If `modulus` is not odd, this function will panic; use [`new_checked`] if you want to be able to detect an invalid modulus.
     pub const fn new(modulus: &Uint<LIMBS>) -> Self {
+        // The modulus must be odd
+        if modulus.ct_is_odd().to_u8() != 1 {
+            panic!("Modulus must be odd.");
+        }
+
```
tarcieri, line 51: You should be able to do something like:

```suggestion
        match Self::new_checked(modulus) {
            Some(ret) => ret,
            None => panic!("modulus must be odd")
        }
```

src/uint/modular/runtime_mod.rs
```diff
@@ -37,8 +37,18 @@ pub struct DynResidueParams<const LIMBS: usize> {
 }
 
 impl<const LIMBS: usize> DynResidueParams<LIMBS> {
-    /// Instantiates a new set of `ResidueParams` representing the given `modulus`.
+    #[deprecated(
+        since = "0.5.3",
+        note = "This will return an `Option` in a future version to account for an invalid modulus, but for now will panic if this happens. Consider using `new_checked` until then."
+    )]
```
tarcieri, line 43: Please place the attribute after the comment

src/uint/modular/runtime_mod.rs
AaronFeickert: Done, thanks

src/uint/modular/runtime_mod.rs
AaronFeickert: Done.

benches/bench.rs
```diff
@@ -73,6 +73,7 @@ fn bench_division<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
     });
 }
 
+#[allow(deprecated)]
```
fjarri, line 76: Why not change it to use the non-deprecated method?

src/uint/modular/runtime_mod.rs
```diff
@@ -37,25 +37,48 @@ pub struct DynResidueParams<const LIMBS: usize> {
...
-        Self {
-            modulus: *modulus,
-            r,
-            r2,
-            r3,
-            mod_neg_inv,
+    /// Instantiates a new set of `ResidueParams` representing the given `modulus` if it is odd.
+    /// Returns an `Option` that is `None` if the provided modulus is not odd; this is a safer version of [`new`][`DynResidueParams::new`], which can panic.
+    pub const fn new_checked(modulus: &Uint<LIMBS>) -> Option<Self> {
+        match modulus.ct_is_odd().to_u8() {
```
fjarri, line 56: I feel like it would be easier to exit early if `modulus` is even, to avoid having most of the function body double-indented.

src/uint/modular/runtime_mod.rs
```diff
@@ -37,25 +37,48 @@ pub struct DynResidueParams<const LIMBS: usize> {
...
 
-        Self {
-            modulus: *modulus,
-            r,
-            r2,
-            r3,
-            mod_neg_inv,
+    /// Instantiates a new set of `ResidueParams` representing the given `modulus` if it is odd.
+    /// Returns an `Option` that is `None` if the provided modulus is not odd; this is a safer version of [`new`][`DynResidueParams::new`], which can panic.
+    pub const fn new_checked(modulus: &Uint<LIMBS>) -> Option<Self> {
```
fjarri, line 55: @tarcieri , just want to make sure: what is the general strategy for propagating `CtChoice`? Should this return `(Self, CtChoice)` instead?

benches/bench.rs
AaronFeickert: Certainly could. The intent was to avoid touching existing code unnecessarily, but if it's better to take care of it all now, will do.

src/uint/modular/runtime_mod.rs
AaronFeickert: Sure, will do.

src/uint/modular/runtime_mod.rs
AaronFeickert: AIUI we're operating on the implicit assumption that moduli are effectively public (as in #235), so this is really just using the existing parity check functionality for simplicity and not for any timing properties.

tarcieri: I was wondering if perhaps it should return `CtOption`, but then it couldn't be `const fn`. Though it seems like all `const fn` use cases could just use `Residue` instead of `DynResidue`.

Then again, I'm not sure what cryptosystem uses a secret modulus, so I'm not sure there's a valid use case.

AaronFeickert: Preference on how to proceed?

tarcieri: `CtOption` would probably better keep with this crate's design philosophy of constant time by default, but admittedly I don't have a use case that would demand it.

AaronFeickert: Will do. This means we can't neatly wrap it for `new`, but such is life.

src/uint/modular/runtime_mod.rs
```diff
@@ -59,6 +71,33 @@ impl<const LIMBS: usize> DynResidueParams<LIMBS> {
         }
     }
 
+    /// Instantiates a new set of `ResidueParams` representing the given `modulus` if it is odd.
+    /// Returns a `CtOption` that is `None` if the provided modulus is not odd; this is a safer version of [`new`][`DynResidueParams::new`], which can panic.
+    pub fn new_checked(modulus: &Uint<LIMBS>) -> CtOption<Self> {
```
fjarri, line 76: Could this just call `new()`, so that we don't have to repeat all the parameter-generating code?

src/uint/modular/runtime_mod.rs
```diff
@@ -194,3 +233,36 @@ impl<const LIMBS: usize> zeroize::Zeroize for DynResidue<LIMBS> {
...
+    // Test that a valid modulus yields `DynResidueParams`
+    fn test_valid_modulus() {
+        let valid_modulus = Uint::<LIMBS>::from(3u8);
+
+        DynResidueParams::<LIMBS>::new_checked(&valid_modulus).unwrap();
+        DynResidueParams::<LIMBS>::new(&valid_modulus);
+    }
+
+    #[test]
+    #[should_panic]
```
fjarri, line 255: A small nitpick, but this method can be tested without making it panic, since the panic currently originates from inside the test, not from inside the method.

src/uint/modular/runtime_mod.rs
AaronFeickert: I don't think one constructor can neatly wrap the other, but I just pushed a commit that has both constructors use a common helper function to do the heavy lifting. Let me know what you think.

fjarri: I think this is fine, we will be removing `new_checked()` in 0.6 anyway.

src/uint/modular/runtime_mod.rs
AaronFeickert: Done.

fjarri (approved):

src/uint/modular/runtime_mod.rs
tarcieri: Yeah exactly, a const constructor for a modulus which is intended to vary at runtime is a bit pointless.

We can just ditch it in the next release and rename `new_checked` to `new`

src/uint/modular/runtime_mod.rs
```diff
@@ -59,6 +61,28 @@ impl<const LIMBS: usize> DynResidueParams<LIMBS> {
         }
     }
 
+    /// Instantiates a new set of `ResidueParams` representing the given `modulus`, which _must_ be odd.
+    /// If `modulus` is not odd, this function will panic; use [`new_checked`][`DynResidueParams::new_checked`] if you want to be able to detect an invalid modulus.
+    #[deprecated(
```
tarcieri, line 66: On second thought, I think a deprecation here might be unhelpful.

We want people to use `new` in the next release, and getting them to rename to `new_checked` probably won’t help there.

So maybe they should just keep `new` for now and update the signature in the next release.

AaronFeickert: Should we keep `new_checked` in this release anyway, to give users the `CtOption` functionality until the signature change? If so, should `new_checked` be immediately deprecated to provide information about the change?

tarcieri: We could potentially keep it as a deprecated alias for `new` in the next breaking release.

Otherwise yes, the purpose of giving it a different name is so we can get it into this release.

AaronFeickert: Sure thing. I'll move the deprecation to `new_checked` instead, so users who want more feedback on modulus validity can have that functionality right away. Will do the same in #243.

tarcieri (approved):

PR #243, opened 2023-05-25
Enforce valid modulus for `Residue` and associated macros

AaronFeickert:
This PR follows in the footsteps of #240 and enforces a valid (odd) modulus for compile-time constructions.

It updates `impl_modulus!` to panic on an invalid modulus.

For `Residue`, it adds a panic to `new` and deprecates it, and adds a `new_checked` constructor that returns a `CtOption` based on the modulus validity. This is added since it's possible to define a modulus without `impl_modulus!` and use `Residue` directly.

Since `const_residue!` uses `Residue::new` internally, this macro will also panic on an invalid modulus. Unfortunately, we can't use `new_checked` since it isn't `const`. There also doesn't appear to be a way to silence the deprecation warning since `#[allow(deprecated)]` has no effect in the macro. Boo.


AaronFeickert:
@fjarri: Would appreciate a review if you're available.

src/uint/modular/constant_mod.rs
```diff
@@ -99,6 +109,23 @@ impl<MOD: ResidueParams<LIMBS>, const LIMBS: usize> Residue<MOD, LIMBS> {
         }
     }
 
+    /// Instantiates a new `Residue` that represents this `integer` mod `MOD` if the modulus is odd.
+    /// Returns a `CtOption` that is `None` if the provided modulus is not odd; this is a safer version of [`new`][`Residue::new`], which can panic.
+    pub fn new_checked(integer: &Uint<LIMBS>) -> CtOption<Self> {
```
fjarri, line 114: I think the same thing as in the DynResidue PR is applicable - can we have a single constructor to avoid repeating the logic?

src/uint/modular/constant_mod.rs
```diff
@@ -176,6 +203,7 @@ impl<MOD: ResidueParams<LIMBS>, const LIMBS: usize> Zero for Residue<MOD, LIMBS>
 }
 
 #[cfg(feature = "rand_core")]
+#[allow(deprecated)]
```
fjarri, line 206: `new_checked()` can be used here

src/uint/modular/constant_mod/macros.rs
```diff
@@ -14,8 +15,16 @@ macro_rules! impl_modulus {
                 $crate::Concat<Output = $crate::Uint<DLIMBS>>,
         {
             const LIMBS: usize = { $crate::nlimbs!(<$uint_type>::BITS) };
-            const MODULUS: $crate::Uint<{ $crate::nlimbs!(<$uint_type>::BITS) }> =
-                <$uint_type>::from_be_hex($value);
+            const MODULUS: $crate::Uint<{ $crate::nlimbs!(<$uint_type>::BITS) }> = {
+                let res = <$uint_type>::from_be_hex($value);
+
+                // Check that the modulus is odd
+                if (res.as_limbs()[0].0 & 1) == 0 {
```
fjarri, line 22: It's interesting that `clippy` doesn't complain here about the unnecessary parentheses. Is it because it's in a macro?

AaronFeickert: Oh, possibly. I can remove it.

src/uint/modular/constant_mod.rs
AaronFeickert: I don't think it can be, without changing the trait definition in a breaking way.

src/uint/modular/constant_mod.rs
AaronFeickert: Sure can!

AaronFeickert:
@fjarri: I'll wait to address your review comments until we decide on a common deprecation strategy, as suggested by @tarcieri in #240.

src/uint/modular/constant_mod.rs
```diff
@@ -88,7 +88,17 @@ impl<MOD: ResidueParams<LIMBS>, const LIMBS: usize> Residue<MOD, LIMBS> {
     };
 
     /// Instantiates a new `Residue` that represents this `integer` mod `MOD`.
+    /// If the modulus represented by `MOD` is not odd, this function will panic; use [`new_checked`][`Residue::new_checked`] if you want to be able to detect an invalid modulus.
+    #[deprecated(
+        since = "0.5.3",
+        note = "This will return an `Option` in a future version to account for an invalid modulus, but for now will panic if this happens. Consider using `new_checked` until then."
+    )]
     pub const fn new(integer: &Uint<LIMBS>) -> Self {
```
tarcieri, line 96: Unlike `DynResidue`, I don't think we need `new_checked` on `Residue`, and we definitely shouldn't deprecate it as having a const constructor is important.

Instead, I think we can lean on the compile-time checks in `impl_modulus!`, along with asserts to catch anyone trying to construct `ResidueParams` by hand.

AaronFeickert: > having a const constructor is important

Very good point. Will update.

src/uint/modular/constant_mod/macros.rs
kayabaNerve: They aren't considered unnecessary since they don't wrap the complete expression. clippy allows parentheses, even if their lack of placement forms an equivalent expr, so long as it splits a larger term.

fjarri (approved):

src/uint/modular/constant_mod.rs
tarcieri: Okay, fine, we can do that for now since we can't make invalid `ResidueParams` 100% unrepresentable yet without `generic_const_exprs`.

Can you at least add a TODO to remove `new_checked` when we can? We should really lean on compile time checks and the type system for this ASAP. Using runtime checks is a hack because we can't yet rule this case out with the type system.

src/uint/modular/constant_mod.rs
```diff
@@ -99,6 +99,27 @@ impl<MOD: ResidueParams<LIMBS>, const LIMBS: usize> Residue<MOD, LIMBS> {
...
+        if MOD::MODULUS.ct_is_odd().to_u8() == 0 {
+            panic!("modulus must be odd");
+        }
+
+        Self::generate_residue(integer)
+    }
+
+    /// Instantiates a new `Residue` that represents this `integer` mod `MOD` if the modulus is odd.
+    /// Returns a `CtOption` that is `None` if the provided modulus is not odd; this is a safer version of [`new`][`Residue::new`], which can panic.
+    pub fn new_checked(integer: &Uint<LIMBS>) -> CtOption<Self> {
```
tarcieri, line 115:
```suggestion
    /// Instantiates a new `Residue` that represents this `integer` mod `MOD` if the modulus is odd.
    /// Returns a `CtOption` that is `None` if the provided modulus is not odd; this is a safer version of [`new`][`Residue::new`], which can panic.
    // TODO: remove this method when we can use `const_evaluatable_checked` to ensure the modulus is
    // always valid.
    pub fn new_checked(integer: &Uint<LIMBS>) -> CtOption<Self> {
```

src/uint/modular/constant_mod.rs
AaronFeickert: Agreed, and done.

tarcieri: Hmm, now that I think about it, we can probably add a const-friendly assertion that the modulus is odd which will be evaluated at compile-time, and therefore obviate the need to have a runtime-checked version.

Let me make an inline suggestion.

src/uint/modular/constant_mod.rs
```diff
@@ -99,6 +99,29 @@ impl<MOD: ResidueParams<LIMBS>, const LIMBS: usize> Residue<MOD, LIMBS> {
         }
     }
 
+    /// Instantiates a new `Residue` that represents this `integer` mod `MOD`.
+    /// If the modulus represented by `MOD` is not odd, this function will panic; use [`new_checked`][`Residue::new_checked`] if you want to be able to detect an invalid modulus.
+    pub const fn new(integer: &Uint<LIMBS>) -> Self {
+        // A valid modulus must be odd
+        if MOD::MODULUS.ct_is_odd().to_u8() == 0 {
+            panic!("modulus must be odd");
+        }
```
tarcieri, line 108: Absent inline const (which might be worth leaving a TODO comment about) this will force this check into a const context, evaluating it at compile time:

```suggestion
        const _: () = if MOD::MODULUS.ct_is_odd().to_u8() == 0 {
            panic!("modulus must be odd");
        };
```

...and with a compile-time check as to the validity of the modulus, I don't think we need `new_checked` since there will be no panic condition.

tarcieri: Aah, unfortunate

tarcieri:
Ok, I guess I'll go ahead and merge this, although I'd really, really like to find a way to get rid of `new_checked`

AaronFeickert:
> Ok, I guess I'll go ahead and merge this, although I'd really, really like to find a way to get rid of `new_checked`

Agreed. Being able to rely solely on compile-time checks would be much more elegant and in line with the intent of the design, for sure.

PR #246, opened 2023-06-01
Add support for Binary and Decimal encodings

andrewwhitehead:
Decimal serialization (outside of `write_decimal_bytes`) is currently only supported when the alloc flag is enabled, as the encoded length is potentially unbounded for `Uint<L>`. In practice anything up to `U8192` can be encoded without allocating. Note that unlike the other encodings, decimal encoding is not constant-time.

Unresolved questions:
- Add a trait for encoding numbers as decimal, to match `FromDecimal`? This is a bit tricky as it would require the `alloc` trait (or gain additional methods when the flag is enabled).
- Implement `Error` for `ParseDecimalError`? This would require a `std` flag.
- Add an error type for encoding errors?


tarcieri:
Sorry for taking awhile to review this. The implementation is somewhat complex which makes it difficult, but I also have some initial thoughts.

The existing support for hex encoding is implemented in a constant-timeish manner, avoiding branches on the inputs in favor of arithmetic operations.

This implementation appears to branch on the input in several places. That's okay, but the corresponding functions should be marked as `*_vartime` to reflect that, per this crate's constant-time-by-default policy.

tarcieri:
Closing as stale

PR #247, opened 2023-06-02
Add fallible hex decoding for Uint

andrewwhitehead:
This is mostly for convenience, although it can be irritating to validate hex inputs outside of the library.

Perhaps it would be nice to support hex strings of different lengths?

### DIFF
--- a/src/uint/encoding.rs
+++ b/src/uint/encoding.rs
@@ -11,6 +11,7 @@ use crate::{Encoding, Limb, Word};
 
 impl<const LIMBS: usize> Uint<LIMBS> {
     /// Create a new [`Uint`] from the provided big endian bytes.
+    /// This method will panic if the input length is incorrect.
     pub const fn from_be_slice(bytes: &[u8]) -> Self {
         assert!(
             bytes.len() == Limb::BYTES * LIMBS,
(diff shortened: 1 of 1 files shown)


src/uint/encoding.rs
```diff
@@ -172,6 +190,22 @@ const fn decode_hex_byte(bytes: [u8; 2]) -> u8 {
...
+    }
+    let mut i = 0;
+    while i < bytes.len() {
+        if !matches!(bytes[i], b'0'..=b'9' | b'a' ..= b'f' | b'A'..=b'F') {
+            return None;
+        }
+        i += 1;
+    }
+    Some(bytes.len() / 2)
+}
```
tarcieri, line 207: This is not constant-time.

Perhaps you could change it to be? Or otherwise you need to make the methods `*_vartime` to honor the "constant time by default" contract of this library, but I would prefer you just make the implementation constant-time instead.

It should also return `CtOption` rather than `Option`.

andrewwhitehead: I don't really see how this can be constant time overall, since the implementation short circuits on invalid input, and I'm not sure what the alternative would be. It would also be nice for it to handle leading zeros on hex values, but in that case is it meant to be constant time in the input length or in the number's representation?

tarcieri: > the implementation short circuits on invalid input, and I'm not sure what the alternative would be.

You can defer checking an error condition until the end of the data, which is what `from_be_hex` does already:

https://github.com/RustCrypto/crypto-bigint/blob/2122b4a/src/uint/encoding.rs#L55-L56

tarcieri:
Closing as stale

PR #248, opened 2023-06-07
Multi-exponentiation

### LINKED ISSUE #250: Multi-Exponentiation
As modular exponentiation is a bottleneck for our library, we have explored ways to optimize it which reuslts in #249 which optimizes the underlying multiplication and in implementing multi-exponentiation. In fact, we have implemented an algorithm from "Straus, E. G. Problems and solutions: Addition chains of vectors. American Mathematical Monthly 71 (1964), 806–808." (which is actually just a slight modification of the current `pow_montgomery_form()` implementation) just as a sketch and were able to achieve a `3x` factor which is significant.

There were several design dilemmas that I prefer leaving to maintainers decision (@tarcieri), esp. because multi-exponentiation gets a vector of bases and exponents, and that requires `alloc`. An alternative was to have a fixed-size array as an input, and since the factor of improvement stops after three elements, it might actually be preferred.

I am opening a draft PR for this, more for the purpose of discussion than for review; after we'll agree on how we want to build it, I can implement it and open the PR.

ycscaly:
Resolves #250


ycscaly:
@tarcieri

src/uint/modular/constant_mod/const_pow.rs
```diff
@@ -33,6 +38,40 @@ impl<MOD: ResidueParams<LIMBS>, const LIMBS: usize> Residue<MOD, LIMBS> {
             phantom: core::marker::PhantomData,
         }
     }
+
+    /// Raises each `base` to its `exponent` power,
+    #[cfg(feature = "alloc")]
+    pub fn multi_exponentiate(
+        bases_and_exponents: Vec<(Self, Uint<LIMBS>)>,
```
tarcieri, line 45: Isn't this a similar problem to your other PRs, where this code could potentially be heapless-friendly if you had a const generic parameter to size the temporary storage, and it might be nice to have both const generic and alloc-based APIs, so possibly a trait impl'd for multiple input types (where I think a slice would suffice here, since the types involved are all `Copy` anyway)

ycscaly: hmm I kinda did that internally, but didn't expose. I will

ycscaly: One problem that arises is the fact `pow_montgomery_form()` is `const`. This does not allow me to keep the allocation there and move the initialization to a `_generic()` helper function, since: ` mutable references are not allowed in constant functions`.

tarcieri: Okay, seems fine for now then

tarcieri (approved):

src/uint/modular/constant_mod/const_pow.rs
ycscaly: No please wait I'm close to have something

src/uint/modular/runtime_mod/runtime_pow.rs
```diff
@@ -35,8 +39,72 @@ impl<const LIMBS: usize> DynResidue<LIMBS> {
...
+
+impl<const N: usize, const LIMBS: usize, const RHS_LIMBS: usize>
+    MultiExponentiateBoundedExp<Uint<RHS_LIMBS>, [(Self, Uint<RHS_LIMBS>); N]>
+    for DynResidue<LIMBS>
+{
+    fn multi_exponentiate_bounded_exp(
+        bases_and_exponents: &[(Self, Uint<RHS_LIMBS>); N],
+        exponent_bits: usize,
+    ) -> Self {
+        // TODO: is it fine? will crash if N == 0
```
ycscaly, line 58:
```suggestion
```
@tarcieri can I document that the function panics if the slice is empty?

src/uint/modular/runtime_mod/runtime_pow.rs
```diff
@@ -35,8 +39,72 @@ impl<const LIMBS: usize> DynResidue<LIMBS> {
...
+
+#[cfg(feature = "alloc")]
+impl<const LIMBS: usize, const RHS_LIMBS: usize>
+    MultiExponentiateBoundedExp<Uint<RHS_LIMBS>, [(Self, Uint<RHS_LIMBS>)]> for DynResidue<LIMBS>
+{
+    fn multi_exponentiate_bounded_exp(
+        bases_and_exponents: &[(Self, Uint<RHS_LIMBS>)],
+        exponent_bits: usize,
+    ) -> Self {
+        // TODO: is it fine? will crash if N == 0
```
ycscaly, line 92:
```suggestion
```

src/uint/modular/constant_mod/const_pow.rs
ycscaly: @tarcieri Ok I'm done. The mutable reference forced me to have some shared functionality DRYed-out into a different function, but I'm overall happy with the result, and generally with this framework we came up with of doing a generic trait implemented both for `alloc` of dynamically sized slices and `no-alloc` for const-generic arrays.

ycscaly:
@tarcieri can you help me figure out the build problems? I think its due to my changes

tarcieri:
@ycscaly looks like you did a `cargo update` or something. Try resetting Cargo.lock?

ycscaly:
> @tarcieri can you help me figure out the build problems? I think its due to my changes

Yeah. Because I can't compile against the Cargo.lock from main. I resetted and the CI is green whilst locally I can't run the code

tarcieri:
What's the local error?

ycscaly:
> What's the local error?

```
error[E0635]: unknown feature `proc_macro_span_shrink`
  --> /Users/jcscaly/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.51/src/lib.rs:92:30
   |
92 |     feature(proc_macro_span, proc_macro_span_shrink)
   |                              ^^^^^^^^^^^^^^^^^^^^^^

```

tarcieri:
What Rust version are you using?

ycscaly:
> What Rust version are you using?

```
➜  crypto-bigint git:(multiexp) ✗ cargo --version
cargo 1.75.0-nightly (b4d18d4bd 2023-10-31)
➜  crypto-bigint git:(multiexp) ✗ rustup --version
rustup 1.26.0 (5af9b9484 2023-04-05)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.75.0-nightly (9d83ac217 2023-10-31)`
➜  crypto-bigint git:(multiexp) ✗
```

tarcieri:
Yeah, that seems like a `nightly` regression. Try stable?

ycscaly:
> Yeah, that seems like a `nightly` regression. Try stable?

works, had `rust-toolchain` set to `nightly` for some reason. So the current code is fine to be merged by me.

src/traits.rs
```diff
@@ -292,6 +292,53 @@ pub trait PowBoundedExp<Exponent> {
...
+    fn multi_exponentiate(bases_and_exponents: &BasesAndExponents) -> Self;
+}
+
+impl<
+        T: MultiExponentiateBoundedExp<Exponent, BasesAndExponents>,
+        Exponent: Bounded,
+        BasesAndExponents,
+    > MultiExponentiate<Exponent, BasesAndExponents> for T
+where
+    BasesAndExponents: AsRef<[(Self, Exponent)]> + ?Sized,
```
tarcieri, line 318:
```suggestion
impl<T, Exponent, BasesAndExponents> MultiExponentiate<Exponent, BasesAndExponents> for T
where
    T: MultiExponentiateBoundedExp<Exponent, BasesAndExponents>,
    Exponent: Bounded,
    BasesAndExponents: AsRef<[(Self, Exponent)]> + ?Sized,
```

src/uint/modular/runtime_mod/runtime_pow.rs
tarcieri: It feels like the sort of thing where it should be possible to do some kind of const assertion, although I forget how to work around the "can't use generic parameters from outer item" problem

tarcieri, line 58: Something like this, but unfortunately this does run into the "can't use generic parameters from outer item" error:

```suggestion
        const _: () = assert!(N != 0, "N cannot be zero");
```

I saw some blog post that showed a workaround recently, but now I can't find it anymore.

tarcieri: Absent a solution like this, you can panic for now with a TODO to replace the runtime panic with a const assert/bounds, e.g. using `generic_const_exprs`

tarcieri: Oh, apparently there is a hacky way to make a const assertion work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b522f72bebe0140e1302894391107903

tarcieri: Opened a PR to add some macros that implement such assertions in a much clearer way: https://github.com/RustCrypto/crypto-bigint/pull/293

tarcieri: Okay, with #293 merged, if you rebase, this should now work:

```suggestion
        const_assert_ne!(N, 0, "bases_and_exponents must not be empty");
```

tarcieri: Posted a solution [here](https://github.com/RustCrypto/crypto-bigint/pull/248#discussion_r1397661028)

ycscaly:
@tarcieri OK I did so, but couldn't use `const_assert` in the `alloc` function, so used `assert`.

src/uint/modular/runtime_mod/runtime_pow.rs
```diff
@@ -35,8 +39,76 @@ impl<const LIMBS: usize> DynResidue<LIMBS> {
...
+{
+    fn multi_exponentiate_bounded_exp(
+        bases_and_exponents: &[(Self, Uint<RHS_LIMBS>)],
+        exponent_bits: usize,
+    ) -> Self {
+        assert_ne!(
+            bases_and_exponents.len(),
+            0,
+            "bases_and_exponents must not be empty"
+        );
```
tarcieri, line 96:
```suggestion
        assert!(
            !bases_and_exponents.is_empty(),
            "bases_and_exponents must not be empty"
        );
```

src/traits.rs
```diff
@@ -292,6 +292,51 @@ pub trait PowBoundedExp<Exponent> {
     fn pow_bounded_exp(&self, exponent: &Exponent, exponent_bits: usize) -> Self;
 }
 
+/// Linear combination (extended version).
+///
+/// This trait enables providing an optimized implementation of
+/// linear combinations (e.g. Shamir's Trick).
+// TODO(tarcieri): replace the current `LinearCombination` with this in the next release
```
tarcieri, line 299: This feels like a copy-paste error?

src/traits.rs
```diff
@@ -292,6 +292,51 @@ pub trait PowBoundedExp<Exponent> {
     fn pow_bounded_exp(&self, exponent: &Exponent, exponent_bits: usize) -> Self;
 }
 
+/// Linear combination (extended version).
```
ycscaly, line 295:
```suggestion
```

src/traits.rs
```diff
@@ -292,6 +292,51 @@ pub trait PowBoundedExp<Exponent> {
...
+///
+/// This trait enables providing an optimized implementation of
+/// linear combinations (e.g. Shamir's Trick).
+// TODO(tarcieri): replace the current `LinearCombination` with this in the next release
+
```
ycscaly, line 300:
```suggestion
```

tarcieri (approved):

ycscaly:
Thanks!

ycscaly:
@tarcieri are we too late for the release?

tarcieri:
Nope, though I'd like to get #277 in as well

PR #252, opened 2023-06-13
Expose and optimize wrapping_neg

andrewwhitehead:


src/uint/shl.rs
```diff
@@ -5,6 +5,7 @@ use core::ops::{Shl, ShlAssign};
 
 impl<const LIMBS: usize> Uint<LIMBS> {
     /// Computes `self << 1` in constant-time, returning the overflowing bit as a `CtChoice`.
+    #[allow(unused)]
```
fjarri, line 8: Since it's not public, I think we can safely remove it if it's unused

andrewwhitehead: Updated now

tarcieri (approved):

PR #253, opened 2023-06-16
Concat/split and multiply mixed size Uints

andrewwhitehead:
As a potential replacement for #230 

Adds ConcatMixed and SplitMixed traits, and allows any Uint up to U1024 to be arbitrarily decomposed (down to U64). With better support for generic trait properties and methods these traits might be removed later.

Also adds cross-size multiplication, returning a combined-width result.

This does not affect the build time much (about 0.1-0.2s increase for a clean build on my machine).


src/traits.rs
```diff
@@ -187,26 +187,47 @@ pub trait CheckedSub<Rhs = Self>: Sized {
     fn checked_sub(&self, rhs: Rhs) -> CtOption<Self>;
 }
 
-/// Concatenate two numbers into a "wide" twice-width value, using the `rhs`
+/// Concatenate two numbers into a "wide" combined-width value, using the `lo`
 /// value as the least significant value.
-pub trait Concat<Rhs = Self> {
+pub trait ConcatMixed<Lo: ?Sized = Self> {
```
tarcieri, line 192: I think it would make sense to define this below `Concat` (and same for `Split`/`SplitMixed`)

andrewwhitehead: Thanks, updated

tarcieri (approved):

PR #254, opened 2023-06-19
Make `Uint::from_be_hex()` and `from_le_hex()` constant-time

fjarri:
The general `crypto-bigint` convention is that the things are constant-time, unless explicitly marked otherwise. This PR makes `from_be/le_hex()` constant-time using a constant-time decoding snippet from the `base16ct` crate.

Alternatives to this approach:
- `base16ct` could be possibly used directly, but the high-level functions there take a mutable reference and therefore cannot be `const`. One could expose `base16ct::decode_nibble()`, or perhaps add a variant of functions operating on arrays. 
- We could just mark `from_be/le_hex()` as vartime

Also, should there be non-panicking `from_be/le_hex()` variants that would return an error instead, for more convenient use in runtime contexts? (Edit: this is actually handled by #247)

### DIFF
--- a/src/uint/encoding.rs
+++ b/src/uint/encoding.rs
@@ -46,17 +46,22 @@ impl<const LIMBS: usize> Uint<LIMBS> {
         let mut res = [Limb::ZERO; LIMBS];
         let mut buf = [0u8; Limb::BYTES];
         let mut i = 0;
+        let mut err = 0;
 
         while i < LIMBS {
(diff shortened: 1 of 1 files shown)


tarcieri (approved):

tarcieri:
This seems fine for now. It's a bit of code duplication but figuring out the public API for `base16ct` seems hard. Maybe we can do that eventually.

PR #255, opened 2023-06-26
Add U16384 and U32768

ycscaly:
This PR adds the U16384 and U32768 types.


ycscaly:
@tarcieri now that our threshold-Paillier repo is public, this simple PR is the only one stopping us from working with upstream `crypto-bigint`

ycscaly:
Thank you!

PR #256, opened 2023-06-28
NCC audit fixes

fjarri:
Some small things that were picked up by the security audit.

- Removed conditionals in `Uint::saturating_add` and `saturating_sub` and replaced them with constant-time methods.
- Made checks in the `random_mod` test more logical
- Marked `Uint::sqrt` to be renamed to `sqrt_vartime` (and similarly `checked_sqrt` and `wrapping_sqrt`) on the next breaking release, since that's the crate's convention.

### DIFF
--- a/src/uint/sqrt.rs
+++ b/src/uint/sqrt.rs
@@ -5,11 +5,20 @@ use crate::{Limb, Word};
 use subtle::{ConstantTimeEq, CtOption};
 
 impl<const LIMBS: usize> Uint<LIMBS> {
+    /// See [`Self::sqrt_vartime`].
+    #[deprecated(
+        since = "0.5.3",
+        note = "This functionality will be moved to `sqrt_vartime` in a future release."
+    )]
+    pub const fn sqrt(&self) -> Self {
+        self.sqrt_vartime()
+    }
+
     /// Computes √(`self`)
     /// Uses Brent & Zimmermann, Modern Computer Arithmetic, v0.5.9, Algorithm 1.13
     ///
     /// Callers can check if `self` is a square by squaring the result
-    pub const fn sqrt(&self) -> Self {
+    pub const fn sqrt_vartime(&self) -> Self {
         let max_bits = (self.bits_vartime() + 1) >> 1;
         let cap = Self::ONE.shl_vartime(max_bits);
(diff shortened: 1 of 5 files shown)


src/uint/sqrt.rs
```diff
@@ -9,7 +9,7 @@ impl<const LIMBS: usize> Uint<LIMBS> {
     /// Uses Brent & Zimmermann, Modern Computer Arithmetic, v0.5.9, Algorithm 1.13
     ///
     /// Callers can check if `self` is a square by squaring the result
-    pub const fn sqrt(&self) -> Self {
+    pub const fn sqrt_vartime(&self) -> Self {
```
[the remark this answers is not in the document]
fjarri: Ok, I'll do that instead. This means it will still be renamed in 0.6.0, correct? Do you have any automated procedure to handle these deprecation markers and make sure you don't forget any on release?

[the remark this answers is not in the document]
fjarri: Fixed.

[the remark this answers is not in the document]
tarcieri: Yes, it will be removed in the next release.

There's no automated check for that. It's just something you have to do.

tests/proptests.rs
```diff
@@ -187,6 +187,7 @@ proptest! {
     }
 
     #[test]
+    #[allow(deprecated)]
```
tarcieri, line 190: Perhaps use the new name here?

src/uint/sqrt.rs
```diff
@@ -65,6 +79,8 @@ impl<const LIMBS: usize> Uint<LIMBS> {
 
 #[cfg(test)]
 mod tests {
+    #![allow(deprecated)]
```
tarcieri, line 82: Why are you testing the deprecated names?

fjarri: Sorry, I wasn't sure about the procedure. So we're adding a new name without removing the old one, correct?

tarcieri (approved):

src/uint/sqrt.rs
tarcieri: Looks like you got everything updated.

PR #261, opened 2023-08-03
Add `TryFrom<&[u8]>` bound to `Encoding::Repr`

fjarri:
This helps with deserialization - one doesn't need to create a mutable buffer and use a special purpose `deserialize` mutating its argument (like `serdect` does currently), but instead one can just use a standard `deserialize` with the API compatible with that of `#[serde(with)]` that returns the `Repr` object.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -244,7 +244,12 @@ pub trait Bounded {
 /// Encoding support.
 pub trait Encoding: Sized {
     /// Byte array representation.
-    type Repr: AsRef<[u8]> + AsMut<[u8]> + Copy + Clone + Sized;
+    type Repr: AsRef<[u8]>
+        + AsMut<[u8]>
+        + Copy
+        + Clone
+        + Sized
+        + for<'a> TryFrom<&'a [u8], Error = core::array::TryFromSliceError>;
 
     /// Decode from big endian bytes.
     fn from_be_bytes(bytes: Self::Repr) -> Self;


src/traits.rs
```diff
@@ -244,7 +244,12 @@ pub trait Bounded {
 /// Encoding support.
 pub trait Encoding: Sized {
     /// Byte array representation.
-    type Repr: AsRef<[u8]> + AsMut<[u8]> + Copy + Clone + Sized;
+    type Repr: AsRef<[u8]>
+        + AsMut<[u8]>
+        + Copy
+        + Clone
+        + Sized
+        + for<'a> TryFrom<&'a [u8], Error = core::array::TryFromSliceError>;
```
tarcieri, line 252: I don't think the error type is particularly relevant here? Do we really need to mandate it?

fjarri: It helps a lot with writing down trait bounds in the user code. See https://github.com/rust-lang/rust/issues/113517 for example.

fjarri: To add to that, even with the workaround in the issue above, it does not work when another level is added. That is, if the `foo()` in the issue is a `serde`-standard `deserialize()` function, and I want to make a custom deserializer using `serde_with` as
```rust
impl<'de, const N: usize> serde_with::DeserializeAs<'de, Uint<N>> for UintWrapper<N>
where
    Uint<N>: Encoding,
```
there is no way to write trait bounds that would tell the compiler that the returned error is `Display` (so that it could be wrapped in the `serde` error).

tarcieri: Okay, if there's a legitimate use case I guess we can include it.

It's just a completely useless error that includes no relevant information, so it's easily handled without bounding on it by using `.ok()` or `.map_err(...)`.

But if it helps with `serde`, I guess that's fine.

tarcieri:
Also note: this is a breaking change, so we can't merge it until the next breaking release.

src/traits.rs
fjarri: This one is pretty useless, I agree. But if you have a function that is generic on something that's `TryFrom<&[u8]>` and you want the error to be `Display` (because it may be useful for some types, so you don't want to discard it), fixing the error in this specific case helps with the bounds. I don't particularly like it either, but unfortunately Rust compiler is currently not smart enough to handle bounds without it. I think it should be pretty harmless, since it's essentially an internal trait that's not supposed to be used outside of the crate.

tarcieri:
Bumped version to v0.6.0-pre in #295. Merging

Issue #227, opened 2023-05-06
Constant-time GCD


fjarri:
There is already an item for that in #1, but I would like to make a separate issue to add more info, and to make it more noticeable. It is likely I will have a go at it the next time I have free time available, in which case I will assign myself to the issue; before that it's free for the taking.

So the state of the art seems to be:
- Bernstein & Yang "Fast constant-time gcd computation and modular inversion" (https://gcd.cr.yp.to/safegcd-20190413.pdf)
- Improved bounds for the paper above: https://github.com/sipa/safegcd-bounds
- Pornin "Optimized Binary GCD for Modular Inversion" (https://eprint.iacr.org/2020/972.pdf)
- Implementation tricks used in libsecp256k1: https://github.com/bitcoin-core/secp256k1/blob/master/doc/safegcd_implementation.md

Additionally, with the GCD available, it seems straightforward to implement Jacobi symbol and modular inversion based on it. The latter may supersede the existing `inv_odd_mod_bounded()`, which was ported from GMP and therefore lacks any comments or references explaining the algorithm.

tarcieri:
FWIW, all of the `fiat-crypto` code uses Bernstein-Yang inversions

mratsim:
See implementation in Rust: https://github.com/privacy-scaling-explorations/halo2curves/pull/83

tarcieri:
#372 added an initial implementation of Bernstein-Yang, but currently providing modular inversions only (and not fully wired up to the existing inversion APIs yet).

Constant-time GCD can be implemented in terms of its "divstep" function.

tarcieri:
#472 implements GCD using Bernstein-Yang

PR #263, opened 2023-08-23
Modular inversion improvements

fjarri:
- fix a typo in a proptest for `rem()` which rendered it useless
- make `inv_mod2k()` constant-time both in `self` and `k`
- rename the old `inv_mod2k()` to `inv_mod2k_vartime()`
- make the code in `inv_mod2k_vartime()` clearer and fix the attribution
- add a constant-time `inv_mod()` that supports both even and odd moduli (works via RNS and `inv_mod2k()`/`inv_odd_mod()`; another approach is binary GCD (see #227 for more info) which may or may not be faster, but is definitely more complicated)

Note that this PR does not introduce any backwards-incompatible changes (except for the performance regression for the users of `inv_mod2k()`).

Possible additions:
- We can add `inv_mod2k_bounded()`, for the cases where a bound on `k` is known. 
- We can add `inv_mod_vartime()` (that calls `inv_mod2k_vartime()` internally, which would lead to a significant performance gain)

(questions involving breaking changes are moved to #268)

### DIFF
--- a/src/uint/inv_mod.rs
+++ b/src/uint/inv_mod.rs
@@ -1,28 +1,68 @@
 use super::Uint;
-use crate::{CtChoice, Limb};
+use crate::CtChoice;
 
 impl<const LIMBS: usize> Uint<LIMBS> {
-    /// Computes 1/`self` mod 2^k as specified in Algorithm 4 from
-    /// A Secure Algorithm for Inversion Modulo 2k by
-    /// Sadiel de la Fe and Carles Ferrer. See
-    /// <https://www.mdpi.com/2410-387X/2/3/23>.
+    /// Computes 1/`self` mod `2^k`.
+    /// This method is constant-time w.r.t. `self` but not `k`.
     ///
     /// Conditions: `self` < 2^k and `self` must be odd
-    pub const fn inv_mod2k(&self, k: usize) -> Self {
-        let mut x = Self::ZERO;
-        let mut b = Self::ONE;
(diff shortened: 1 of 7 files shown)


tarcieri:
@fjarri needs a rebase

fjarri:
#267 should go in first, then I'll rebase on top of that.

fjarri:
Rebased.

tarcieri (approved):

src/uint/inv_mod.rs
```diff
@@ -97,10 +137,45 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+    /// Returns `(inverse, CtChoice::TRUE)` if an inverse exists,
+    /// otherwise `(undefined, CtChoice::FALSE)`.
     pub const fn inv_odd_mod(&self, modulus: &Self) -> (Self, CtChoice) {
         self.inv_odd_mod_bounded(modulus, Uint::<LIMBS>::BITS, Uint::<LIMBS>::BITS)
     }
+
+    /// Computes the multiplicative inverse of `self` mod `modulus`.
+    /// Returns `(inverse, CtChoice::TRUE)` if an inverse exists,
+    /// otherwise `(undefined, CtChoice::FALSE)`.
+    pub fn inv_mod(&self, modulus: &Self) -> (Self, CtChoice) {
```
fjarri, line 149: Just realized, it could have been made `const`.

PR #270, opened 2023-08-30
Fix clippy warnings

fjarri:
Fix a bunch of clippy warnings that are raised with Rust 1.72, but do not appear in CI which uses 1.65 for clippy. Makes it easier to run clippy during development. Clippy bumped to 1.72 in CI.

I decided to go with `#![allow(clippy::unwrap_used)]` for test submodules instead of replacing each `unwrap()` there with `expect()`.

### DIFF
--- a/src/boxed/uint/add.rs
+++ b/src/boxed/uint/add.rs
@@ -26,6 +26,7 @@ impl CheckedAdd<&BoxedUint> for BoxedUint {
 }
 
 #[cfg(test)]
+#[allow(clippy::unwrap_used)]
 mod tests {
     use super::{BoxedUint, CheckedAdd, Limb};
 
--- a/src/limb.rs
+++ b/src/limb.rs
@@ -1,8 +1,6 @@
 //! Big integers are represented as an array of smaller CPU word-size integers
 //! called "limbs".
 
-#![allow(clippy::derive_hash_xor_eq)]
-
 mod add;
 mod bit_and;
 mod bit_not;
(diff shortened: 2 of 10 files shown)


src/boxed/uint/add.rs
```diff
@@ -27,6 +27,8 @@ impl CheckedAdd<&BoxedUint> for BoxedUint {
 
 #[cfg(test)]
 mod tests {
+    #![allow(clippy::unwrap_used)]
```
tarcieri, line 30: Perhaps this way instead? (I find it a bit unusual to use `#!` on an inline module)

```suggestion
#[cfg(test)]
#[allow(clippy::unwrap_used)]
mod tests {
```

fjarri: I guess I wanted to create more of an "internal property" feeling. But no problem, I'll change it.

src/limb.rs
```diff
@@ -60,6 +58,8 @@ pub(crate) const HI_BIT: usize = Limb::BITS - 1;
 
 /// Big integers are represented as an array of smaller CPU word-size integers
 /// called "limbs".
+#[allow(renamed_and_removed_lints)]
+#[allow(clippy::derive_hash_xor_eq)] // renamed to `derived_hash_with_manual_eq` since 1.69
```
tarcieri, line 62: Perhaps just bump clippy?

fjarri: I'm not sure what's your policy on CI, currently clippy is run on 1.65 there.

fjarri: I mean, if you want to keep running clippy on MSRV, that's what we have to do.

tarcieri: It doesn't have to be MSRV. We can bump it to 1.72.

It's just pinned so it doesn't randomly break every time there's a new release.

fjarri: Ok, I'll bump it

PR #275, opened 2023-09-09
vartime remark

rickwebiii:
Add remark for bits() method documentation.


tarcieri:
Perhaps add something like "and will operate in constant-time for a fixed index"?

tarcieri (approved):

Issue #268, opened 2023-08-28
A list of breaking changes to consider for v0.6


fjarri:
There's a lot of small stuff so I decided to put them in one issue:
- [x] (#373) The types of the arguments to `shl`/`shr` methods are inconsistent: `std` has them as `u32`, `Limb` as `Limb`, and `Uint` as `usize`. Perhaps we should at least match `Limb` and `Uint`. 
- [x] (#395) What should `Uint::shl_limb()`, `shl_vartime()`, and `shl()` (same for `shr()`) do on `shift` overflow? Currently `shl_limb()` panics, `shl_vartime()` returns `self`, and `shl()` returns zero. Should probably do the same thing. See also #121
- [x] (#416) Should `inv_mod2k()` return `CtChoice::FALSE` if `self` is even? Should `inv_odd_mod()` return `CtChoice::FALSE` if `modulus` is even? 
- [x] (#417) We need to make the usage of `ct_` and `const_` prefixes uniform. One prefix, either `const_` or `ct_`, should be used for `const fn` when there is a non-const `fn` with the same functionality (otherwise no prefix needed); `ct_` meaning "constant-time" really shouldn't be used at all because that's the default. `uint/div.rs` is especially bad in this respect.
- [x] (#419) Should `const fn` division methods take a `NonZero`-wrapped modulus? Should `div_rem_limb_with_reciprocal()` take a `CtOption` of the reciprocal, or the caller should do the mapping instead?

AaronFeickert:
The functionality for `DynResidueParams::new_checked` added in #240 was done with the understanding that it could be moved to `DynResidueParams::new` in a future breaking release.

tarcieri:
`Uint::{sqrt, checked_sqrt, wrapping_sqrt}` should be removed and/or replaced with a constant-time implementation

tarcieri:
We're getting the ball rolling on another cycle of breaking changes over at https://github.com/rustcrypto/traits

I can cut a v0.5.4 release here and we can switch `master` to be `v0.6.0-pre`. If need be we can make a `0-5-stable` branch for changes to the current release series.

tarcieri:
Starting on this in #295

tarcieri:
#296 made `DynResidueParams::new` fallible

tarcieri:
[...]

PR #277, opened 2023-09-13
Constant-time square root

HastD:
This un-deprecates `Uint::sqrt` and provides a constant-time implementation. (Unfortunately this makes it non-`const`, so it's technically a breaking change; I don't see a way to handle zero without using `checked_div`, which is why this can't be `const`. For `const` contexts, `Uint::sqrt_vartime` can still be used.) The algorithm is just a slight modification of Algorithm 1.13 of Brent-Zimmermann as used in `sqrt_vartime`: instead of iterating until `xn` increases, we just iterate a fixed number of times (logarithmic in the number of bits) and then take the minimum of `guess` and `xn` at the end.

The reason this works is because of two properties of this algorithm that I prove here: [ct_sqrt.pdf](https://github.com/RustCrypto/crypto-bigint/files/12600669/ct_sqrt.pdf). (We can upload this somewhere more permanent—not sure where would be best. Also, it'd be great if someone could look this over to make sure I didn't make any mistakes in the proof.) First, we have an upper bound the number of iterations required. Second, once the algorithm reaches the desired value `sqrt(self)`, it will either repeat `sqrt(self)` or it will alternate between `sqrt(self)` and `sqrt(self) + 1`; thus, we can return the minimum of `guess` and `xn`.

I also removed some dead code in `Uint::sqrt_vartime`: unless I'm really missing something here, the first `while` loop can never run because the initial value of `guess` is defined such that it's never an underestimate.

### DIFF
--- a/src/uint/sqrt.rs
+++ b/src/uint/sqrt.rs
@@ -4,13 +4,33 @@ use super::Uint;
 use subtle::{ConstantTimeEq, CtOption};
 
 impl<const LIMBS: usize> Uint<LIMBS> {
-    /// See [`Self::sqrt_vartime`].
-    #[deprecated(
-        since = "0.5.3",
-        note = "This functionality will be moved to `sqrt_vartime` in a future release."
-    )]
-    pub const fn sqrt(&self) -> Self {
-        self.sqrt_vartime()
+    /// Computes √(`self`) in constant time.
+    /// Based on Brent & Zimmermann, Modern Computer Arithmetic, v0.5.9, Algorithm 1.13
+    ///
+    /// Callers can check if `self` is a square by squaring the result
+    pub fn sqrt(&self) -> Self {
+        let max_bits = (self.bits() + 1) >> 1;
+        let cap = Self::ONE.shl(max_bits);
+        let mut guess = cap; // ≥ √(`self`)
+        let mut xn = {
+            let q = self.wrapping_div(&guess);
+            let t = guess.wrapping_add(&q);
+            t.shr_vartime(1)
+        };
+
+        // Repeat enough times to guarantee result has stabilized.
+        // See Hast, "Note on computation of integer square roots" for a proof of this bound.
+        for _ in 0..Self::BITS.ilog2() + 1 {
+            guess = xn;
+            xn = {
+                let q = self.checked_div(&guess).unwrap_or(Self::ZERO);
+                let t = guess.wrapping_add(&q);
+                t.shr_vartime(1)
+            };
+        }
+
+        // at least one of `guess` and `xn` is now equal to √(`self`), so return the minimum
+        Self::ct_select(&guess, &xn, Uint::ct_gt(&guess, &xn))
     }
 
     /// Computes √(`self`)
--- a/src/uint/sqrt.rs
+++ b/src/uint/sqrt.rs
@@ -1,7 +1,6 @@
 //! [`Uint`] square root operations.
 
 use super::Uint;
-use crate::{Limb, Word};
 use subtle::{ConstantTimeEq, CtOption};
 
 impl<const LIMBS: usize> Uint<LIMBS> {
@@ -27,21 +26,7 @@ impl<const LIMBS: usize> Uint<LIMBS> {
             let t = guess.wrapping_add(&q);
             t.shr_vartime(1)
         };
-
-        // If guess increased, the initial guess was low.
-        // Repeat until reverse course.
-        while Uint::ct_lt(&guess, &xn).is_true_vartime() {
-            // Sometimes an increase is too far, especially with large
-            // powers, and then takes a long time to walk back.  The upper
-            // bound is based on bit size, so saturate on that.
-            let le = Limb::ct_le(Limb(xn.bits_vartime() as Word), Limb(max_bits as Word));
-            guess = Self::ct_select(&cap, &xn, le);
-            xn = {
-                let q = self.wrapping_div(&guess);
-                let t = guess.wrapping_add(&q);
-                t.shr_vartime(1)
-            };
-        }
+        // Note, xn <= guess at this point.
 
         // Repeat while guess decreases.
         while Uint::ct_gt(&guess, &xn).is_true_vartime() && xn.ct_is_nonzero().is_true_vartime() {


tarcieri:
> We can upload this somewhere more permanent—not sure where would be best.

arXiv perhaps?

HastD:
Sure, I can upload it to arXiv (cs.DS probably?). I'd like to wait until at least one other person has looked it over it reduce the chance of errors slipping through, but after that I'll go ahead.

tarcieri:
cc @fjarri

src/uint/sqrt.rs
```diff
@@ -1,17 +1,36 @@
...
+    ///
+    /// Callers can check if `self` is a square by squaring the result
+    pub fn sqrt(&self) -> Self {
+        let max_bits = (self.bits() + 1) >> 1;
+        let cap = Self::ONE.shl(max_bits);
+        let mut guess = cap; // ≥ √(`self`)
+        let mut xn = {
+            let q = self.wrapping_div(&guess);
+            let t = guess.wrapping_add(&q);
+            t.shr_vartime(1)
```
fjarri, line 18: Should be just `shr()`

fjarri:
> I don't see a way to handle zero without using `checked_div`, which is why this can't be const

I am not sure I see the problem here, you can always use the underlying `Uint::ct_div_rem()` and use the returned `CtChoice` to conditionally select the result.

src/uint/sqrt.rs
fjarri: Actually no, disregard that. My mistake.

HastD:
> I am not sure I see the problem here, you can always use the underlying `Uint::ct_div_rem()` and use the returned `CtChoice` to conditionally select the result.

I just overlooked that, so there's no obstruction to keeping it `const`. Actually, it looks like `Uint::ct_div_rem` [already internally does that conditional selection](https://github.com/RustCrypto/crypto-bigint/blob/master/src/uint/div.rs#L75) (so the returned quotient is zero if `rhs` is zero); can that behavior be relied on or is it an internal implementation detail of `ct_div_rem`? For reference, the relevant lines of code in `ct_div_rem`:

```rust
    pub(crate) const fn ct_div_rem(&self, rhs: &Self) -> (Self, Self, CtChoice) {
        let mb = rhs.bits_vartime();
        // [...]
        let is_some = Limb(mb as Word).ct_is_nonzero();
        quo = Self::ct_select(&Self::ZERO, &quo, is_some);
        (quo, rem, is_some)
    }
```

src/uint/sqrt.rs
```diff
@@ -1,17 +1,36 @@
...
+    /// Computes √(`self`) in constant time.
+    /// Based on Brent & Zimmermann, Modern Computer Arithmetic, v0.5.9, Algorithm 1.13
+    ///
+    /// Callers can check if `self` is a square by squaring the result
+    pub fn sqrt(&self) -> Self {
+        let max_bits = (self.bits() + 1) >> 1;
+        let cap = Self::ONE.shl(max_bits);
+        let mut guess = cap; // ≥ √(`self`)
+        let mut xn = {
+            let q = self.wrapping_div(&guess);
```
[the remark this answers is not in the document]
HastD: My mistake, I didn't see that `wrapping_div` is only constant-time for a fixed `rhs`. Actually, I think there might be a documentation issue here—I can't find anywhere in the public documentation that says this, only the documentation for `ct_div_rem` (which doesn't appear in the public docs since it's `pub(crate)`).

What do you think would be the better approach here: making a new function that's like `ct_div_rem` but constant-time with respect to both inputs, or modifying `ct_div_rem` to have this stronger constant-time guarantee and moving the "constant-time only for fixed rhs" behavior to a new function?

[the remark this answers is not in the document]
fjarri: I agree that the name is misleading - it should have had a `vartime` suffix (and the whole `div.rs` is kind of a mess in terms of naming - see #268). So the proper way to proceed I think would be to rename the current one to `_vartime`, and implement a constant-time one in its place.

fjarri:
> can that behavior be relied on or is it an internal implementation detail of `ct_div_rem`

I think it should be treated as an implementation detail.

fjarri:
Also:
- I'll go through your proof in a few days, looks like it'll require some time and concentration :)
- Perhaps a proptest (see `tests/proptests.rs`) would be good to add here, for both vartime and const-time versions.

HastD:
Okay, I just made the necessary changes so `sqrt` and `wrapping_sqrt` are `const` functions again, and I modified the `wrapping_sqrt` proptest to test both the vartime and constant-time versions. However, as discussed above, it's still not actually constant-time yet because `ct_div_rem` (and hence also `wrapping_div`) are currently only constant-time for fixed divisor.

HastD:
Okay, I added an implementation of division that's constant-time in both arguments, and moved the old division implementation (that's constant-time only for fixed divisor) to `_vartime` functions (`div_rem_vartime`, `wrapping_div_vartime`, etc). If this has been implemented correctly then `Uint::sqrt` should be fully constant-time now.

My constant-time division implementation is a straightforward modification of the vartime algorithm, just replacing vartime methods with their constant-time variants, and running for the maximum number of iterations with a `ct_select` to avoid modifying the outputs beyond when the loop would've terminated in the vartime version. I suspect it's possible to do this more efficiently but I'll put this out there as a starting point. How's it look?

src/uint/sqrt.rs
```diff
@@ -1,17 +1,39 @@
...
+            let t = guess.wrapping_add(&q);
+            t.shr_vartime(1)
+        };
+
+        // Repeat enough times to guarantee result has stabilized.
+        // See Hast, "Note on computation of integer square roots" for a proof of this bound.
+        let mut i = 0;
+        while i < usize::BITS - Self::BITS.leading_zeros() {
```
fjarri, line 24: Self::LOG2_BITS could be used here

src/uint/sqrt.rs
```diff
@@ -1,17 +1,39 @@
...
+        let cap = Self::ONE.shl(max_bits);
+        let mut guess = cap; // ≥ √(`self`)
+        let mut xn = {
+            let q = self.wrapping_div(&guess);
+            let t = guess.wrapping_add(&q);
+            t.shr_vartime(1)
+        };
+
+        // Repeat enough times to guarantee result has stabilized.
+        // See Hast, "Note on computation of integer square roots" for a proof of this bound.
```
fjarri, line 22: When your paper has a more permanent link (e.g. on arxiv), please make a PR referencing it here

src/uint/sqrt.rs
```diff
@@ -1,17 +1,39 @@
...
+        let mut i = 0;
+        while i < usize::BITS - Self::BITS.leading_zeros() {
+            guess = xn;
+            xn = {
+                let (q, _, is_some) = self.const_div_rem(&guess);
+                let q = Self::ct_select(&Self::ZERO, &q, is_some);
```
fjarri, line 28: Is this needed specifically to handle the case of `self == 0`?

src/uint/sqrt.rs
```diff
@@ -1,17 +1,39 @@
...
+                let t = guess.wrapping_add(&q);
+                t.shr_vartime(1)
+            };
+            i += 1;
+        }
+
+        // at least one of `guess` and `xn` is now equal to √(`self`), so return the minimum
```
fjarri, line 35: So at this point `guess == x_n`, and `xn == x_{n+1}`, where `n = floor(log2(Self::BITS))`. But in the paper it says that it should be `n = floor(log2(Self::BITS)) + 1` - am I missing something?

src/uint/sqrt.rs
```diff
@@ -23,31 +45,17 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
-            xn = {
-                let q = self.wrapping_div(&guess);
-                let t = guess.wrapping_add(&q);
-                t.shr_vartime(1)
-            };
-        }
+        // Note, xn <= guess at this point.
 
         // Repeat while guess decreases.
         while Uint::ct_gt(&guess, &xn).is_true_vartime() && xn.ct_is_nonzero().is_true_vartime() {
```
fjarri, line 55: We don't need to use `ct_gt()` and `ct_is_nonzero()` here, those are constant-time

src/uint/sqrt.rs
```diff
@@ -56,29 +64,26 @@ impl<const LIMBS: usize> Uint<LIMBS> {
         Self::ct_select(&Self::ZERO, &guess, self.ct_is_nonzero())
```
fjarri, line 64: Similarly here, we don't need constant-timeness.

fjarri:
Sorry for the delay, I was kind of swamped at work. I've read the paper, and it looks sound to me (although a bit too terse :). Have you submitted it anywhere yet? Would be nice to have a reference in the comments to the code. 

I've also got some relatively minor comments to the code.

tarcieri:
> I've read the paper, and it looks sound to me (although a bit too terse :). Have you submitted it anywhere yet? Would be nice to have a reference in the comments to the code.

I think with a bit more polish on it you could throw it on the IACR ePrint service.

src/uint/sqrt.rs
```diff
@@ -103,13 +108,24 @@ mod tests {
 
     #[test]
     fn edge() {
+        assert_eq!(U256::ZERO.sqrt(), U256::ZERO);
+        assert_eq!(U256::ONE.sqrt(), U256::ONE);
+        let mut half = U256::ZERO;
+        for i in 0..half.limbs.len() / 2 {
+            half.limbs[i] = Limb::MAX;
+        }
+        assert_eq!(U256::MAX.sqrt(), half);
```
fjarri, line 117: An idea for the edge case test: a number that actually needs the maximum amount of iterations to converge. According to my tests, the ones before 10000 are 80, 99, 4224, 4355, 4488, 4623, 4760, 4899; but please check independently. (Also an interesting mathematical question - is there some rule for their distribution)

src/uint/sqrt.rs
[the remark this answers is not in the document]
tarcieri: It appears this is a blocker on merging this PR. I guess we can go ahead and flip over to the v0.6 series per #268 and try to land this PR afterward.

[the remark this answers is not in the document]
tarcieri: Checking on this: `wrapping_div` calls `const_div_rem`, which claims:

```rust
    /// This function is constant-time with respect to both `self` and `rhs`.
    pub(crate) const fn const_div_rem(&self, rhs: &Self) -> (Self, Self, CtChoice) {
```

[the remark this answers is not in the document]
fjarri: Which commit are you looking at? The current `master` has no `const_div_rem()`, and `Uint::wrapping_div()` uses `ct_div_rem()`, which is not constant-time in `rhs`.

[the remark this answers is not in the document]
tarcieri: This PR changes `wrapping_div` to call `const_div_rem`, which is also added by this PR

tarcieri:
Merged in #376

PR #279, opened 2023-09-18
Miller-Rabin primality test and primes generation

maksimryndin:
Hi guys! Thanks for a nice library on bigints!

This PR implements Miller-Rabin primality test as a starting point for checking primality and generating random primes.

- Miller-Rabin primality test method `is_prime_miller_rabin_vartime` for `UInt`
- `RandomPrime` trait for generating random prime `UInt`s

Miller-Rabin can be implemented using only existing crate machinery.
As far as I understood, all other algorithms are more advanced and not necessarily demanded for practical purposes (
https://crypto.stackexchange.com/questions/71/how-can-i-generate-large-prime-numbers-for-rsa
https://math.stackexchange.com/questions/1064784/do-we-still-need-probabilistic-primality-testing-methods-for-practical-applicati#1069342)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.5.4 (2023-08-18)
+### Added
+- Miller-Rabin primality test method `is_prime_miller_rabin_vartime` for `UInt` ([#279])
+- `RandomPrime` trait for generating random prime `UInt`s ([#279])
+
+[#279]: https://github.com/RustCrypto/crypto-bigint/pull/279
+
 ## 0.5.3 (2023-09-04)
 ### Added
 - `BoxedUint`: heap-allocated fixed-precision integers ([#221])
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -112,6 +112,23 @@ pub trait RandomMod: Sized + Zero {
     fn random_mod(rng: &mut impl CryptoRngCore, modulus: &NonZero<Self>) -> Self;
 }
 
+/// Random prime number generation support.
+#[cfg(feature = "rand_core")]
+pub trait RandomPrime: Random {
+    /// Generate a cryptographically secure random number.
+    ///
+    /// This function uses Miller-Rabin primality test with k=50 rounds
+    /// so the probability of returning a composite number is 4^(-50).
+    /// Note this method doesn't check whether the generated number
+    /// is a safe prime (i.e. that (p-1)/2 is also prime).
+    ///
+    /// The variable-time nature of the algorithm should not pose a security
+    /// issue so long as the underlying random number generator is truly a
+    /// CSRNG, where previous outputs are unrelated to subsequent
+    /// outputs and do not reveal information about the RNG's internal state.
+    fn random_prime(rng: &mut impl CryptoRngCore) -> Self;
+}
+
 /// Compute `self + rhs mod p`.
 pub trait AddMod<Rhs = Self> {
     /// Output type.
--- a/src/uint.rs
+++ b/src/uint.rs
@@ -23,6 +23,7 @@ mod mul;
 mod mul_mod;
 mod neg;
 mod neg_mod;
+mod prime;
 mod resize;
 mod shl;
 mod shr;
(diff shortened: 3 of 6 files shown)


tarcieri:
FYI, there's already a primes library built on `crypto-bigint` called `crypto-primes`: https://github.com/entropyxyz/crypto-primes

We don't mention it anywhere though so the discoverability is a little bad /cc @fjarri

maksimryndin:
> FYI, there's already a primes library built on `crypto-bigint` called `crypto-primes`: https://github.com/entropyxyz/crypto-primes
> 
> We don't mention it anywhere though so the discoverability is a little bad /cc @fjarri

Oh, I see, cool :)

fjarri:
It's listed on cryptography.rs now, but otherwise I'm not sure where else I can mention it to help people find it

tarcieri:
The `crypto-bigint` README would be a good start.

I’d also still be in favor of merging the two, which would definitely address the discoverability issue.

fjarri:
> I’d also still be in favor of merging the two, which would definitely address the discoverability issue.

I am not really in favor of that. It is well-isolated and moves at a different velocity. I wouldn't want to have to wait for a new `crypto-bigint` release to get access to some new feature.

tarcieri:
Okay cool, well we should add at least add a link in the README.md still.

That leaves the question of what to do with this PR.

fjarri:
(That said, I'm not opposed to moving `crypto-bigint` under the RustCrypto umbrella, but I think you were against adding another repo)

maksimryndin:
> Okay cool, well we should add at least add a link in the README.md still.
> 
> That leaves the question of what to do with this PR.

No worries, guys, I am absolutely ok if you're going to close it :) Would be grateful if you could make a quick code review for the sake of my learning :)

Anyway thanks for a quick response and useful links!

tarcieri:
Going to go ahead and close this, but thanks for the attempt!

> (That said, I'm not opposed to moving crypto-bigint under the RustCrypto umbrella, but I think you were against adding another repo)

@fjarri that's an interesting idea

fjarri:
> Would be grateful if you could make a quick code review for the sake of my learning :)

Sorry, can't justify going into much detail for a PR that's not going to be merged, but in general:
- The performance without pre-sieving will be pretty bad. Especially when finding safe primes. And when I say bad I mean pretty much unusable for practical primes sizes (say, 1024 bits) :)
- A sieving iterator (so that you generate a random start and search for primes starting from it) improves performance even more, but may lead to biased distribution (although both GMP and OpenSSL use it). 
- BPSW test (MR + Lucas) is the current state of the art for primality testing, and unlike pure MR is believed to not have false positives (at least for the practical bit sizes). OpenSSL uses pure MR, but only because they have a super-optimized MR method.
- The default 50 rounds is neither here nor there. GMP uses 25, FIPS requires 64-128. 
- The use of inner functions (`get_random_base()` and `mod_pow()`) doesn't seem justified to me, they could be just inlined.
- In the `for _ in 0..k` loop you repeatedly create `DynResidueParams` and cast integers in and out of the Montgomery representation. This is slow, you can just do it once and run the whole loop in Montgomery representation. 
- More tests would be good. MR has known false positives that it could be tested against (https://oeis.org/A001262). If you're testing a Carmichael number, you can check it properties with respect to specific bases.

maksimryndin:
> > Would be grateful if you could make a quick code review for the sake of my learning :)
> 
> Sorry, can't justify going into much detail for a PR that's not going to be merged, but in general:
> 
>     * The performance without pre-sieving will be pretty bad. Especially when finding safe primes. And when I say bad I mean pretty much unusable for practical primes sizes (say, 1024 bits) :)
> 
>     * A sieving iterator (so that you generate a random start and search for primes starting from it) improves performance even more, but may lead to biased distribution (although both GMP and OpenSSL use it).
> 
>     * BPSW test (MR + Lucas) is the current state of the art for primality testing, and unlike pure MR is believed to not have false positives (at least for the practical bit sizes). OpenSSL uses pure MR, but only because they have a super-optimized MR method.
> 
>     * The default 50 rounds is neither here nor there. GMP uses 25, FIPS requires 64-128.
> 
>     * The use of inner functions (`get_random_base()` and `mod_pow()`) doesn't seem justified to me, they could be just inlined.
> 
>     * In the `for _ in 0..k` loop you repeatedly create `DynResidueParams` and cast integers in and out of the Montgomery representation. This is slow, you can just do it once and run the whole loop in Montgomery representation.
> 
>     * More tests would be good. MR has known false positives that it could be tested against (https://oeis.org/A001262). If you're testing a Carmichael number, you can check its properties with respect to specific bases.

@fjarri thanks a lot for a detailed response and a lot of comments to improve on!

PR #285, opened 2023-11-08
Make `Uint::random_mod()` work identically on 32- and 64-bit targets

fjarri:
The current version of `Uint::random_mod()` produces different results on 32- and 64-bit targets because it exhausts the byte stream differently if the number of bytes in `modulus` is not a multiple of 8 (e.g. if it is 20 bytes, the 32-bit version will read 20 bytes - 5 limbs, but the 64-bit version will read 24 bytes - 3 limbs). This PR makes the behavior identical, so that an RNG with the same seed will produce the same stream of `Uint`s (as long as the RNG itself produces the bytestream consistently, of course). 

I am not sure if that is actually a desired behavior, but I did bump into it when I generated a ZK proof challenge on a 64-bit and 32-bit clients, and expected them to be the same. If it is not a guarantee we want to provide, perhaps an explicit note in the docs will be helpful.

The code in this PR would be much simpler if I could use `Uint::from_le_bytes()`, but then I would have to add an `Encoding` bound.

### DIFF
--- a/src/uint/rand.rs
+++ b/src/uint/rand.rs
@@ -1,7 +1,7 @@
 //! Random number generator support
 
 use super::Uint;
-use crate::{Limb, NonZero, Random, RandomMod};
+use crate::{Encoding, Limb, NonZero, Random, RandomMod};
 use rand_core::CryptoRngCore;
 use subtle::ConstantTimeLess;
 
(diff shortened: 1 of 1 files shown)


fjarri:
Not sure why the `powerpc` job failed, it has been queued for 20 minutes previously waiting for the job runner. I don't think I have the necessary permission to restart it.

tarcieri:
Looks like a random GitHub Actions runner failure. I restarted it.

src/uint/rand.rs
```diff
@@ -30,18 +30,29 @@ impl<const LIMBS: usize> RandomMod for Uint<LIMBS> {
...
-            for i in 0..n_limbs {
-                n.limbs[i] = Limb::random(&mut rng);
+            for i in 0..n_limbs - 1 {
+                rng.fill_bytes(bytes.as_mut());
+                // Need to deserialize from little-endian to make sure that two 32-bit limbs
+                // deserialized sequentially are equal to one 64-bit limb produced from the same
+                // byte stream.
+                n.limbs[i] = Limb::from_le_bytes(bytes);
             }
-            n.limbs[n_limbs - 1] = n.limbs[n_limbs - 1] & mask;
```
tarcieri, line 44: Have you evaluated the performance impact of removing the masking?

fjarri: I don't think the impact is noticeable compared to replacing `next_u64()/next_u32()` (which were called inside `Limb::random()`) with `fill_bytes()`. I need to be sure I'm drawing the same number of bytes from the RNG on every platform, so masking does not really fit into the algorithm.

fjarri: Now, an RNG implementation may actually skip bytes differently on every platform between successive calls to `fill_bytes()` - this is allowed as long as the distribution remains random. So this implementation is only consistent as long as the RNG is, and an RNG is not guaranteed to be, so perhaps we shouldn't provide that guarantee at all.

In my application I am now using an XOF instead, anyway; this PR is only to protect other uses from the same issue I encountered, but perhaps it's better to make an explicit disclaimer in the docs instead.

tarcieri (approved):

PR #286, opened 2023-11-10
Update docs for consistency

drcapybara:
The front of the repo explicitly states under Goals:

"Constant-time by default. Variable-time functions are explicitly marked as such."

Some constant-time functions in the repo are documented that they run in constant-time, while other constant-time functions do not. This inconsistency may lead a newcomer to question if a function is constant-time despite being explicitly chosen as such. This simple PR updates the docs on the functions in question to avoid confusion and remain consistent with the stated goals of the repo.


tarcieri (approved):

tarcieri:
I think there may be remaining places, but this is a start

PR #365, opened 2023-11-30
Round up `bits_precision` when creating `BoxedUint`

### LINKED ISSUE #258: `nlimbs!()` macro doesn't support rounding up to the limb size
The `nlimbs!` macro assumes the bit-size provided is a multiple of the limb bit-size. This is reflected in all the tests for the macro and its documentation.

This falls down for examples like `nlimbs!(63) = 0` (should be 1), and `nlimbs!(65) = 1` (should be 2) on a 64-bit platform.

fjarri:
I understand that this change may be controversial, but let me just put it out there. The reason being, why panic when we don't have to, and also to prevent bugs when something works on a 32 bit but doesn't on a 64 bit target.

- `BoxedUint::zero_with_precision()`, `one_with_precision()`, `max()` round up `bits_precision` to a multiple of `Limb::BITS`. It's a little weird in the case of `max()`, but I am not sure this method should be public in the first place (and perhaps should be renamed). 
- Support arbitrary `bits_precision` in `BoxedUint::random()` (generating a random number in range `[0, 2^bits_precision)`). 
- Fix the tests in `boxed/rand.rs` which didn't actually test `BoxedUint`. 
- Round up `bits_precision` when decoding `BoxedUint` from bytes, to match the constructors' behavior.
- Round up `nlimbs()` in the same way. Fixes #258.

An option here is to keep the given `bits_precision` as a field in the `BoxedUint` without rounding it up. Then we will have to zeroize MSBs in arithmetic operations.

### DIFF
--- a/src/limb/encoding.rs
+++ b/src/limb/encoding.rs
@@ -38,9 +38,10 @@ impl Limb {
     pub(crate) fn from_be_slice(bytes: &[u8]) -> Self {
         let mut repr = Self::ZERO.to_be_bytes();
         let repr_len = repr.len();
-        if bytes.len() > repr_len {
-            panic!("The given slice is larger than the limb size");
-        }
+        assert!(
+            bytes.len() <= repr_len,
+            "The given slice is larger than the limb size"
+        );
         repr[(repr_len - bytes.len())..].copy_from_slice(bytes);
         Self::from_be_bytes(repr)
     }
@@ -51,9 +52,10 @@ impl Limb {
     pub(crate) fn from_le_slice(bytes: &[u8]) -> Self {
         let mut repr = Self::ZERO.to_le_bytes();
         let repr_len = repr.len();
-        if bytes.len() > repr_len {
-            panic!("The given slice is larger than the limb size");
-        }
+        assert!(
+            bytes.len() <= repr_len,
+            "The given slice is larger than the limb size"
+        );
         repr[..bytes.len()].copy_from_slice(bytes);
         Self::from_le_bytes(repr)
     }
--- a/src/limb/encoding.rs
+++ b/src/limb/encoding.rs
@@ -34,16 +34,28 @@ impl Encoding for Limb {
 impl Limb {
     /// Decode limb from a big endian byte slice.
     ///
-    /// Panics if the slice is not the same size as [`Limb::Repr`].
+    /// Panics if the slice is larger than [`Limb::Repr`].
     pub(crate) fn from_be_slice(bytes: &[u8]) -> Self {
-        Self::from_be_bytes(bytes.try_into().expect("slice not limb-sized"))
+        let mut repr = Self::ZERO.to_be_bytes();
+        let repr_len = repr.len();
+        if bytes.len() > repr_len {
+            panic!("The given slice is larger than the limb size");
+        }
+        repr[(repr_len - bytes.len())..].copy_from_slice(bytes);
+        Self::from_be_bytes(repr)
     }
 
     /// Decode limb from a little endian byte slice.
     ///
     /// Panics if the slice is not the same size as [`Limb::Repr`].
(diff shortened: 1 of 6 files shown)


src/modular/boxed_residue/add.rs
```diff
@@ -63,6 +63,7 @@ impl AddAssign<BoxedResidue> for BoxedResidue {
 }
 
 #[cfg(test)]
+#[allow(clippy::unwrap_used)]
```
tarcieri, line 66: This seems superfluous (as do a few others)

fjarri: It's not though, there's `unwrap()` right below in `add_overflow()`.

tarcieri: It passes for me locally as well as in CI. Do you have some local configuration overridden that checks for `unwrap()` in tests?

fjarri: I'm using Rust 1.73, and CI uses 1.72, perhaps that's the reason?

fjarri: Ah also CI doesn't have `--tests` in the `clippy` call

tarcieri:
```
$ rustc --version
rustc 1.74.0 (79e9716c9 2023-11-13)
$ cargo clippy
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
```

tarcieri: If I add `--tests` I get several warnings about the tests, but we usually don't run clippy on tests

fjarri: Ok, no problem, I usually do. These clippy fixes are all in the tests, should I roll them back?

tarcieri: Yeah (or at least file it in a separate PR)

fjarri:
Should I add rounding up there too? I think it is self-evident that `nlimbs!(127)` does not produce an `U127`, but instead gives the number of limbs enough to fit 127 bits.

src/modular/boxed_residue/add.rs
fjarri: Rolled back.

tarcieri:
> Should I add rounding up there too? 

@fjarri sure, sounds good

fjarri:
Done

tarcieri:
@fjarri this needs a rebase now

tarcieri:
@fjarri this needs a rebase now

fjarri:
Rebased

src/limb/encoding.rs
```diff
@@ -34,16 +34,28 @@ impl Encoding for Limb {
...
     /// Decode limb from a little endian byte slice.
     ///
     /// Panics if the slice is not the same size as [`Limb::Repr`].
     pub(crate) fn from_le_slice(bytes: &[u8]) -> Self {
-        Self::from_le_bytes(bytes.try_into().expect("slice not limb-sized"))
+        let mut repr = Self::ZERO.to_le_bytes();
+        let repr_len = repr.len();
+        if bytes.len() > repr_len {
+            panic!("The given slice is larger than the limb size");
+        }
```
tarcieri, line 56:
```suggestion
    assert!(bytes.len() <= repr_len, "The given slice is larger than the limb size");
```

src/uint/boxed/encoding.rs
```diff
@@ -8,10 +8,10 @@ use core::fmt;
 /// Decoding errors for [`BoxedUint`].
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum DecodeError {
-    /// Input is not a valid size.
+    /// Input size is too small to fit in the given precision.
```
tarcieri, line 11: Might need to update the text in the `Display` impl

fjarri: Done

PR #372, opened 2023-12-02
Bernstein-Yang modular inversion algorithm

tarcieri:
Adapted from: https://github.com/privacy-scaling-explorations/halo2curves/pull/83

Original code is Apache 2.0+MIT. Attribution has been added to the top of the module.

See also: #227

cc @dignifiedquire @fjarri @mratsim

### DIFF
--- a/src/modular.rs
+++ b/src/modular.rs
@@ -21,6 +21,7 @@ mod reduction;
 mod residue;
 
 mod add;
+mod bernstein_yang;
 mod div_by_2;
 mod inv;
 mod mul;
@@ -31,6 +32,7 @@ mod sub;
 pub(crate) mod boxed_residue;
 
 pub use self::{
+    bernstein_yang::BYInverter,
     dyn_residue::{DynResidue, DynResidueParams},
     reduction::montgomery_reduction,
     residue::{Residue, ResidueParams},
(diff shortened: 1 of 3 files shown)


tarcieri:
Hmm, re: 32-bit support, I was attempting to go off of what section 12.3 of https://gcd.cr.yp.to/safegcd-20190413.pdf had to say regarding the number of bits/iterations, but it seems something else is amiss

src/limb.rs
```diff
@@ -45,6 +45,10 @@ pub type WideWord = u64;
 #[cfg(target_pointer_width = "32")]
 pub(crate) type SignedWord = i32;
 
+/// Signed equivalent of a wide word.
+#[cfg(target_pointer_width = "32")]
+pub(crate) type SignedWideWord = i32;
```
dignifiedquire, line 50: pretty sure this should be `i64`

src/limb.rs
```diff
@@ -61,6 +65,10 @@ pub type WideWord = u128;
 #[cfg(target_pointer_width = "64")]
 pub(crate) type SignedWord = i64;
 
+/// Signed equivalent of a wide word.
+#[cfg(target_pointer_width = "64")]
+pub(crate) type SignedWideWord = i64;
```
dignifiedquire, line 70: pretty sure this should be `i128`

PR #373, opened 2023-12-02
Make bit ops use `u32` for shifts

fjarri:
A part of #268:
- make all shift arguments in various bit shifts use `u32`
- the corresponding constants (e.g. `BITS`) and methods (e.g. `bit_precision()`) return `u32`. Note that the methods and constants that are used for indexing (`LIMBS` and `BYTES`) are still `usize`.
- added some methods to `CtChoice` (working on `u32` arguments) and normalized their names.

Notes:
- I also added `inline` to all the `CtChoice` methods which bumped performance quite a bit in my tests (tens of percents in various benchmarks). 
- I tried using `u8` and `u32` in `CtChoice`, but it seems to marginally decrease performance, so `Word` stays for now.

### DIFF
--- a/src/ct_choice.rs
+++ b/src/ct_choice.rs
@@ -15,63 +15,132 @@ impl CtChoice {
     /// The truthy value.
     pub const TRUE: Self = Self(Word::MAX);
 
+    #[inline]
+    pub(crate) const fn as_u32_mask(&self) -> u32 {
+        #[allow(trivial_numeric_casts)]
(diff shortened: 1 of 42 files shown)


tarcieri:
I'm definitely in favor of this since I found our use of `usize` oddly inconsistent with `u32`. Also as-is `subtle` has no trait impls for `usize` which meant casting back to `u32` from `usize` (which is not guaranteed to be lossless).

Looks like it needs a rebase though.

fjarri:
Rebased

PR #375, opened 2023-12-02
Add `std` feature

tarcieri:
- Impl `std::error::Error` for `DecodeError`
- Use `Arc` for storing `BoxedResidueParams` when available


fjarri:
Out of curiosity, what is it for?

tarcieri:
Oh, using `Arc` means that values computed from other `BoxedResidues` can share the same `BoxedResidueParams` as the original value without cloning it each time

PR #376, opened 2023-12-02
Constant-time square root and division

tarcieri:
This is the implementation from #277, rebased and with most of the review comments addressed.

The ones I didn't address are:

- [still a bit confused about constant-time division, left comments](https://github.com/RustCrypto/crypto-bigint/pull/277/files/6e6c5442bde4f1f3e1860bc2a9e4fc2f425325ab#diff-62d1bf839c2bbb8be6ed2db8a80b297aabc227fcb68d58f9e6d1521a04a7fa0e)
- [mismatch between paper and implementation?](https://github.com/RustCrypto/crypto-bigint/pull/277#discussion_r1351040253)
- [testing improvement suggestion](https://github.com/RustCrypto/crypto-bigint/pull/277#pullrequestreview-1666138090)

This was already getting a bit hard to rebase, so if we can get it over the finish line before it becomes too stale that seems good.

### DIFF
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -47,11 +47,47 @@ impl<const LIMBS: usize> Uint<LIMBS> {
     /// NOTE: Use only if you need to access const fn. Otherwise use [`Self::div_rem`] because
     /// the value for is_some needs to be checked before using `q` and `r`.
     ///
+    /// This function is constant-time with respect to both `self` and `rhs`.
+    pub(crate) const fn const_div_rem(&self, rhs: &Self) -> (Self, Self, CtChoice) {
+        let mb = rhs.bits();
+        let mut rem = *self;
+        let mut quo = Self::ZERO;
+        let mut c = rhs.shl(Self::BITS - mb);
+
+        let mut i = Self::BITS;
+        let mut done = CtChoice::FALSE;
+        loop {
+            let (mut r, borrow) = rem.sbb(&c, Limb::ZERO);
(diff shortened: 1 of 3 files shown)


src/uint/div.rs
```diff
@@ -1,5 +1,7 @@
 //! [`Uint`] division operations.
 
+#![allow(trivial_numeric_casts)]
```
fjarri, line 3: Can it be more targeted?

tarcieri: Moved to `Uint::const_div_rem` in [60fb612](https://github.com/RustCrypto/crypto-bigint/pull/376/commits/60fb612ca0446f80dc293d14d345cdb8f67cdbc3)

fjarri:
A test for `U192` (to check a non-power-of-2 bit size) can be, for example:
```
x = 131760037584061584202147902974795492358573409165975616061
r = 11478677518950586156071689761
where x = (r + 1)^2 - 583
```
This also uses up the maximum number of iterations

tarcieri:
@fjarri pushed up the `U192` and `U256` edge cases tests you suggested in [927ebaf](https://github.com/RustCrypto/crypto-bigint/pull/376/commits/927ebafbd525a617f79b3c957f915c728931efda)

Wasn't sure if you were suggesting translating the Python example into Rust as well.

fjarri:
No, the Python code is just to document what I was using for cross-checking. I'm approving this, you can merge, and then I'll make a separate PR for the items from https://github.com/RustCrypto/crypto-bigint/pull/376#issuecomment-1837272800

fjarri (approved):

PR #387, opened 2023-12-04
Revert "BoxedUint: improve efficiency of `square` (#382)"

tarcieri:
This reverts commit cd9cdcecf9cd27fae48843905581b053c31952e0.

Benchmarks reveal this caused a performance regression to modpow. Reverting it yields the following improvement:

```
                        time:   [76.362 µs 76.381 µs 76.403 µs]
                        change: [-72.846% -72.795% -72.747%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/uint/boxed/mul.rs
+++ b/src/uint/boxed/mul.rs
@@ -1,9 +1,6 @@
 //! [`BoxedUint`] multiplication operations.
 
-use crate::{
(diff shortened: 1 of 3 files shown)


tarcieri:
Still need to investigate why this caused as much performance degradation as it did

PR #395, opened 2023-12-05
Bring the overflow behavior in bit shifts in sync with `std`

### LINKED ISSUE #121: Diverging from primitive behavior in overflowing shift
The rust docs say the following about overflowing shift:
> If the shift value is too large, then value is masked (N-1) where N is the number of bits, and this value is then used to perform the shift.
(Source: https://doc.rust-lang.org/std/primitive.u64.html#method.overflowing_shl)
And you can see this via the following code:
```rust
#[allow(arithmetic_overflow)]
fn main() {
    let a: u64 = 500;
    let b = a << 119;
    println!("{}", b);
}
```
Which prints `18014398509481984000`.

On the other hand in the implementation here zeros will be returned: https://github.com/RustCrypto/crypto-bigint/blob/36dbb69a73cfdba1eef90e10a8424cde39602e3b/src/uint/shl.rs#L18

fjarri:
A part of #268 
Fixes #121

- `const fn` bit shifts for `Uint` return the overflow status as `CtChoice` (and set the result to zero in that case, which is documented, so it's a part of the API now). `Option` would be better for the vartime shifts, but its methods are not `const` yet in stable.
- `shl/shr` for `BoxedUint` return `(Self, Choice)` (not `CtOption` since most of its methods need the type to be `ConditionallySelectable`, which `BoxedUint` isn't). The vartime equivalents return `Option<Self>`.
- operator impls panic on overflow (which is the default behavior for built-in integers)
- made the implementations in `uint/shl.rs` and `shr.rs` more uniform and improved vartime shift performance (before it was calling a constant-time shift-by-no-more-than-limb which added some overhead)  
- improved constant-time shift performance for `BoxedUint` by reducing the amount of allocations
- added an optimized `BoxedUint::shl1()` implementation
- added some inlines for `Limb` methods which improved shift performance noticeably
- added more benchmarks for shifts and simplify benchmark hierarchy a little (create test group directly in the respective function)
- fixed an inefficiency in `Uint` shifts: we need to iterate to log2(BITS-1), not log2(BITS), because that's the maximum size of the shift.
- Renamed `sh(r/l)1_with_overflow()` to `sh(r/l)1_with_carry` to avoid confusion - in the context of shifts we call the shift being too large an overflow.

### DIFF
--- a/src/uint/boxed/shl.rs
+++ b/src/uint/boxed/shl.rs
@@ -1,43 +1,54 @@
 //! [`BoxedUint`] bitwise left shift operations.
 
-use crate::{BoxedUint, CtChoice, Limb, Word};
+use crate::{BoxedUint, Limb};
 use core::ops::{Shl, ShlAssign};
 use subtle::{Choice, ConstantTimeLess};
 
 impl BoxedUint {
     /// Computes `self << shift`.
-    /// Returns zero if `shift >= Self::BITS`.
-    pub fn shl(&self, shift: u32) -> Self {
+    /// Returns `None` if `shift >= Self::BITS`.
+    pub fn shl(&self, shift: u32) -> (Self, Choice) {
+        // `floor(log2(bits_precision - 1))` is the number of bits in the representation of `shift`
+        // (which lies in range `0 <= shift < bits_precision`).
+        let shift_bits = u32::BITS - (self.bits_precision() - 1).leading_zeros();
         let overflow = !shift.ct_lt(&self.bits_precision());
         let shift = shift % self.bits_precision();
-        let log2_bits = u32::BITS - self.bits_precision().leading_zeros();
         let mut result = self.clone();
 
-        for i in 0..log2_bits {
+        for i in 0..shift_bits {
             let bit = Choice::from(((shift >> i) & 1) as u8);
-            result = Self::conditional_select(&result, &result.shl_vartime(1 << i), bit);
+            result = Self::conditional_select(
+                &result,
+                // Will not overflow by construction
+                &result.shl_vartime(1 << i).expect("shift within range"),
+                bit,
+            );
         }
 
-        Self::conditional_select(
-            &result,
-            &Self::zero_with_precision(self.bits_precision()),
+        (
+            Self::conditional_select(
+                &result,
+                &Self::zero_with_precision(self.bits_precision()),
+                overflow,
+            ),
             overflow,
         )
     }
 
     /// Computes `self << shift`.
+    /// Returns `None` if `shift >= Self::BITS`.
     ///
     /// NOTE: this operation is variable time with respect to `shift` *ONLY*.
     ///
     /// When used with a fixed `shift`, this function is constant-time with respect to `self`.
     #[inline(always)]
-    pub fn shl_vartime(&self, shift: u32) -> Self {
+    pub fn shl_vartime(&self, shift: u32) -> Option<Self> {
+        if shift >= self.bits_precision() {
+            return None;
+        }
+
         let nlimbs = self.nlimbs();
         let mut limbs = vec![Limb::ZERO; nlimbs].into_boxed_slice();
-
-        if shift >= self.bits_precision() {
-            return Self { limbs };
-        }
 
         let shift_num = (shift / Limb::BITS) as usize;
         let rem = shift % Limb::BITS;
@@ -48,39 +59,27 @@ impl BoxedUint {
             limbs[i] = self.limbs[i - shift_num];
         }
 
-        let (new_lower, _carry) = (Self { limbs }).shl_limb(rem);
-        new_lower
-    }
-
-    /// Computes `self << shift` where `0 <= shift < Limb::BITS`,
-    /// returning the result and the carry.
-    #[inline(always)]
-    pub(crate) fn shl_limb(&self, shift: u32) -> (Self, Limb) {
-        let nlimbs = self.nlimbs();
-        let mut limbs = vec![Limb::ZERO; nlimbs].into_boxed_slice();
-
-        let nz = CtChoice::from_u32_nonzero(shift);
-        let lshift = shift;
-        let rshift = nz.if_true_u32(Limb::BITS - shift);
-        let carry = nz.if_true_word(self.limbs[nlimbs - 1].0.wrapping_shr(Word::BITS - shift));
-
-        let mut i = nlimbs - 1;
-        while i > 0 {
-            let mut limb = self.limbs[i].0 << lshift;
-            let hi = self.limbs[i - 1].0 >> rshift;
-            limb |= nz.if_true_word(hi);
-            limbs[i] = Limb(limb);
-            i -= 1
+        if rem == 0 {
+            return Some(Self { limbs });
         }
-        limbs[0] = Limb(self.limbs[0].0 << lshift);
 
-        (Self { limbs }, Limb(carry))
+        let mut carry = Limb::ZERO;
+
+        while i < nlimbs {
+            let shifted = limbs[i].shl(rem);
+            let new_carry = limbs[i].shr(Limb::BITS - rem);
+            limbs[i] = shifted.bitor(carry);
+            carry = new_carry;
+            i += 1;
+        }
+
+        Some(Self { limbs })
     }
 
     /// Computes `self >> 1` in constant-time.
     pub(crate) fn shl1(&self) -> Self {
         // TODO(tarcieri): optimized implementation
-        self.shl_vartime(1)
+        self.shl_vartime(1).expect("shift within range")
     }
 
     /// Computes `self >> 1` in-place in constant-time.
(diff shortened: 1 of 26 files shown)


src/uint/shl.rs
```diff
@@ -94,23 +104,28 @@ impl<const LIMBS: usize> Uint<LIMBS> {
     /// Computes `self >> 1` in constant-time.
     pub(crate) const fn shl1(&self) -> Self {
         // TODO(tarcieri): optimized implementation
```
tarcieri, line 106: Sidebar, but I've been wondering if we could implement this using a more optimized overflowing shift on the level of each limb, so as to better optimize carry bit handling (are there e.g. CPU flags which make retrieving overflow possible?)

tarcieri:
I think these are reasonably OK things to bring in line with `std`

src/uint/shl.rs
fjarri: I tried that (using `overflowing_shift()`, which is supposed to compile down to a shift + flag checking), and it doesn't really change the performance. Plus I am more confident that the explicit overflow will compile into constant-time code, because if there are architectures where there's no overflow flag, the compiler will probably just put in a condition.

tarcieri: Better guarantees of constant-time are a good point, but it might still be good to have something like `Limb::overflowing_sh*` to provide a common implementation in places where this is useful, as it would be here, regardless of how it's implemented.

In the implementation of the function on `Limb`, we could potentially decide on an architecture-by-architecture basis to use the primitive methods if we're happy with the generated assembly, and have the portable implementation fall back to one we hope will be mostly constant time-ish.

src/uint/boxed/shl.rs
```diff
@@ -1,108 +1,127 @@
...
     ///
     /// NOTE: this operation is variable time with respect to `shift` *ONLY*.
     ///
     /// When used with a fixed `shift`, this function is constant-time with respect to `self`.
     #[inline(always)]
-    pub fn shl_vartime(&self, shift: u32) -> Self {
-        let nlimbs = self.nlimbs();
-        let mut limbs = vec![Limb::ZERO; nlimbs].into_boxed_slice();
-
+    fn shl_vartime_into(&self, dest: &mut Self, shift: u32) -> Option<()> {
```
tarcieri, line 48: As a convention I'm a fan of putting this sort of output parameter on the end, at least as far as an output buffer which is simply written into otherwise uninitialized, though I will acknowledge that's inconsistent with `*_assign` APIs, which operate on `&mut self` (and might those be applicable here?)

fjarri:
Running the benchmarks vs `master`, and while most of them improved, there are some regressions in Monthomery ops with BoxedUints. Investigating now.

src/uint/boxed/shl.rs
fjarri: It's not exactly `assign`, since it writes in a new buffer. But I don't mind changing the argument order, it's a private method anyway.

fjarri:
Hm, I was comparing the wrong commits. But it seems that there was a massive performance drop in "Montgomery arithmetic/multiplication, BoxedUint*BoxedUint" testcase between b169d2fafba73359fbd22727ffd969b4a6f60f22 and 1a0399d4f1f72aabba720074be6893eebf08c39b

tarcieri:
@fjarri was it 207040790d48b44293f79b6a68abfbd814ee110b? That recalibrated the benchmarks using a larger input, so any of the earlier benchmarks aren't relevant, you'd have to revert the implementation and compare (but they were busted anyway, because they weren't using `black_box`)

fjarri:
Yep, I bisected to that point too, seems to be the reason.

src/uint/boxed/shl.rs
```diff
@@ -1,108 +1,127 @@
...
-use crate::{BoxedUint, CtChoice, Limb, Word};
+use crate::{BoxedUint, Limb};
 use core::ops::{Shl, ShlAssign};
 use subtle::{Choice, ConstantTimeLess};
 
 impl BoxedUint {
     /// Computes `self << shift`.
-    /// Returns zero if `shift >= Self::BITS`.
-    pub fn shl(&self, shift: u32) -> Self {
+    /// Returns `None` if `shift >= Self::BITS`.
```
tarcieri, line 9: Looks like some copypasta: `BoxedUint` doesn't have a `Self::BITS`, but instead a dynamic `self.bits_precision()` (and there are several comments of this nature in this PR)

src/uint/boxed/shl.rs
```diff
@@ -1,108 +1,127 @@
...
+    pub fn shl(&self, shift: u32) -> (Self, Choice) {
```
tarcieri, line 10: This could perhaps be:

```suggestion
    pub fn shl(&self, shift: u32) -> CtOption<Self> {
```

src/uint/boxed/shl.rs
```diff
@@ -1,108 +1,127 @@
...
 
-        Self::conditional_select(
-            &result,
-            &Self::zero_with_precision(self.bits_precision()),
+        (
+            Self::conditional_select(
+                &result,
+                &Self::zero_with_precision(self.bits_precision()),
+                overflow,
+            ),
```
tarcieri, line 34: While you're here, it seems like this could be replaced with a loop over the limbs which does `conditional_assign(&Limb::ZERO, overflow)` and thus avoids an allocation

src/uint/boxed/shl.rs
fjarri: I mentioned that in the description - the return type is not `CtOption` here since most of its methods need the type to be `ConditionallySelectable`, which `BoxedUint` isn't, so it would be pretty useless

tarcieri: There are some fake `CtOption` combinator methods defined on `BoxedUint`:

- https://docs.rs/crypto-bigint/0.6.0-pre.1/crypto_bigint/struct.BoxedUint.html#method.conditional_and_then
- https://docs.rs/crypto-bigint/0.6.0-pre.1/crypto_bigint/struct.BoxedUint.html#method.conditional_map

Though ideally this would get solved upstream on `subtle`: https://github.com/dalek-cryptography/subtle/pull/118

cc @isislovecruft

src/uint/boxed/shl.rs
fjarri: Added `BoxedUint::conditional_set_to_zero()` for that

src/uint/boxed/shl.rs
fjarri: Fixed

src/uint/boxed/shl.rs
tarcieri: Also there are other methods for the `Boxed*` types already using `CtOption`, so we should probably choose a convention for `BoxedUint`/`BoxedResidue` and stick to it

fjarri: The problem is that these methods *return* a `CtOption`, but there are no methods that *take* a `CtOption`, so if you end up with one, there is not much you can do besides unwrapping it in variable time. In one place in the codebase I need an analogue of `unwrap_or()`, how would I implement that?

tarcieri: Yeah, really this would ideally be solved upstream in `subtle`.

I think the `(Self, CtChoice)` tuples are a bit error-prone since they still expose the inner value even if `CtChoice` is false. If you wanted to define a similar analogue to `CtOption` which works for the `const` cases and `CtChoice`, that seems good to me

fjarri: > I think the (Self, CtChoice) tuples are a bit error-prone since they still expose the inner value even if CtChoice is false.

In the specific case of shifts we can return zero in case of overflow, which is a perfectly logical and often useful value (so sometimes you can just ignore the overflow flag). It is documented for `Uint` shifts, but not for `BoxedUint`, so if we document it there too, it should be relatively ok. 

> If you wanted to define a similar analogue to CtOption which works for the const cases and CtChoice, that seems good to me

I thought about it too. It would also be useful to distinguish the cases where `CtChoice` denotes an overflow, a carry, or that the result exists at all (like in inversions).

fjarri: If `BoxedUint` were to use this new const analogue of `CtOption` it would not be consisted with the rest of `BoxedUint` API that uses `subtle`, so I think for now a tuple is better.

tarcieri: Ok, let's go with this for now, though we should perhaps make a tracking issue to make all of the APIs consistent

fjarri: Filed #422 for the `CtOption` equivalent

src/uint/shr.rs
```diff
@@ -139,30 +156,58 @@ mod tests {
...
+        assert_eq!(x.shr_limb(0), (x, Limb::ZERO));
+        assert_eq!(
+            x.shr_limb(8),
+            (
+                U128::from_be_hex("0000112233445566778899aabbccddee"),
+                Limb(0xff << (Limb::BITS - 8))
+            )
         );
     }
+    */
```
tarcieri, line 212: What's up with this?

tarcieri (approved):
Looks fine now aside from the commented out test

src/uint/shr.rs
fjarri: Good catch, commented when I was trying out removing `shr_limb()`. The method is not there anymore, so the test is not necessary.

PR #399, opened 2023-12-06
`BoxedResidue::pow` using "almost Montgomery multiplication"

tarcieri:
Implements `BoxedResidue::pow` using the "almost Montgomery multiplication" method used by `num-bigint` as originally added in rust-num/num-bigint@f523b9c359b0b4e8718176aba693c6e061cd0353

### DIFF
--- a/benches/boxed_residue.rs
+++ b/benches/boxed_residue.rs
@@ -6,10 +6,15 @@ use crypto_bigint::{
     modular::{BoxedResidue, BoxedResidueParams},
     BoxedUint,
 };
+use num_bigint::BigUint;
 use rand_core::OsRng;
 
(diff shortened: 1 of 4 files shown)


tarcieri:
Though there are various discrepancies with the implementation in `num-bigint`, the larger problem seems to be the `BoxedResidue::pow` function is always returning zero (or at least, that's all I'm seeing in test failures thus far).

tarcieri:
Relevant benchmark:

```
                        time:   [24.333 ms 24.360 ms 24.389 ms]
                        change: [-38.998% -38.883% -38.776%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #403, opened 2023-12-07
[WIP] Expand multiplication methods

dignifiedquire:
Currently implements karatusba and schoolbook based on num-bigint-dig.


src/uint/add.rs
```diff
@@ -154,6 +154,49 @@ impl<const LIMBS: usize> AddAssign<&Checked<Uint<LIMBS>>> for Checked<Uint<LIMBS
...
+}
+
+#[inline]
+pub(crate) fn adc(a: Word, b: Word, acc: &mut WideWord) -> Word {
+    *acc += a as WideWord;
+    *acc += b as WideWord;
+    let lo = *acc as Word;
+    *acc >>= Word::BITS;
+    lo
+}
```
tarcieri, line 198: Feels like you could do much of this with `Limb` and its existing methods, and that it could probably just be an `AddAssign` impl on `BoxedUint`?

dignifiedquire: very likely, I just copy pasted the methods from num-bigint for the moment to get sth working, before searching through the lib

src/uint/mul.rs
```diff
@@ -9,6 +9,11 @@ use crate::{
 use core::ops::{Mul, MulAssign};
 use subtle::CtOption;
 
+#[cfg(feature = "alloc")]
+mod karatsuba;
```
tarcieri, line 13: Since this is gated under `alloc` and not shared with `Uint`, maybe just put it under `uint/boxed/karatsuba.rs`?

src/uint/mul/schoolbook.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+#[inline]
+fn mac_with_carry(a: Limb, b: Limb, c: Limb, acc: &mut WideWord) -> Limb {
+    *acc += a.0 as WideWord;
+    *acc += (b.0 as WideWord) * (c.0 as WideWord);
+    let lo = *acc as Word;
+    *acc >>= Word::BITS;
+    Limb(lo)
+}
```
tarcieri, line 38: Perhaps you could use the `impl_schoolbook_multiplication!` macro instead of this, so we just have one implementation?

src/uint/add.rs
```diff
@@ -99,6 +101,49 @@ impl<const LIMBS: usize> WrappingAdd for Uint<LIMBS> {
...
```
tarcieri, line 145: `Limb` has an `adc` method you can call directly.

Also this whole module feels like implementation details of Karatsuba. Maybe move it to `karatsuba.rs`?

src/uint/mul/schoolbook.rs
dignifiedquire: that didn’t work before, because I needed Wors, probably can go back now

src/uint/mul.rs
dignifiedquire: well my goal was to have this work with `Uint` before merging, just trying to get the first thing working

src/uint/add.rs
dignifiedquire: I am using parts of this in the `div` impl, that's why I moved it into the shared piece.

src/uint/mul.rs
tarcieri: Aah ok. Right now sharing code involves macros like `impl_schoolbook_multiplication!` since without `const_mut_refs` there's no way to construct anything other than a const generic-sized array in a `const fn`

src/limb/div.rs
```diff
@@ -0,0 +1,16 @@
+use crate::{WideWord, Word};
+
+/// Divide a two digit numerator by a one digit divisor, returns quotient and remainder:
+///
+/// Note: the caller must ensure that both the quotient and remainder will fit into a single digit.
+/// This is _not_ true for an arbitrary numerator/denominator.
+///
+/// (This function also matches what the x86 divide instruction does).
+#[inline]
+pub(crate) fn div_wide(hi: Word, lo: Word, divisor: Word) -> (Word, Word) {
```
tarcieri, line 10:
```suggestion
pub(crate) fn div_wide_vartime(hi: Word, lo: Word, divisor: Word) -> (Word, Word) {
```

src/uint/boxed/div.rs
```diff
@@ -106,29 +98,117 @@ impl BoxedUint {
...
+
+        debug_assert!(a < b);
+
+        (q.normalize_vartime(), a >> shift)
+    }
+
+    /// Division by a single word.
+    ///
+    /// This function operates in variable-time.
+    fn div_rem_digit(&self, b: Word) -> (Self, Word) {
```
tarcieri, line 201:
```suggestion
    fn div_rem_digit_vartime(&self, b: Word) -> (Self, Word) {
```

tarcieri:
Karatsuba was added in #649

PR #405, opened 2023-12-07
Make core `pow` implementation more like `num-bigint`

tarcieri:
Note that this approach introduces timing variability. It also doesn't seem to improve performance:

                        time:   [25.614 ms 25.642 ms 25.677 ms]
                        change: [-1.3757% -1.1214% -0.8876%] (p = 0.00 < 0.05)


tarcieri:
This was mostly pushed up for comparison purposes. Since it doesn't seem to help, closing.

PR #406, opened 2023-12-07
Implement `Zeroize` for `NonZero` wrapper

aumetra:
When attempting to port over `dsa` to `crypto-bigint`, I noticed that the `NonZero` wrapper has no implementation for the `Zeroize` trait.  

That made it a bit annoying to implement the internals that expect the private factor `x` to be wrapped into a `Zeroizing` wrapper.

### DIFF
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -367,6 +367,13 @@ impl<T: Serialize + Zero> Serialize for NonZero<T> {
     }
 }
 
+#[cfg(feature = "zeroize")]
+impl<T: zeroize::Zeroize + Zero> zeroize::Zeroize for NonZero<T> {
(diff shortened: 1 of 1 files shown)


tarcieri (approved):

tarcieri:
> When attempting to port over dsa to crypto-bigint...

@aumetra how's that going?

aumetra:
> how's that going?

Pretty good actually. Looking at the code, it should only leak public factors through the vartime operations needed.

The only thing I'm waiting for right now is `crypto-primes` for the key generation functions.

tarcieri:
@aumetra FYI, new `crypto-primes` prerelease is out: https://docs.rs/crypto-primes/0.6.0-pre.0/crypto_primes/

PR #409, opened 2023-12-10
Eliminate some timing variability in `montgomery_mul`

tarcieri:
Always performs `sub_vv` on the lower half of the output, and conditionally assigns `upper` to `lower` using `subtle`.

It doesn't matter if we unconditionally subtract, since the output will be overwritten in the case where it's "unnecessary" (but actually necessary for constant-time operation).

There are still some data-dependent branches, however, which have been annotated with a TODO.

### DIFF
--- a/src/modular/boxed_residue/mul.rs
+++ b/src/modular/boxed_residue/mul.rs
@@ -11,6 +11,7 @@ use core::{
     borrow::Borrow,
     ops::{Mul, MulAssign},
 };
+use subtle::{ConditionallySelectable, ConstantTimeEq};
 
 #[cfg(feature = "zeroize")]
 use zeroize::Zeroize;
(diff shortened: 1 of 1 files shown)


src/modular/boxed_residue/mul.rs
```diff
@@ -208,15 +211,17 @@ pub(crate) fn montgomery_mul(z: &mut [Word], x: &[Word], y: &[Word], m: &[Word],
         let cx = c.wrapping_add(c2);
         let cy = cx.wrapping_add(c3);
         z[n + i] = cy;
+
+        // TODO(tarcieri): eliminate data-dependent branches
```
tarcieri, line 215: I left this to handle in a followup because the performance hit seems substantial (~17%), whereas this PR itself seems to have only a slight (~1%) impact.

src/modular/boxed_residue/mul.rs
```diff
@@ -186,20 +187,22 @@ impl Drop for MontgomeryMultiplier<'_> {
...
     // It also assumes that x, y are already reduced mod m, or else the result will not be properly
     // reduced.
-    let mut c: Word = 0;
     let n = m.len();
+    debug_assert_eq!(z.len(), n * 2);
+    debug_assert_eq!(x.len(), n);
+    debug_assert_eq!(y.len(), n);
+    debug_assert_eq!(m.len(), n);
 
-    let pre_cond = z.len() > n && z[n..].len() == n && x.len() == n && y.len() == n && m.len() == n;
```
dignifiedquire, line 199: Why did you remove this, @newpavlov suggested this in zulip, as this removes bounds checks from the code later down the line. Now this will not happen as the `debug_assert`s don't get compiled in release mode

newpavlov: It's probably worth to add a comment explaining this together with `z.len() > n && z[n..].len() == n` intentionally being used instead of `z.len() == 2*n`, since the latter prevents compiler from eliminating some bound checks.

tarcieri: I tried reverting it and it had no observable change on benchmarks, FWIW

PR #410, opened 2023-12-10
BoxedUint: use `Limb`s in `montgomery_mul`

tarcieri:
This was previously written using raw `Word`s to ensure maximum performance, however it precludes access to the constant-time helper functions implemented on `Limb`.

This carefully rewrites the implementation to using `Limb` instead of `Word` while avoiding any performance impacts.

### Benchmarks (unchanged)

```
                        time:   [24.424 ms 24.444 ms 24.465 ms]
                        change: [-0.1629% -0.0399% +0.0760%] (p = 0.52 > 0.05)
```

### DIFF
--- a/src/modular/boxed_residue/mul.rs
+++ b/src/modular/boxed_residue/mul.rs
@@ -6,12 +6,12 @@
 //! Originally (c) 2014 The Rust Project Developers, dual licensed Apache 2.0+MIT.
 
 use super::{BoxedResidue, BoxedResidueParams};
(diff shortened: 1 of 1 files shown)


tarcieri:
@dignifiedquire I suspect what was hitting us before were the places there are currently branches, namely the comparisons between words/limbs. I suspect that `subtle` has a rather slow constant-time impl of integer comparisons which made things massively slower.

The data-dependent branching still needs to be addressed but I have some ideas.

dignifiedquire:
cool, I am glad this is possilbe

PR #413, opened 2023-12-10
Limb: optimize constant-time comparisons

tarcieri:
Optimizes `ConstantTimeGreater`/`ConstantTimeLess` impls by using borrowing subtraction and checking whether a borrow occurred.

### Benchmarks

```
ops/ct_lt               time:   [900.95 ps 909.83 ps 922.37 ps]
                        change: [-76.095% -75.752% -75.373%] (p = 0.00 < 0.05)
                        Performance has improved.
```

```
ops/ct_gt               time:   [902.46 ps 938.83 ps 1.0134 ns]
                        change: [-55.763% -53.637% -50.192%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/limb/cmp.rs
+++ b/src/limb/cmp.rs
@@ -44,19 +44,26 @@ impl ConstantTimeEq for Limb {
     fn ct_eq(&self, other: &Self) -> Choice {
         self.0.ct_eq(&other.0)
     }
+
+    #[inline]
+    fn ct_ne(&self, other: &Self) -> Choice {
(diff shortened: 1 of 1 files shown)


fjarri:
Is it a better algorithm than the one used in `CtChoice`? We should pick one and use it in both places.

tarcieri:
Definitely worth comparing options, as this is a major bottleneck for #415

fjarri:
Filed #418

PR #419, opened 2023-12-12
Make division methods take `NonZero`-wrapped divisors

fjarri:
This simplifies the API allowing us to avoid returning `CtChoice` results, and making some methods redundant. Also that's what `BoxedUint` is already doing.

- Simplify `Reciprocal` API by taking a `NonZero<Limb>` in `Reciprocal::new()` which is now `const`.
- `Reciprocal::ct_new()` is removed. 
- `Uint::div_rem_limb()` and `Uint::div_rem_limb_with_reciprocal()` take a `NonZero<Limb>` and are now `const`.
- `Uint::ct_div_rem_limb()` and `Uint::ct_div_rem_limb_with_reciprocal()` are removed.
- `Uint::div_rem()` and `Uint::div_rem_vartime()` are now `const`.
- `Uint::const_div_rem()` and `Uint::const_div_rem_vartime()` are removed.
- Make `Uint::rem()` and `rem_wide()` take `NonZero` rhs and make them const.
- Make `Uint::wrapping_div(_vartime)` take `NonZero` rhs
- Wrap `ResidueParams::MODULUS` in `NonZero`.
- `Residue::new_checked()` removed since `ResidueParams` has already done that check in compile time.
- Added a `const fn` `NonZero::as_ref()` to get access to the inner value in a const context.

### DIFF
--- a/src/modular/dyn_residue.rs
+++ b/src/modular/dyn_residue.rs
@@ -78,7 +78,7 @@ impl<const LIMBS: usize> DynResidueParams<LIMBS> {
         P: ResidueParams<LIMBS>,
     {
         Self {
-            modulus: P::MODULUS,
+            modulus: P::MODULUS.0,
             r: P::R,
             r2: P::R2,
             r3: P::R3,
--- a/src/modular/dyn_residue.rs
+++ b/src/modular/dyn_residue.rs
@@ -13,8 +13,8 @@ use super::{
     residue::{Residue, ResidueParams},
     Retrieve,
 };
-use crate::{Integer, Limb, NonZero, Uint, Word};
-use subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
+use crate::{Limb, NonZero, Uint, Word, Zero};
+use subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
 
 /// Parameters to efficiently go to/from the Montgomery form for an odd modulus provided at runtime.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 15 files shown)


tarcieri:
@fjarri if you'd like to get this one merged next, it needs a rebase

fjarri:
rebased

src/modular/dyn_residue.rs
```diff
@@ -35,13 +35,19 @@ pub struct DynResidueParams<const LIMBS: usize> {
 impl<const LIMBS: usize> DynResidueParams<LIMBS> {
     /// Instantiates a new set of `ResidueParams` representing the given `modulus` if it is odd.
     ///
-    /// Returns a `CtOption` that is `None` if the provided modulus is not odd.
-    pub fn new(modulus: &Uint<LIMBS>) -> CtOption<Self> {
-        let r = Uint::MAX.const_rem(modulus).0.wrapping_add(&Uint::ONE);
-        let r2 = Uint::const_rem_wide(r.square_wide(), modulus).0;
+    /// Returns `None` if the provided modulus is not odd.
+    pub fn new(modulus: &Uint<LIMBS>) -> Option<Self> {
```
tarcieri, line 39: Why no `CtOption` here?

fjarri: I didn't feel like `CtOption` added any security there, because what are you going to do with the `None` value besides returning an error? It doesn't really make sense to chain the result into anything. And with `BoxedUint`, you actually cannot chain it because of that `subtle` problem.

But I can change it back to `CtOption` for the sake of uniformity.

fjarri: Rolled back

tarcieri: Note I don't actually have any use cases which require a secret modulus, so perhaps it's not worth providing.

The reason you'd use `CtOption` in a hypothetical protocol which required one would be to follow it up with `map` and perform some subsequent operation before handling the error.

fjarri: You can do it with an `Option`, if you're going to return an error anyway, you don't need constant-timeness. But both boxed and stack residues should have the same API, so I don't mind leaving this return type alone for now.

tarcieri: You might handle the error via a rejection symbol, but again, this is spitballing about a problem I don't actually have (i.e. secret modulus)

fjarri: I do use secret moduli (in some helper methods related to Paillier encryption), but they are guaranteed to be odd, so no information is revealed this way (and if it was, it's only one bit anyway)

tarcieri: In that case it might be good to provide separate `new` and `new_vartime` constructors like `BoxedResidueParams`, so they can accordingly call constant-time versus variable-time remainder

src/modular/residue.rs
```diff
@@ -39,6 +39,8 @@ pub trait ResidueParams<const LIMBS: usize>:
 
     /// The constant modulus
     const MODULUS: Uint<LIMBS>;
+    /// The constant modulus pre-wrapped in NonZero
+    const MODULUS_NZ: NonZero<Uint<LIMBS>>;
```
tarcieri, line 43: Perhaps the modulus should just be `NonZero` to begin with?

fjarri: Good idea. Fixed. Also note that I removed `Residue::new_checked()` and the check in `Residue::new()` because `ResidueParams` is already guaranteed to have an odd modulus.

fjarri: Also had to add a `const fn as_ref()` to `NonZero` to use in macros

tarcieri (approved):

PR #420, opened 2023-12-13
Restore preconditions check in montgomery_mul

newpavlov:
Partially reverts changes from #409


tarcieri:
As I noted in #409, this seems to have no observable effect on the benchmarks, but I guess we can merge anyway.

newpavlov:
Yes, it's expected since bound checks can be easily predicted, but explicitly checked pre-conditions result in a cleaner binary and may even reduce time variability of the code (on the first run execution may trigger several branch misses).

PR #423, opened 2023-12-14
BoxedUint: optimize `shl_assign`

tarcieri:
Allows `shl_assign` to operate in-place on `self` (although it still requires a temporary buffer)

### DIFF
--- a/src/uint/boxed/shl.rs
+++ b/src/uint/boxed/shl.rs
@@ -10,27 +10,37 @@ impl BoxedUint {
     /// Returns a zero and a truthy `Choice` if `shift >= self.bits_precision()`,
     /// or the result and a falsy `Choice` otherwise.
     pub fn shl(&self, shift: u32) -> (Self, Choice) {
+        let mut result = self.clone();
+        let overflow = result.shl_assign_with_carry(shift);
+        (result, overflow)
+    }
+
+    /// Computes `self <<= shift`.
+    ///
+    /// Returns a zero and a truthy `Choice` if `shift >= self.bits_precision()`,
+    /// or the result and a falsy `Choice` otherwise.
+    fn shl_assign_with_carry(&mut self, shift: u32) -> Choice {
         // `floor(log2(bits_precision - 1))` is the number of bits in the representation of `shift`
         // (which lies in range `0 <= shift < bits_precision`).
         let shift_bits = u32::BITS - (self.bits_precision() - 1).leading_zeros();
         let overflow = !shift.ct_lt(&self.bits_precision());
         let shift = shift % self.bits_precision();
-        let mut result = self.clone();
         let mut temp = self.clone();
 
         for i in 0..shift_bits {
             let bit = Choice::from(((shift >> i) & 1) as u8);
             temp.set_zero();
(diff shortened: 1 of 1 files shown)


src/uint/boxed/shl.rs
```diff
@@ -10,27 +10,37 @@ impl BoxedUint {
     /// Returns a zero and a truthy `Choice` if `shift >= self.bits_precision()`,
     /// or the result and a falsy `Choice` otherwise.
     pub fn shl(&self, shift: u32) -> (Self, Choice) {
```
tarcieri, line 12: Perhaps this should be `shl_with_carry` so as to not have a different type signature from `Shl::shl`?

src/uint/boxed/shl.rs
```diff
@@ -10,27 +10,37 @@ impl BoxedUint {
...
+        let mut result = self.clone();
+        let overflow = result.shl_assign_with_carry(shift);
+        (result, overflow)
+    }
+
+    /// Computes `self <<= shift`.
+    ///
+    /// Returns a zero and a truthy `Choice` if `shift >= self.bits_precision()`,
+    /// or the result and a falsy `Choice` otherwise.
+    fn shl_assign_with_carry(&mut self, shift: u32) -> Choice {
```
fjarri, line 22: I don't think that "with_carry" is appropriate here, it doesn't return the bits that were shifted out of bounds. The returned `Choice` indicates the `shift` overflow.

tarcieri: `with_overflow` then? Or other suggestions?

fjarri: Why not just `shl_assign`?

tarcieri: @fjarri similar concern to [this comment](https://github.com/RustCrypto/crypto-bigint/pull/423#discussion_r1426092711) - I think it's confusing to have both an inherent method and a trait method with the same name but different type signatures. Where they do overlap I think they should have the same shape

fjarri: Then probably `overflowing_shl()` would be best. Although to match the method with the same name of the primitive types we would need to do `shift = shift % bits_precision`

tarcieri: It already does: [L17 below](https://github.com/RustCrypto/crypto-bigint/pull/423/files#diff-f78b90032333fb7f636676d68f9ad9dc0a504a3d44c31561540d0c1dbd054443L17)

tarcieri: Also good call on `overflowing_shl*`, renamed in [fee6c3f](https://github.com/RustCrypto/crypto-bigint/pull/423/commits/fee6c3f4a041202dc956ec198f16d9a26a43cbe7).

I could also rename `shl` above to `overflowing_shl` to match (since `shl` would still be available via the `Shl` trait).

tarcieri: (I'd probably prefer to do that in a followup PR though to keep this one focused)

tarcieri:
Gonna merge this and open a separate PR for `overflowing_shl`

PR #425, opened 2023-12-14
Impl `Wrapping*` traits from `num-traits`

tarcieri:
Until now we have used traits defined in this crate, primarily because many of these traits are predicates that return `bool` and in constant-time code we want those to return `Choice`/`CtOption`.

`Wrapping*` is an example of the sort of trait we'd want to incorporate unchanged, and there are potentially others, so this adds `num-traits` as a hard dependency (previously our only hard dependency was `subtle`). Note that `num-traits` itself has no transitive dependencies (or rather, they're optional but not enabled).

The `Wrapping*` traits have bounds on the corresponding op traits being impl'd with `Self` operands e.g. `WrappingAdd: Add<Self, Output = Self>` so this PR also adds impls of those traits.

We've previously avoided these as in `std` they panic on overflow/underflow in debug builds and silently wrap in release builds. This PR always panics.

This required some changes to the `Mul` impls which were conditional on `ConcatMixed` and implicitly widened. To accomodate impls which are always available and require no bounds (in order to allow us to impl `WideningMul`), and renames the following:

- `Uint::mul` -> `Uint::widening_mul`
- `Uint::mul_wide` -> `Uint::widening_mul_split`

TODO:

- [x] `WrappingNeg`
- [x] `WrappingShl`
- [x] `WrappingShr`

cc @xuganyu96

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ subtle = { version = "2.5", default-features = false }
 # optional dependencies
 der = { version = "0.7", optional = true, default-features = false }
 generic-array = { version = "0.14", optional = true }
+num-traits = { version = "0.2", default-features = false }
 rand_core = { version = "0.6.4", optional = true }
 rlp = { version = "0.5", optional = true, default-features = false }
 serdect = { version = "0.2", optional = true, default-features = false }
@@ -33,7 +34,6 @@ criterion = { version = "0.5", features = ["html_reports"] }
 hex-literal = "0.4"
 num-bigint = { package = "num-bigint-dig", version = "0.8" }
 num-integer = "0.1"
-num-traits = "0.2"
 proptest = "1"
 rand_core = { version = "0.6", features = ["std"] }
 rand_chacha = "0.3"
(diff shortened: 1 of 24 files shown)


tarcieri:
Until now we've avoided these sort of panicking impls, requiring the selection of either `Wrapping` or `Checked` semantics in order to ensure that overridden operators don't come with hidden panics.

Other PRs have added panics that brought the behavior more in-line with `core`/`std` semantics (#395) so I thought I'd float this approach.

I'm not entirely sure I like it, to be honest, but it can be annoying not to have a convenient way to perform an operation you're sure won't overflow on e.g. a `BoxedUint` without first having to place it in a `Wrapping` wrapper.

tarcieri:
The main argument in favor of this PR is probably UX/approachability: we've had several complaints this library is hard-to-use, and the absence of the `core::ops` traits on `Uint<_>` are a big part of that, IMO.

Adding them should make this library easier to learn because it will align the library with the `core`/`std` mental model of how things should operate, while also ensuring overflows don't silently happen in release builds.

PR #428, opened 2023-12-14
Boxed uintlike coverage

xuganyu96:
This is an attempt to implement for `BoxedUint` the missing arithmetic needed for `crypto-primes` (see https://github.com/entropyxyz/crypto-primes/pull/37#issuecomment-1856331110).

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -7,8 +7,8 @@ pub use num_traits::{
 use crate::{Limb, NonZero};
 use core::fmt::Debug;
 use core::ops::{
-    BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Div, DivAssign, Not, Rem, Shl,
-    ShlAssign, Shr, ShrAssign,
+    Add, BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Div, DivAssign, Mul, Not,
+    Rem, Shl, ShlAssign, Shr, ShrAssign, Sub,
 };
 use subtle::{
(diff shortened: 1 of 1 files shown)


src/modular/boxed_residue.rs
```diff
@@ -21,6 +21,18 @@ use std::sync::Arc;
...
+fn div_by_2(a: &BoxedUint, modulus: &BoxedUint) -> BoxedUint {
+    let (half, is_odd) = a.shr1_with_carry();
+    let half_modulus = modulus.shr1();
+
+    let if_even = half.clone();
+    let if_odd = half
+        .wrapping_add(&half_modulus)
+        .wrapping_add(&BoxedUint::one_with_precision(a.bits_precision()));
+
+    BoxedUint::conditional_select(&if_even, &if_odd, is_odd)
```
tarcieri, line 33: You can cut down on a number of allocations here:
```suggestion
    let (mut ret, is_odd) = a.shr1_with_carry();
    let half_modulus = modulus.shr1();

    let if_odd = ret
        .wrapping_add(&half_modulus)
        .wrapping_add(&BoxedUint::one());

    ret.conditional_assign(&if_odd, is_odd);
    ret
```

Cargo.toml
```diff
@@ -39,7 +39,7 @@ rand_core = { version = "0.6", features = ["std"] }
 rand_chacha = "0.3"
 
 [features]
-default = ["rand"]
+default = ["rand", "alloc"]
```
tarcieri, line 42: Most of the downstream crates which use this crate don't need the `alloc` feature. It's really just for `rsa`/`dsa`.

PR #430, opened 2023-12-14
Align with `core`/`std` on `overflowing_sh*`

tarcieri:
Changes all shift functions which return an overflow flag (as a `Choice` or `ConstChoice`) to use the `overflowing_sh*` name prefix, which aligns with similar APIs in `core`/`std`.

In their place, adds new `Uint::{shl, shr}` functions which provide the trait-like behavior (i.e. panic on overflow) but work in `const fn` contexts (and can now panic at compile time on overflow).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,7 +45,7 @@
 //!     U256::from_be_hex("ffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551");
 //!
 //! // Compute `MODULUS` shifted right by 1 at compile time
-//! pub const MODULUS_SHR1: U256 = MODULUS.shr(1).0;
+//! pub const MODULUS_SHR1: U256 = MODULUS.shr(1);
 //! ```
 //!
 //! Note that large constant computations may accidentally trigger a the `const_eval_limit` of the compiler.
--- a/src/uint/shl.rs
+++ b/src/uint/shl.rs
@@ -5,9 +5,22 @@ use core::ops::{Shl, ShlAssign};
 
 impl<const LIMBS: usize> Uint<LIMBS> {
     /// Computes `self << shift`.
+    ///
+    /// Panics if `shift >= Self::BITS`.
+    pub const fn shl(&self, shift: u32) -> Self {
+        let (result, overflow) = self.overflowing_shl(shift);
+        assert!(
+            !overflow.is_true_vartime(),
+            "attempt to shift left with overflow"
+        );
+        result
+    }
+
+    /// Computes `self << shift`.
+    ///
     /// If `shift >= Self::BITS`, returns zero as the first tuple element,
     /// and `ConstChoice::TRUE` as the second element.
-    pub const fn shl(&self, shift: u32) -> (Self, ConstChoice) {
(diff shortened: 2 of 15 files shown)


src/lib.rs
```diff
@@ -45,7 +45,7 @@
 //!     U256::from_be_hex("ffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551");
 //!
 //! // Compute `MODULUS` shifted right by 1 at compile time
-//! pub const MODULUS_SHR1: U256 = MODULUS.shr(1).0;
+//! pub const MODULUS_SHR1: U256 = MODULUS.shr(1);
```
tarcieri, line 48: This feels like an ergonomics improvement

src/uint/shl.rs
```diff
@@ -118,10 +133,15 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+    pub(crate) const fn shl1(&self) -> Self {
+        self.overflowing_shl1().0
+    }
+
     /// Computes `self << 1` in constant-time, returning [`ConstChoice::TRUE`]
-    /// if the most significant bit was set, and [`ConstChoice::FALSE`] otherwise.
+    /// if the most significant bit was set (i.e. if overflow occurred), and [`ConstChoice::FALSE`] otherwise.
     #[inline(always)]
-    pub(crate) const fn shl1_with_carry(&self) -> (Self, ConstChoice) {
+    pub(crate) const fn overflowing_shl1(&self) -> (Self, ConstChoice) {
```
tarcieri, line 144: These should arguably remain named `shl1_with_carry` as it's carry, not overflow

tarcieri:
There are a few API gaps which should probably be filled before merging if we're happy this, like every `Uint::overflowing_sh*` method should have a corresponding `Uint::sh*` method which is `const fn`

src/uint/shl.rs
fjarri: Yes, I agree.

fjarri: I mean, they should keep the `carry` suffix.

src/uint/shl.rs
```diff
@@ -5,9 +5,22 @@ use core::ops::{Shl, ShlAssign};
 
 impl<const LIMBS: usize> Uint<LIMBS> {
     /// Computes `self << shift`.
+    ///
+    /// Panics if `shift >= Self::BITS`.
+    pub const fn shl(&self, shift: u32) -> Self {
```
fjarri, line 10: Do we even need these methods? Primitive types don't have them.

tarcieri: It's what makes [this](https://github.com/RustCrypto/crypto-bigint/pull/430#discussion_r1427312479) possible, i.e. using them in a `const fn` context.

Primitive types are a bit magical in that arithmetic operators like `<<` and `>>` work in const contexts even though for all other types they'd thunk through the `Shl`/`Shr` traits which won't be able to do that until `const impl` is available.

fjarri:
LGTM, except that I think `_with_carry()` methods should keep the names.

tarcieri:
Will fix. Thanks!

Issue #351, opened 2023-11-29
Reopening `*Residue` naming bikeshed


tarcieri:
In #70 we discussed `FieldElement` and `Residue` as names for these types.

However @fjarri has pointed out, and I agree, that what these types really represent is Montgomery form. It's not just an implementation detail: we have `*_montgomery` methods which provide conversions and direct access to the Montgomery form value.

I think it would probably make sense to rename them to something with `*Mont` in the name, although I could go a few different directions on specific names.

cc @haslersn

fjarri:
Three independent proposals:
1) Regardless of other choices, drop the `Dyn` prefix for runtime residues and instead use the `Const` prefix for constant-time ones. This goes in line with the rest of the naming.
2) Rename "Residue" to "MontgomeryForm" (or "MgForm" for short). Montgomery form is not a mere implementation detail, so it seems logical that it's reflected in the name.
3) Rename "Params" to "Precomputed" or something of the kind.

dignifiedquire:
In favor for all three suggestions from @fjarri, especially 2 and 3 are more in line with what I have seen in other contexts.

tarcieri:
`MontgomeryForm`/`MontForm`/`MontyForm`/`MgForm` or the same thing with `Montgomery/Mont/etc/MgRepr` (i.e. representative) all seem like reasonable names to me (though elsewhere we and `ff`/`group` use `Repr` for "bytestring representation")

 I would probably side on a shorter name, since e.g. `ConstMontgomeryForm`/`BoxedMontgomeryForm` are quite verbose.

I like the idea of dropping `Dyn` and adding `Const`.

> Rename "Params" to "Precomputed" or something of the kind.

I think "params" is fine, and can use whatever "monty" prefix we decide, e.g. `MontParams`. `BoxedMontPrecomputed` would get a little verbose.

dignifiedquire:
- slight favor of `MontyForm` over the other variants
- openssl uses `Ct` instead of `Const`, so could also use that for naming if we want shorter names

fjarri:
"Ct" is a little confusing because it can also mean "constant time". Also in the other parts of the library we're using "const" prefixes for compile-time stuff, so if we use "ct" here, those should all be changed too.

"Precomputed" can be shortened to "Precomp", but I am not sure how it sounds for English speakers :)

dignifiedquire:
[...]

Issue #348, opened 2023-11-29
Tracking issue for `BoxedUint`/`BoxedResidue` missing features


tarcieri:
I recently spiked out both of these types, but there are likely to be plenty of small gaps feature-wise.

This is a tracking issue for those gaps:

- [x] [Impl `RandomMod` for `BoxedUint`](https://github.com/RustCrypto/crypto-bigint/pull/349)
- [ ] Trait for abstracting over `*Residue` types

cc @dignifiedquire @fjarri

dignifiedquire:
- Comparisons between `NonZero<BoxedUint>` and `BoxedUint`
- `impl Hash for BoxedUint`

dignifiedquire:
- [ ] Impl Hash for `NonZero<BoxedUint>`

tarcieri:
Aside from the `*Residue`-like trait, which we can discuss in #431/#448, the `Boxed*` types are pretty feature-complete at this point. Closing.

PR #431, opened 2023-12-15
Add traits to connect integers and residue represenations

### LINKED ISSUE #448: Trait for `*Residue` / Montgomery form
See also: #348, #431

fjarri:
A part of #348 
Fixes #448

All the naming is up for debate (see also #351)

- add `Monty` associated type to `Integer` linking it to the corresponding Montgomery form
- add `Monty` trait with an associated type linking back to the corresponding `Integer` and a bunch of bounds (similar to `Integer`)
- remove the bound on `Square` which made it hard to make `Square` a bound to other traits. This means that there's no default implementation anymore.
- add `SquareAssign` trait.

I will make a separate PR adding `div_by_2()` to `Monty` after #436 is merged

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -141,6 +141,10 @@ pub trait Integer:
     + WrappingShr
     + Zero
 {
+    /// The corresponding Montgomery representation,
+    /// optimized for the performance of modular operations at the price of a conversion overhead.
+    type MontyForm: MontyFormLike<Integer = Self>;
+
     /// The value `1`.
     fn one() -> Self;
 
@@ -514,16 +518,10 @@ pub trait WideningMul<Rhs = Self>: Sized {
     fn widening_mul(&self, rhs: Rhs) -> Self::Output;
 }
 
-/// This integer has a representation optimized for the performance of modular operations.
-pub trait HasMontyForm {
-    /// The representation type.
-    type MontyForm: MontyFormLike<Raw = Self>;
-}
-
 /// A representation of an integer optimized for the performance of modular operations.
 pub trait MontyFormLike {
     /// The original integer type.
-    type Raw: HasMontyForm<MontyForm = Self>;
+    type Integer: Integer<MontyForm = Self>;
 
     /// The precomputed data needed for this representation.
     type Params: Clone;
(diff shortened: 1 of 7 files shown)


tarcieri:
Elsewhere `*Repr` is used for serialized bytestrings.

It might be good to consider the naming for these traits in conjunction with #351

tarcieri:
Instead of `HasResidueRepr` I think we can just add an associated type to `Integer`.

fjarri:
I wasn't sure if bundling everything into one trait is a good idea. But I don't insist on it.

tarcieri:
Well, if we do end up putting it in its own trait, it seems like a good trait for `Integer` to bound on, anyway

src/traits.rs
```diff
@@ -513,3 +517,27 @@ pub trait WideningMul<Rhs = Self>: Sized {
     /// Perform widening multiplication.
     fn widening_mul(&self, rhs: Rhs) -> Self::Output;
 }
+
+/// A representation of an integer optimized for the performance of modular operations.
+pub trait MontyFormLike {
```
tarcieri, line 522: Maybe just call this `Monty`? I think `*Like` is a little redundant in a trait name.

fjarri: `MontyForm` is already a struct name.

fjarri:
Can't make it a draft, but I'm still working on it: adding trait bounds to the Montgomery trait to make it more like `Integer`. 

One problem not sure how to solve: the current trait bound on `Square` makes it difficult to use it as a bound. Maybe just remove it? It's only needed for the default implementation which is trivial.

src/traits.rs
tarcieri: `Monty` isn't a struct name, however

tarcieri:
Sure, you can remove `Square` and just add it as a trait method elsewhere

fjarri:
No, I meant remove the bound but keep the trait. That should be enough.

src/traits.rs
fjarri: Ah I see. That seems a little vague to me, but I can live with it.

fjarri: Renamed

fjarri:
Ok this should be ready for review now.

tarcieri:
Will look if you want to remove `[WIP]`

src/traits.rs
```diff
@@ -513,3 +519,51 @@ pub trait WideningMul<Rhs = Self>: Sized {
...
+    type Integer: Integer<Monty = Self>;
+
+    /// The precomputed data needed for this representation.
+    type Params: Clone;
+
+    /// Create the precomputed data.
+    ///
+    /// Can return `None` if `modulus` is not valid for the representation;
+    /// see the documentation of the specific type for the requirements.
+    fn new_params(modulus: Self::Integer) -> CtOption<Self::Params>;
```
tarcieri, line 559: Maybe `params_for_modulus`?

tarcieri (approved):
One suggestion but this looks fine

src/traits.rs
fjarri: I don't know, feels a little redundant? Fixed the docstring though since it's explicitly Montgomery representation.

PR #436, opened 2023-12-15
Boxed uintlike coverage

xuganyu96:
This is an attempt to implement for BoxedUint the missing arithmetic needed for crypto-primes (see https://github.com/entropyxyz/crypto-primes/pull/37#issuecomment-1856331110).

This is the continuation of #428.


xuganyu96:
@tarcieri @fjarri this PR it ready for review. Thank you!

src/uint/boxed.rs
```diff
@@ -248,6 +261,13 @@ impl NonZero<BoxedUint> {
     pub fn widen(&self, bits_precision: u32) -> Self {
         NonZero(self.0.widen(bits_precision))
     }
+
+    /// Creates a new non-zero integer in a const context.
+    /// The second return value is `FALSE` if `n` is zero, `TRUE` otherwise.
+    pub const fn const_new(n: BoxedUint) -> (Self, ConstChoice) {
+        let is_nonzero = n.is_nonzero();
+        (Self(n), is_nonzero)
+    }
```
tarcieri, line 270: `BoxedUint` can't be instantiated in a `const fn` context because it's heap-allocated

src/uint/boxed/cmp.rs
```diff
@@ -10,6 +10,52 @@ use subtle::{
...
+        let mut i = 0;
+        while i < limbs.len() {
+            limbs[i] = Limb::select(a.limbs[i], b.limbs[i], c);
+            i += 1;
+        }
+
+        Self {
+            limbs: limbs.into(),
+        }
+    }
```
tarcieri, line 29: This is already impl'd as [`BoxedUint::conditional_select`](https://docs.rs/crypto-bigint/0.6.0-pre.2/crypto_bigint/struct.BoxedUint.html#method.conditional_select). You can use `Into` to convert from `ConstChoice` to `Choice`.

src/uint/boxed/cmp.rs
```diff
@@ -10,6 +10,52 @@ use subtle::{
...
+
+    /// Returns the truthy value if `self >= rhs` and the falsy value otherwise.
+    #[inline]
+    pub(crate) fn gt(lhs: &Self, rhs: &Self) -> ConstChoice {
+        let (_res, borrow) = rhs.sbb(lhs, Limb::ZERO);
+        ConstChoice::from_word_mask(borrow.0)
+    }
```
tarcieri, line 36: This is already impl'd as [`BoxedUint::ct_gt`](https://docs.rs/crypto-bigint/0.6.0-pre.2/crypto_bigint/struct.BoxedUint.html#method.ct_gt). Really there's not a lot of point in supporting `ConstChoice` with `BoxedUint`, and the least-common-denominator trait API should use `Choice` instead (which provides an optimization barrier in non-`const fn` contexts, which is the only place `BoxedUint` can be used)

src/uint/boxed/sqrt.rs
```diff
@@ -0,0 +1,295 @@
+//! [`BoxedUint`] square root operations.
```
tarcieri, line 1: This is fine to get this implemented, but in a followup PR (which I'm happy to do) it should probably get a similar macro treatment to the multiply functions so `Uint` and `BoxedUint` can share a common implementation.

src/uint/boxed.rs
xuganyu96: When should a function return `(Value, Choice)` instead of `CtChoice(Value)`?

In `boxed/sqrt.rs` the implementation of `sqrt` uses a `NonZero::<BoxedUint>::new(x)` whose value needs to be used whether `x` is zero or not, but `BoxedUint` is not `ConditionallySelectable` so I can't use `CtChoice`'s `map` and `unwrap_or`. It seems like having a `NonZero::<BoxedUint>::new` that returns `(NonZero<BoxedUint>, Choice)` would be necessary, but the current `new`'s signature doesn't work this way, and such a `new` is definitely not `const_new`.

tarcieri: The (hopefully temporary) workaround for that sort of thing is [`BoxedUint::conditional_map`](https://docs.rs/crypto-bigint/0.6.0-pre.3/crypto_bigint/struct.BoxedUint.html#method.conditional_map).

I'm [trying to find an upstream solution in `subtle`](https://github.com/dalek-cryptography/subtle/pull/118) but it may be necessary to define traits shaped like the ones in that PR in `crypto-bigint` as a stopgap (and a `CtOption`-alike type, possibly `ConstOption`) /cc @fjarri

xuganyu96:
Note to myself: use `conditional_map` in `BoxedUint::sqrt`

src/modular/boxed_residue.rs
```diff
@@ -233,6 +234,18 @@ impl BoxedResidue {
...
     }
+
+    /// Performs the modular division by 2, that is for given `x` returns `y`
+    /// such that `y * 2 = x mod p`. This means:
+    /// - if `x` is even, returns `x / 2`,
+    /// - if `x` is odd, returns `(x + p) / 2`
+    ///   (since the modulus `p` in Montgomery form is always odd, this divides entirely).
+    pub fn div_by_2(&self) -> Self {
+        Self {
+            montgomery_form: div_by_2(&self.montgomery_form, &self.residue_params.modulus),
```
tarcieri, line 245: This is a little confusing due to the name clash which makes it almost look like it's going to recurse, although it's hard to suggest something better

xuganyu96: What about using fully qualified namespace like `div_by_2::boxed::div_by_2`?

tarcieri: Sure, sounds good

src/uint/boxed.rs
xuganyu96: I saw the `ConstantTimeSelect` trait and will use `ct_select` and `ct_assign` where appropriate. Thank you for the implementation!

xuganyu96:
Newest commits:
- Used `ConstantTimeSelect::{ct_assign, ct_select}` where appropriate (see `boxed/sqrt.rs` and `div_by_2`)
- Added `ConstCtOption<NonZero<Limb>>::expect` (for use in `crypto-primes`)

src/uint/boxed.rs
```diff
@@ -92,6 +94,17 @@ impl BoxedUint {
...
 
+    /// Returns the truthy value if `self`!=0 or the falsy value otherwise.
+    pub(crate) fn is_nonzero(&self) -> Choice {
+        let mut b = 0;
+        let mut i = 0;
+        while i < self.limbs.len() {
+            b |= self.limbs[i].0;
+            i += 1;
+        }
+        Limb(b).is_nonzero().into()
```
tarcieri, line 105:
```suggestion
        let choice = Choice::from(1u8);
        
        for limb in &self.limbs {
            choice &= limb.ct_eq(&Limb::ZERO);
        }
        
        choice
```

src/uint/boxed.rs
```diff
@@ -245,6 +258,13 @@ impl BoxedUint {
 }
 
 impl NonZero<BoxedUint> {
+    /// TODO: this is not really "const", but I need a way to return (value, choice) since
+    /// BoxedUint is not [`ConditionallySelectable`] so `CtChoice::map` and such does not work
+    pub fn const_new(n: BoxedUint) -> (Self, Choice) {
+        let nonzero = n.is_nonzero();
+        (Self(n), nonzero)
+    }
```
tarcieri, line 266: I factored these into methods on e.g. in this case `BoxedUint`: https://github.com/RustCrypto/crypto-bigint/pull/484

You can use `ConstCtChoice` now.

src/uint/boxed/encoding.rs
```diff
@@ -131,6 +131,40 @@ impl BoxedUint {
 
         out.into()
     }
+
+    /// Create a new [`BoxedUint`] from the provided big endian hex string.
+    pub fn from_be_hex(hex: &str, bits_precision: u32) -> Self {
```
tarcieri, line 136: Since these can't be constructed at compile-time, it would probably be good to make it fallible

src/uint/boxed/shl.rs
```diff
@@ -111,19 +111,6 @@ impl BoxedUint {
...
-    ///
-    /// NOTE: this operation is variable time with respect to `shift` *ONLY*.
-    ///
-    /// When used with a fixed `shift`, this function is constant-time with respect to `self`.
-    #[inline(always)]
-    pub fn shl_vartime(&self, shift: u32) -> Option<Self> {
-        let mut result = Self::zero_with_precision(self.bits_precision());
-        let success = self.shl_vartime_into(&mut result, shift);
-        success.map(|_| result)
-    }
```
tarcieri, line 125: ?

src/uint/boxed/shr.rs
```diff
@@ -115,19 +115,6 @@ impl BoxedUint {
...
-    ///
-    /// NOTE: this operation is variable time with respect to `shift` *ONLY*.
-    ///
-    /// When used with a fixed `shift`, this function is constant-time with respect to `self`.
-    #[inline(always)]
-    pub fn shr_vartime(&self, shift: u32) -> Option<Self> {
-        let mut result = Self::zero_with_precision(self.bits_precision());
-        let success = self.shr_vartime_into(&mut result, shift);
-        success.map(|_| result)
-    }
```
tarcieri, line 129: ?

src/uint/boxed/shl.rs
xuganyu96: I originally made them to accommodate the requirement from `crypto-primes`, but now I realized I can use `overflowing_shl`/`overflowing_shr` to achieve the same thing, so I felt no need to add `shl_vartime`/`shr_vartime` for `BoxedUint`.

src/uint/boxed/shr.rs
xuganyu96: Same as above.

src/uint/boxed.rs
xuganyu96: How about just `!self.is_zero()`?

tarcieri: I mean, we can do that, but other instances can be replaced with `!x.is_zero()` too. It makes less sense when `const fn` isn't required

src/uint/boxed/shl.rs
tarcieri: Err, I added `shl_vartime` in #330. The reason we have `*_vartime` equivalents of the `sh*` methods is because they can be quite a bit faster.

tarcieri:
@xuganyu96 sorry about the rebase you're gonna have to do after #485, but we'll try to get this merged ASAP once you do!

benches/boxed_uint.rs
```diff
@@ -11,7 +11,7 @@ fn bench_shifts(c: &mut Criterion) {
     group.bench_function("shl_vartime", |b| {
         b.iter_batched(
             || BoxedUint::random(&mut OsRng, UINT_BITS),
-            |x| black_box(x.shl_vartime(UINT_BITS / 2 + 10)),
+            |x| black_box(x.overflowing_shl(UINT_BITS / 2 + 10).0),
```
fjarri, line 14: Why the change here? And if you're benchmarking `overflowing_shl` now, the string label should be changed too.

src/modular/boxed_monty_form.rs
```diff
@@ -242,6 +243,18 @@ impl BoxedMontyForm {
...
+
+    /// Performs the modular division by 2, that is for given `x` returns `y`
+    /// such that `y * 2 = x mod p`. This means:
+    /// - if `x` is even, returns `x / 2`,
+    /// - if `x` is odd, returns `(x + p) / 2`
+    ///   (since the modulus `p` in Montgomery form is always odd, this divides entirely).
+    pub fn div_by_2(&self) -> Self {
+        Self {
+            montgomery_form: div_by_2::boxed::div_by_2(&self.montgomery_form, &self.params.modulus),
+            params: self.params.clone(), // TODO: avoid clone?
```
fjarri, line 255: There would be no need for clone if it was an inplace method, but if it creates a new number, I think cloning is unavoidable.

tarcieri: When `std` is available they're stored as an `Arc`, so cloning is cheap (incrementing an atomic reference counter)

src/modular/div_by_2.rs
```diff
@@ -28,3 +28,23 @@ pub(crate) fn div_by_2<const LIMBS: usize>(a: &Uint<LIMBS>, modulus: &Uint<LIMBS
 
     Uint::<LIMBS>::select(&if_even, &if_odd, is_odd)
 }
+
+#[cfg(feature = "alloc")]
+pub(crate) mod boxed {
```
fjarri, line 33: I think putting different variants of this function in modules is a bit of an overcomplication, perhaps just different suffixes would suffice.

src/uint/boxed/bits.rs
```diff
@@ -84,7 +140,7 @@ mod tests {
     fn uint_with_bits_at(positions: &[u32]) -> BoxedUint {
         let mut result = BoxedUint::zero_with_precision(256);
         for &pos in positions {
-            result |= BoxedUint::one_with_precision(256).shl_vartime(pos).unwrap();
+            result |= BoxedUint::one_with_precision(256).overflowing_shl(pos).0;
```
fjarri, line 143: Any specific reason for the change here?

src/uint/boxed/cmp.rs
```diff
@@ -10,6 +10,28 @@ use subtle::{
     Choice, ConditionallySelectable, ConstantTimeEq, ConstantTimeGreater, ConstantTimeLess,
 };
 
+impl BoxedUint {
+    /// Returns the Ordering between `self` and `rhs` in variable time.
+    pub fn cmp_vartime(&self, rhs: &Self) -> Ordering {
```
fjarri, line 15: I have a feeling that using comparisons between limbs instead of subtraction may be faster. Put a TODO perhaps?

src/uint/boxed/div.rs
```diff
@@ -38,7 +52,7 @@ impl BoxedUint {
         let mut bd = self.bits_precision() - mb;
         let mut rem = self.clone();
         // Will not overflow since `bd < bits_precision`
-        let mut c = rhs.shl_vartime(bd).expect("shift within range");
+        let mut c = rhs.overflowing_shl(bd).0;
```
fjarri, line 55: It's a vartime method, no need to use a constant-time `overflowing_shl` here

src/uint/boxed/div.rs
```diff
@@ -112,7 +135,7 @@ impl BoxedUint {
         let mut remainder = self.clone();
         let mut quotient = Self::zero_with_precision(self.bits_precision());
         // Will not overflow since `bd < bits_precision`
-        let mut c = rhs.shl_vartime(bd).expect("shift within range");
+        let mut c = rhs.overflowing_shl(bd).0;
```
fjarri, line 138: Same as above, no need for a constant-time shift here.

src/uint/reciprocal.rs
```diff
@@ -0,0 +1,208 @@
+//! Reciprocal, shared across Uint and BoxedUint
```
fjarri, line 1: This should reference the paper the algorithm was taken from.

src/uint/boxed/div_limb.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+    (BoxedUint { limbs: q.into() }, Limb(r >> reciprocal.shift()))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{div2by1, Reciprocal};
+    use crate::{Limb, NonZero, Word};
+    #[test]
+    fn div2by1_overflow() {
```
fjarri, line 32: Since this test tests a function from `uint::reciprocal`, perhaps it should be located there.

src/uint/boxed/shl.rs
fjarri: Oh now I see the reason for all the replaced `_vartime` methods previously. Yes, we should keep those.

src/uint/reciprocal.rs
```diff
@@ -0,0 +1,208 @@
...
+    let x = v3.wrapping_add(1);
+    let (hi, _lo) = primitives::mulhilo(x, d);
+    let hi = ConstChoice::from_word_nonzero(x).select_word(d, hi);
+
+    v3.wrapping_sub(hi).wrapping_sub(d)
+}
+
+/// Returns `u32::MAX` if `a < b` and `0` otherwise.
+#[inline]
+const fn lt(a: u32, b: u32) -> u32 {
```
fjarri, line 70: `lt` and `select` should really be moved to `ConstChoice` - forgot to do that in #458

benches/boxed_uint.rs
xuganyu96: Yeah sorry I think I got really confused between the various `shl`/`shr` functions.

src/uint/boxed/bits.rs
xuganyu96: Another instance of me being confused. Will restore back to its original state.

src/uint/boxed/div_limb.rs
xuganyu96: This test is actually an exact copy of the same test in `uint::div_limb`. I will delete this one in `boxed/div_limb.rs`.

src/uint/reciprocal.rs
xuganyu96: I've added them as TODO's.

xuganyu96:
@tarcieri @fjarri sorry but I felt I lost track of where I made changes that overrode your work (e.g. `shl_vartime` and `shr_vartime`). I've opened another PR #489.

PR #440, opened 2023-12-16
ConstChoice improvements

### LINKED ISSUE #418: Use wide arithmetic in `CtChoice`
We are currently using two different algorithms for constant-time comparison of `Word`s:
- the one in `CtChoice` (e.g. `from_word_lt()`), using arithmetic on `Word`
- the one in `Limb`'s implementations of `subtle` traits, using arithmetic on `WideWord`

The latter seems to be faster than the former (from the `limb.rs` benchmarks). We should use just one version of the algorithms both for performance reasons, and for simplicity. Probably should be implemented in `CtChoice`, and the `Limb`'s impls should call it and convert `CtChoice` to `Choice`.

Also there is a third way to do it, by using the built-in `overflowing_sub()` method instead of `sbb`. Seems to be producing branch-free assembly, according to Godbolt - how safe is it?

### LINKED ISSUE #422: Make a `const` equivalent of `subtle::CtOption`
Since for the time being `subtle` does not work in `const` context (until const traits are stabilized), we cannot use `CtOption` to denote optional results. This leads to returning tuples `(Uint, CtChoice)` which are prone to misuse.

We should implement an analogue of `CtOption` (`ConstCtOption`?) that works in `const fn` contexts.

fjarri:
- A leftover from #417 - rename `ct_choice` module to `const_choice`, to match the struct name
- Fixes #418 (in a way) - now `Limb`'s `subtle` impls piggyback on `ConstChoice`, so that we don't have two different algorithms doing the same thing. I use the Hacker's Delight approach - it is marginally safer, and doesn't seem to lead to performance reduction.
- Fixes #422 - add `ConstOption` and use it across the API.

I thought about adding `NonZero::const_new_unchecked()` and `ConstOption::unwrap_unchecked()` for use in tight loops, but using `expect` doesn't seem to decrease performance.


src/const_choice.rs
```diff
@@ -71,10 +71,19 @@ impl ConstChoice {
     /// Returns the truthy value if `x < y`, and the falsy value otherwise.
     #[inline]
     pub(crate) const fn from_word_lt(x: Word, y: Word) -> Self {
+        // See "Hacker's Delight" 2nd ed, section 2-12 (Comparison predicates)
         let bit = (((!x) & y) | (((!x) | y) & (x.wrapping_sub(y)))) >> (Word::BITS - 1);
```
tarcieri, line 75: This is also duplicated here, FWIW: https://github.com/RustCrypto/crypto-bigint/blob/c529c86/src/modular/boxed_residue/mul.rs#L251

fjarri: It is based on the same algorithm, but it is not exactly the same - it cannot make use of `ConstChoice::from_word_lt()`

tarcieri: Maybe we could provide these two algorithms implemented in a more general form in a common module somewhere? (a new private toplevel `bits` module, perhaps?)

fjarri: I thought that it would be nice to have some kind of a "primitives" module with all the little algorithms for bit magic, operations on `Word`s with carry, and so one, which are currently spread around the crate. So that people knew where to find them and didn't write their own versions.

tarcieri: Sounds good to me

src/const_choice.rs
```diff
@@ -164,11 +174,144 @@ impl PartialEq for ConstChoice {
     }
 }
 
+/// An equivalent of `subtle::CtOption` usable in a `const fn` context.
+#[derive(Debug, Clone)]
+pub struct ConstOption<T> {
```
tarcieri, line 179: Perhaps this should be `ConstCtOption` so the mapping is:

- `subtle::Choice` => `ConstChoice`
- `subtle::CtOption` => `ConstCtOption`

?

fjarri: Changed the name.

> I also find ConstOption a little confusing as Option is already const-friendly

It is actually not, only `is_some()`, `is_none()` and `as_ref()` are `const`.

src/const_choice.rs
fjarri: I suppose that should be a separate PR?

dignifiedquire: yeah that would be helpful with my multiplication PR as well

tarcieri: @fjarri whatever works, just a suggestion

tarcieri (approved):

src/const_choice.rs
tarcieri: Well, pattern matching gets you most of the functionality. The rest of that is just sugar.

src/const_choice.rs
fjarri: Working on the primitives module in #458

PR #443, opened 2023-12-16
BoxedResidue: ensure modpow results are fully reduced

tarcieri:
Modular exponentiation of `BoxedResidue` is performed using "almost Montgomery multiplication" whose output is reduced to the bit size of the modulus, but not the modulus itself. This is fine for the intermediate values in the operation, which consist of a series of squarings/multiplies, but the final result needs to be fully reduced.

This is implemented by conditionally subtracting the modulus in constant-time.

A final debug assertion ensures the result is fully reduced.


tarcieri:
It'd be nice if we had a test case to fully exercise this

PR #454, opened 2023-12-17
Add `ConstantTimeSelect` trait

tarcieri:
We can't impl `subtle::ConditionallySelectable` for `Boxed*` types due to a bound on `Copy`.

I've proposed various ways to try to fix this upstream, e.g. dalek-cryptography/subtle#118, from which the `ConstantTimeSelect` trait has been extracted. It provides the same API as
`ConditionallySelectable` but without the `Copy` bound.

A blanket impl of `ConstantTimeSelect` for all `ConditionallySelectable` types means that all stack allocated types can continue to use `ConditionallySelectable` and will receive an impl of `ConstantTimeSelect` as well.

A bound on `ConstantTimeSelect` has been added to the `Integer` trait as well, allowing it to be used on all `*Uint` types in this crate with `Integer` as the trait abstraction.

### DIFF
--- a/src/modular/boxed_residue.rs
+++ b/src/modular/boxed_residue.rs
@@ -12,7 +12,7 @@ use super::{
     reduction::{montgomery_reduction_boxed, montgomery_reduction_boxed_mut},
     Retrieve,
 };
-use crate::{BoxedUint, Integer, Limb, NonZero, Word};
+use crate::{BoxedUint, ConstantTimeSelect, Integer, Limb, NonZero, Word};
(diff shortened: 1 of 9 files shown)


tarcieri:
This of course doesn't work with the `CtOption` combinators (most of which are arguably better implemented without it), but with #440 introducing a new const-friendly `CtOption`-alike we can potentially work around that.

tarcieri:
Merging this so we have some sort of trait-based abstraction over these types

PR #455, opened 2023-12-17
BoxedResidue: add a second conditional subtraction to `pow`

tarcieri:
Proptest found an edge case (checked in) where the output of `BoxedResidue::pow` was still greater than the modulus.

The original implementation in `num-bigint` this was adapted from conditionally subtracted (not in constant-time, mind you) and then checked again, conditionally applying a full remainder in the event it still exceeded the modulus.

We might need to end up doing something similar, but for now, subtracting the modulus again seems to at least address the proptest failure.


src/modular/boxed_residue/pow.rs
```diff
@@ -113,6 +113,7 @@ fn pow_montgomery_form(
     // Ensure output is fully reduced (AMM only reduces to the bit length of the modulus)
     // See RustCrypto/crypto-bigint#441
     z.conditional_sbb_assign(modulus, !z.ct_lt(modulus));
+    z.conditional_sbb_assign(modulus, !z.ct_lt(modulus));
```
tarcieri, line 116: Alternatively we could always compute the remainder and then select between the original `z` and the remainder based on `z.ct_lt(modulus)`, skipping the conditional subtractions, but that would add allocations and as things currently stand a decent performance hit if we were to use constant-time remainder.

PR #458, opened 2023-12-17
Move various operations on Words to a new `primitves` module

fjarri:
TODO: review function names

### DIFF
--- a/src/primitives.rs
+++ b/src/primitives.rs
@@ -1,0 +1,73 @@
+use crate::{WideWord, Word};
+
+/// Multiplies `x` and `y`, returning the most significant
+/// and the least significant words as `(hi, lo)`.
+#[inline(always)]
+pub(crate) const fn mulhilo(x: Word, y: Word) -> (Word, Word) {
+    let res = (x as WideWord) * (y as WideWord);
+    ((res >> Word::BITS) as Word, res as Word)
+}
+
+/// Adds wide numbers represented by pairs of (most significant word, least significant word)
+/// and returns the result in the same format `(hi, lo)`.
+#[inline(always)]
+pub(crate) const fn addhilo(x_hi: Word, x_lo: Word, y_hi: Word, y_lo: Word) -> (Word, Word) {
+    let res = (((x_hi as WideWord) << Word::BITS) | (x_lo as WideWord))
+        + (((y_hi as WideWord) << Word::BITS) | (y_lo as WideWord));
+    ((res >> Word::BITS) as Word, res as Word)
+}
+
+/// Computes `lhs + rhs + carry`, returning the result along with the new carry (0, 1, or 2).
+#[inline(always)]
+pub const fn adc(lhs: Word, rhs: Word, carry: Word) -> (Word, Word) {
+    // We could use `Word::overflowing_add()` here analogous to `overflowing_add()`,
+    // but this version seems to produce a slightly better assembly.
+    let a = lhs as WideWord;
+    let b = rhs as WideWord;
+    let carry = carry as WideWord;
+    let ret = a + b + carry;
+    (ret as Word, (ret >> Word::BITS) as Word)
+}
+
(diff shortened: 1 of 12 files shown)


src/primitives.rs
```diff
@@ -0,0 +1,72 @@
...
+pub(crate) const fn addhilo(x_hi: Word, x_lo: Word, y_hi: Word, y_lo: Word) -> (Word, Word) {
+    let res = (((x_hi as WideWord) << Word::BITS) | (x_lo as WideWord))
+        + (((y_hi as WideWord) << Word::BITS) | (y_lo as WideWord));
+    ((res >> Word::BITS) as Word, res as Word)
+}
+
+/// Computes `lhs + rhs + carry`, returning the result along with the new carry (0, 1, or 2).
+#[inline(always)]
+pub const fn adc3(lhs: Word, rhs: Word, carry: Word) -> (Word, Word) {
+    let a = lhs as WideWord;
```
fjarri, line 23: This can also be expressed via `overflowing_add()`

src/primitives.rs
```diff
@@ -0,0 +1,72 @@
...
+    let b = rhs as WideWord;
+    let carry = carry as WideWord;
+    let ret = a + b + carry;
+    (ret as Word, (ret >> Word::BITS) as Word)
+}
+
+/// Computes `lhs + rhs`, returning the result along with the carry (0 or 1).
+#[inline(always)]
+pub const fn adc(lhs: Word, rhs: Word) -> (Word, Word) {
+    let (res, carry) = lhs.overflowing_add(rhs);
```
fjarri, line 33: This should be constant-time according to Godbolt. Safe to use?

src/primitives.rs
fjarri: In general, we would want to have both `adc` and `adc3` use the same method.

tarcieri:
I guess this is draft because it's WIP? Otherwise looks great so far

src/primitives.rs
tarcieri: The main thing that seems weird is that it doesn't accept a carry as input, so you can't actually use it to build a carry chain?

tarcieri: I guess that's what you're calling `adc3` and I suppose they're both valuable

fjarri: Yes, judging by the benchmarks the compiler doesn't optimize out `carry = 0`.

fjarri:
Adjusted the names and added some comments.

src/primitives.rs
tarcieri: Aah cool, I like this being renamed to `overflowing_add`, but in that case perhaps we can just use the method on the primitive type?

fjarri: I wanted it to imply that using `Word::overflowing_add()` was "vetted" as opposed to just using it directly which may seem like an omission. Also make it easier to try different approaches e.g. via `WideWord`.

tarcieri (approved):
This seems good unless you wanted to add the Hacker's Delight overflow checks and such

fjarri:
If they're used elsewhere other than `const_choise.rs` we can add them, but for now I think this is good enough.

tarcieri:
There is duplication between `const_choice.rs` and `modular/boxed_residue/mul.rs`

fjarri:
Not anymore.

tarcieri:
It seems like these two duplicate each other:

https://github.com/RustCrypto/crypto-bigint/blob/00e1a60/src/const_choice.rs#L71-L85

And it's also duplicated here:

https://github.com/RustCrypto/crypto-bigint/blob/00e1a60/src/modular/boxed_residue/mul.rs#L305

Potentially all three could call a common function?

fjarri:
The one in `boxed_residue/mul` has been removed in this PR, replaced with `sbb()`. The one in `const_choice` can indeed be expressed through the existing method, fixed. The `u32`/`Word` repetitions can be removed with a macro, but I'm not sure if it's worth it.

PR #461, opened 2023-12-18
`BoxedResidueInverter` tests and benchmarks

tarcieri:
Even without any optimization work, it seems significantly faster:

```
Montgomery arithmetic/invert, U256
                        time:   [6.8324 ms 6.8402 ms 6.8482 ms]
```

```
Montgomery arithmetic/Bernstein-Yang invert, U256
                        time:   [342.58 µs 343.71 µs 344.81 µs]
```

### DIFF
--- a/benches/boxed_residue.rs
+++ b/benches/boxed_residue.rs
@@ -4,7 +4,7 @@ use criterion::{
 };
 use crypto_bigint::{
     modular::{BoxedResidue, BoxedResidueParams},
-    BoxedUint,
+    BoxedUint, Inverter, NonZero, PrecomputeInverter, RandomMod,
 };
 use num_bigint::BigUint;
 use rand_core::OsRng;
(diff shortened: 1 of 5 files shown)


benches/boxed_residue.rs
```diff
@@ -21,6 +21,32 @@ fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
...
+            },
+            |x| black_box(x).invert(),
+            BatchSize::SmallInput,
+        )
+    });
+
+    group.bench_function("Bernstein-Yang invert, U256", |b| {
+        b.iter_batched(
+            || {
+                let inverter = params.precompute_inverter();
```
dignifiedquire, line 39: I think for a fair comparison, this needs to be part of the benchmark loop

tarcieri: Well, the idea was to compare the performance once the one-time precomputation penalty had been paid, although it isn't that much, and I'd like to play with replacing the existing implementation of modular inversions for `BoxedResidue` to use Bernstein-Yang for odd moduli, which would benchmark everything.

PR #487, opened 2023-12-20
Add `Odd` newtype

### LINKED ISSUE #479: `Odd` wrapper type?
Similar to `NonZero`, it seems like we have a lot of places that could benefit from knowing a value is odd with type safety, because such type safety would allow for infallible APIs.

For example, the `*Residue` types ensure an odd modulus, which we store. If that modulus were ensured `Odd` at the type level, we could infallibly instantiate a Bernstein-Yang inverter for the same modulus. Right now this requires an `expect`.

If we make both `NonZero` and `Odd` into `#[repr(transparent)]`, with a bit of unsafety we could `impl AsRef<&NonZero<T>> for Odd<T>` as well, allowing any `&Odd<T>` value to be used where `&NonZero<T>` is accepted (and since you can cast in a `const fn` context, we could have a `const fn` method to do that conversion as well)

tarcieri:
Adds a newtype similar to `NonZero` which identifies odd integers at the type level.

This can be used for infallible conversions for moduli, i.e. reusing a `*MontParams` modulus as a Bernstein-Yang modulus infallibly.

Closes #479

### DIFF
--- a/src/modular/const_monty_form/macros.rs
+++ b/src/modular/const_monty_form/macros.rs
@@ -46,7 +46,7 @@ macro_rules! impl_modulus {
 
             const R3: $uint_type = $crate::modular::montgomery_reduction(
                 &Self::R2.square_wide(),
-                Self::MODULUS.as_ref(),
+                &Self::MODULUS,
                 Self::MOD_NEG_INV,
             );
         }
--- a/src/modular/const_monty_form/macros.rs
+++ b/src/modular/const_monty_form/macros.rs
@@ -21,27 +21,20 @@ macro_rules! impl_modulus {
             $uint_type: $crate::ConcatMixed<MixedOutput = $crate::Uint<DLIMBS>>,
         {
             const LIMBS: usize = <$uint_type>::LIMBS;
-            const MODULUS: $crate::NonZero<$uint_type> = {
+            const MODULUS: $crate::Odd<$uint_type> = {
                 let res = <$uint_type>::from_be_hex($value);
-
-                // Check that the modulus is odd
-                if res.as_limbs()[0].0 & 1 == 0 {
-                    panic!("modulus must be odd");
-                }
-
-                // Can unwrap here since `res` was asserted to be odd.
-                res.to_nz().expect("modulus ensured non-zero")
+                res.to_odd().expect("modulus ensured non-zero")
             };
 
             // `R mod MODULUS` where `R = 2^BITS`.
             // Represents 1 in Montgomery form.
             const ONE: $uint_type = $crate::Uint::MAX
-                .rem_vartime(&Self::MODULUS)
+                .rem_vartime(Self::MODULUS.as_nz_ref())
                 .wrapping_add(&$crate::Uint::ONE);
 
             // `R^2 mod MODULUS`, used to convert integers to Montgomery form.
             const R2: $uint_type =
-                $crate::Uint::rem_wide_vartime(Self::ONE.square_wide(), &Self::MODULUS);
+                $crate::Uint::rem_wide_vartime(Self::ONE.square_wide(), Self::MODULUS.as_nz_ref());
 
             const MOD_NEG_INV: $crate::Limb = $crate::Limb(
                 $crate::Word::MIN.wrapping_sub(
--- a/src/modular/const_monty_form/macros.rs
+++ b/src/modular/const_monty_form/macros.rs
@@ -23,7 +23,7 @@ macro_rules! impl_modulus {
             const LIMBS: usize = <$uint_type>::LIMBS;
             const MODULUS: $crate::Odd<$uint_type> = {
                 let res = <$uint_type>::from_be_hex($value);
-                res.to_odd().expect("modulus ensured non-zero")
+                res.to_odd().expect("modulus must be odd")
             };
 
             // `R mod MODULUS` where `R = 2^BITS`.
--- a/src/modular/const_monty_form/macros.rs
+++ b/src/modular/const_monty_form/macros.rs
@@ -21,10 +21,7 @@ macro_rules! impl_modulus {
             $uint_type: $crate::ConcatMixed<MixedOutput = $crate::Uint<DLIMBS>>,
         {
             const LIMBS: usize = <$uint_type>::LIMBS;
-            const MODULUS: $crate::Odd<$uint_type> = {
-                let res = <$uint_type>::from_be_hex($value);
-                res.to_odd().expect("modulus must be odd")
-            };
+            const MODULUS: $crate::Odd<$uint_type> = $crate::Odd::<$uint_type>::from_be_hex($value);
 
             // `R mod MODULUS` where `R = 2^BITS`.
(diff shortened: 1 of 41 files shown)


src/modular/const_monty_form/macros.rs
```diff
@@ -21,27 +21,20 @@ macro_rules! impl_modulus {
...
                 let res = <$uint_type>::from_be_hex($value);
-
-                // Check that the modulus is odd
-                if res.as_limbs()[0].0 & 1 == 0 {
-                    panic!("modulus must be odd");
-                }
-
-                // Can unwrap here since `res` was asserted to be odd.
-                res.to_nz().expect("modulus ensured non-zero")
+                res.to_odd().expect("modulus must be odd")
```
fjarri, line 26: I think the phrasing in "expect" messages is supposed to be in line with the verb, that is `.expect("modulus ensured odd")`. That said, the formatting in `panic!()` calls in `ConstCtOption` may need to be adjusted (e.g. `assert!(self.is_some.is_true_vartime(), "Expected {}", msg);`)

src/modular/monty_form.rs
```diff
@@ -37,33 +37,38 @@ impl<const LIMBS: usize> MontyParams<LIMBS> {
     ///
     /// Returns `None` if the provided modulus is not odd.
     pub fn new(modulus: &Uint<LIMBS>) -> CtOption<Self> {
```
fjarri, line 39: Should we just take `Odd` here thus removing the need to return `CtOption`?

tarcieri: Yeah, seems like a good idea

src/modular/const_monty_form/macros.rs
tarcieri: That's not what the documentation says: https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style

They explicitly suggest phrasing like "slice should not be empty" (although as someone who constantly reads documents with RFC2119 phrasing, I prefer "must" over "should" for cases like this)

fjarri: I personally think it's a bad recommendation, and in that example `.expect("non-empty slice");` reads much better. But if we are to follow it, a lot of `expect()` usages in the crate should be amended.

tarcieri: I think the "ensured" phrasing is fine in cases where code earlier in the function actually ensured something, i.e. internal usages, such as the ones around the existence of modular inverses in these constructors.

For external usages that might hit end-users, I find the kind of wording recommended in `Option::expect` much less confusing.

fjarri: Mostly because the corresponding `panic()` calls in `expect()` implementations expect the recommended phrasing from the docs. That is, they just print the message verbatim, without an "expects: " prefix, so what reads good in the code (with `expect` present) is confusing for the user and vice versa.

tarcieri: I think "modulus must be odd" is quite clear in those kinds of cases, whereas "modulus ensured non-zero" is not

fjarri: It is different though. "modulus must be odd" means that the modulus may not be odd at this point, in which case there's a panic. "modulus ensured odd" means that the modulus *cannot* be odd at this point due to the previous code, and there won't be a panic (unless there's a bug in the code).

tarcieri: Yeah, that's what I just said re: internal usages, highlighting `inv_mod2k_vartime` (though there I used the wording "modular inverse should exist").

But this particular instance is for an external-facing user error, where they have supplied an even modulus to `impl_modulus!` by mistake.

fjarri: But the reason for the user facing/dev facing difference is just the formatting of the message in the implementation of `expect()`. We can change the formatting in `ConstCtOption` thus eliminating that difference.

fjarri: Ah wait, I see what you're saying. This is another thing, `expect` being used as "try to unwrap with a message on failure" or as "assert that this should always be true unless there are bugs". These should really be two different methods, but we have what we have.

tarcieri: Yeah, the message is particularly relevant for the compile error surfaced for an invalid modulus:

```
error[E0080]: evaluation of `<Modulus as ConstMontyParams<{ <$uint_type>::LIMBS }>>::MODULUS` failed
   --> /Users/tarcieri/src/RustCrypto/crypto-bigint/src/const_choice.rs:317:9
    |
317 |         assert!(self.is_some.is_true_vartime(), "{}", msg);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'modulus must be odd', /Users/tarcieri/src/RustCrypto/crypto-bigint/src/const_choice.rs:317:9
```

Perhaps the error would be better as an explicit (const) panic instead of using `expect`, because it's a little bit hard to tell what's going on there.

tarcieri: I removed the `expect` here. It's now just:

```rust
            const MODULUS: $crate::Odd<$uint_type> = $crate::Odd::<$uint_type>::from_be_hex($value);
```

The `from_be_hex` method calls the same named method on `Uint` and does an assert that the returned value is odd or panics: https://github.com/RustCrypto/crypto-bigint/pull/487/files#diff-747dcf65a56c97dcb57a534812fd0eaa55a56a8c3f5da3d0267456deb3a10cd9R49

This makes for a better error message:

```
error[E0080]: evaluation of `<TestMod as ConstMontyParams<{ <$uint_type>::LIMBS }>>::MODULUS` failed
  --> /Users/tarcieri/src/RustCrypto/crypto-bigint/src/odd.rs:49:9
   |
49 |         assert!(uint.is_odd().is_true_vartime(), "number must be odd");
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'number must be odd', /Users/tarcieri/src/RustCrypto/crypto-bigint/src/odd.rs:49:9
```

src/modular/monty_form.rs
tarcieri: Impl'd in 7e511d7 but it looks like it needs a rebase

tests/uint_proptests.rs
```diff
@@ -141,8 +141,8 @@ proptest! {
 
     #[test]
     fn add_mod_nist_p256(a in uint_mod_p(P), b in uint_mod_p(P)) {
-        assert!(a < P);
-        assert!(b < P);
+        assert!(a < *P);
+        assert!(b < *P);
```
tarcieri, line 145: A `PartialCmp<Odd<Uint<LIMBS>>` impl for `Uint<LIMBS>` might be handy here

tarcieri:
Still needs `Boxed*` support but it should be good for everything stack-allocated

PR #489, opened 2023-12-21
Expand integer trait

xuganyu96:
This is a another continuation of #436 after I lost track of where I overrode the master branch's changes.

### DIFF
--- a/src/uint/boxed.rs
+++ b/src/uint/boxed.rs
@@ -19,6 +19,7 @@ mod neg;
 mod neg_mod;
 mod shl;
 mod shr;
+mod sqrt;
 mod sub;
 mod sub_mod;
 
@@ -90,6 +91,13 @@ impl BoxedUint {
         self.limbs
             .iter()
             .fold(Choice::from(1), |acc, limb| acc & limb.is_zero())
+    }
+
+    /// Is this [`BoxedUint`] not equal to zero?
+    pub fn is_nonzero(&self) -> Choice {
+        // TODO: why not just !self.is_zero()?
+        !self.limbs.iter()
+            .fold(Choice::from(0), |acc, limb| acc | limb.is_zero())
     }
 
     /// Is this [`BoxedUint`] equal to one?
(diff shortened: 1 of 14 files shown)


src/uint/boxed/encoding.rs
```diff
@@ -131,6 +132,42 @@ impl BoxedUint {
...
+            }
+            res[nlimbs - i - 1] = Limb(Word::from_be_bytes(buf));
+            i += 1;
+        }
+
+        if err == 0 {
+            CtOption::new(Self { limbs: res.into() }, Choice::from(1))
+        } else {
+            CtOption::new(Self::zero_with_precision(bits_precision), Choice::from(0))
+        }
```
tarcieri, line 169:
```suggestion
        CtOption::new(Self { limbs: res.into() }, Choice::from((err == 0) as u8)
```

There may be a better way to construct the `Choice`, though I don't remember the possible values of `err` offhand

src/uint/boxed.rs
```diff
@@ -92,6 +94,14 @@ impl BoxedUint {
             .fold(Choice::from(1), |acc, limb| acc & limb.is_zero())
     }
 
+    /// Is this [`BoxedUint`] not equal to zero?
+    pub fn is_nonzero(&self) -> Choice {
+        // TODO: why not just !self.is_zero()?
+        self.limbs
+            .iter()
+            .fold(Choice::from(0), |acc, limb| acc | limb.is_nonzero().into())
```
tarcieri, line 102: `!self.is_zero()` is fine

src/uint/boxed/cmp.rs
```diff
@@ -10,6 +10,30 @@ use subtle::{
...
 };
 
+impl BoxedUint {
+    /// Returns the Ordering between `self` and `rhs` in variable time.
+    pub fn cmp_vartime(&self, rhs: &Self) -> Ordering {
+        debug_assert_eq!(self.limbs.len(), rhs.limbs.len());
+        let mut i = self.limbs.len() - 1;
+        loop {
+            // TODO: investigate if directly comparing limbs is faster than performing a
+            // subtraction between limbs
```
tarcieri, line 20: It could short-circuit, so that's definitely a yes

tarcieri (approved):
This is fine aside from addressing some inline TODOs.

There are various allocations that could be optimized away in the boxed implementations but we can do that as a followup.

tarcieri:
I'm going to go ahead and land this since it's been so long in the tooth.

@xuganyu96 if you'd like to open some followups for those TODOs or optimizations, that'd be great.

PR #490, opened 2023-12-21
Add `div_by_2()` to `Monty` trait

fjarri:
- add `Monty::div_by_2()` method
- make `MontyForm::div_by_2()` and `ConstMontyForm::div_by_2()` const 
- make the internal `div_by_2()` take an `Odd`-wrapped modulus

Thought about making it a trait (like `Square`), but it's somewhat tied to the Montgomery modulus.


fjarri:
Actually wait, let me rephrase the docstrings. They're currently a little confusing.

tarcieri (approved):

PR #496, opened 2023-12-23
Add `DivRemLimb` and `RemLimb` traits

fjarri:
This is the next step for https://github.com/entropyxyz/crypto-primes/pull/36

- add more bounds on `Monty::Params` (goes in line with the bounds on `Monty` and `Integer`);
- add `From<Limb>` bound for `Integer`;
- use original `num-bigint` instead of `num-bigint-dig` for compatibility with `num-modular` (used in tests); 
- fix some vartime usage in constant time division methods
- add `DivRemLimb` and `RemLimb` traits (and separate `rem_limb()` methods - there's a significant speedup, and these are the ones I actually use in `crypto-primes`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,8 +32,9 @@ zeroize = { version = "1", optional = true,  default-features = false }
 bincode = "1"
 criterion = { version = "0.5", features = ["html_reports"] }
 hex-literal = "0.4"
-num-bigint = { package = "num-bigint-dig", version = "0.8" }
+num-bigint = "0.4"
 num-integer = "0.1"
+num-modular = { version = "0.5", features = ["num-bigint"] }
 proptest = "1"
 rand_core = { version = "0.6", features = ["std"] }
 rand_chacha = "0.3"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,12 +78,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f30e7476521f6f8af1a1c4c0b8cc94f0bee37d91763d0ca2665f299b6cd8aec"
 
 [[package]]
-name = "byteorder"
-version = "1.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
-
-[[package]]
 name = "bytes"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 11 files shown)


fjarri:
Making a draft temporarily, need to add comments about the unwraps.

tarcieri:
> I don't need gcd_vartime() for crypto-primes, it was just added for completeness sake.

Since we have Bernstein-Yang now defined on both `Uint` and `BoxedUint`, perhaps it's unnecessary?

I can add a trait for computing GCD and bound `Integer` on it.

fjarri:
> Since we have Bernstein-Yang now defined on both Uint and BoxedUint (just landed in https://github.com/RustCrypto/crypto-bigint/pull/497), perhaps it's unnecessary?

But that's constant-time, correct? I'm using it in a vartime context, so running the whole thing leads to performance degradatation.

Cargo.toml
```diff
@@ -32,8 +32,10 @@ zeroize = { version = "1", optional = true,  default-features = false }
 bincode = "1"
 criterion = { version = "0.5", features = ["html_reports"] }
 hex-literal = "0.4"
-num-bigint = { package = "num-bigint-dig", version = "0.8" }
+num-bigint-dig = "0.8"
+num-bigint = "0.4"
```
tarcieri, line 36: Is there a particular reason we need both?

fjarri: That's exactly my question. Why was the `-dig` one used originally? I assumed it was because it had a better performance, so it was used as a reference in Boxed benchmarks.

tarcieri: If we can migrate fully to `num-bigint` we should. It was just simpler to start with `num-bigint-dig` to implement the gaps for `rsa` since it's what `rsa` already used

fjarri: So some preliminary tests: for `U1024+Limb`, const-time GCD takes 11us, and `gcd_limb_vartime()` takes 264ns. `gcd_vartime()` is quite slow though, because `rem_vartime()` is actually not much faster than `rem()`.

fjarri:
Actually let me roll it back. There are a few assorted considerations:
- The GCD and Jacobi symbol functions I need are somewhat specialized, and if `Integer` provides the required functionality, I can just keep them in `crypto-primes` for now. Plus I'm not exactly sure how to implement Jacobi symbol in terms of Bernstein-Yang (although I remember reading that it's possible).
- The vartime division algorithm we have is super inefficient, I want to make a separate PR for it.
- Vartime division-by-limb can be added, but that's also a separate PR (although the constant-time one is already fast enough since in my application I only need to do it once).

tarcieri (approved):

daxpedda:
> * Plus I'm not exactly sure how to implement Jacobi symbol in terms of Bernstein-Yang (although I remember reading that it's possible).

@fjarri just stumbled on https://www.shiftleft.org/papers/byj/byj.pdf.

PR #497, opened 2023-12-23
Add `BoxedUint::gcd`

tarcieri:
Support for computing the greatest common divisor of two `BoxedUint`s using Bernstein-Yang

### DIFF
--- a/src/uint/boxed/gcd.rs
+++ b/src/uint/boxed/gcd.rs
@@ -1,0 +1,33 @@
+//! Support for computing greatest common divisor of two `BoxedUint`s.
+
+use super::BoxedUint;
+use crate::{modular::bernstein_yang, Odd};
+
+impl Odd<BoxedUint> {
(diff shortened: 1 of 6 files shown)


src/uint/boxed/gcd.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use super::BoxedUint;
+use crate::{modular::bernstein_yang, Odd};
+
+impl Odd<BoxedUint> {
+    /// Compute the greatest common divisor (GCD) of this number and another.
+    pub fn gcd(&self, rhs: &BoxedUint) -> BoxedUint {
+        bernstein_yang::boxed::gcd(self, rhs)
+    }
+}
```
tarcieri, line 11: Defining the method on `Odd` allows it to be infallible (since Bernstein-Yang requires `self` is odd), though perhaps there should be a more convenient way to call it and get a `CtOption`

PR #507, opened 2023-12-25
Add `BitOps` trait

fjarri:
- Add a trait for various bit counting/setting operations. Some were moved from the `Integer` trait.
- Extract the implementations of various methods into functions acting on `&[Limb]`, to be shared between `Uint` and `BoxedUint`.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -95,6 +95,7 @@ pub trait Integer:
     + for<'a> BitXor<&'a Self, Output = Self>
     + BitXorAssign
     + for<'a> BitXorAssign<&'a Self>
+    + BitOps
     + CheckedAdd
     + CheckedSub
     + CheckedMul
@@ -153,24 +154,6 @@ pub trait Integer:
 
     /// The value `1`.
     fn one() -> Self;
-
-    /// Calculate the number of bits required to represent a given number.
-    fn bits(&self) -> u32;
-
-    /// Calculate the number of bits required to represent a given number in variable-time with
-    /// respect to `self`.
-    fn bits_vartime(&self) -> u32;
-
-    /// Precision of this integer in bits.
-    fn bits_precision(&self) -> u32;
-
-    /// Precision of this integer in bytes.
-    fn bytes_precision(&self) -> usize;
-
-    /// Calculate the number of leading zeros in the binary representation of this number.
-    fn leading_zeros(&self) -> u32 {
-        self.bits_precision() - self.bits()
-    }
 
     /// Number of limbs in this integer.
     fn nlimbs(&self) -> usize;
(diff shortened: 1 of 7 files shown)


src/traits.rs
```diff
@@ -484,6 +467,69 @@ pub trait RemLimb: Sized {
     fn rem_limb_with_reciprocal(&self, reciprocal: &Reciprocal) -> Limb;
 }
 
+/// Bit counting and bit operations.
+pub trait BitOperations {
```
tarcieri, line 471: I think it might be a bit more idiomatic to call this `BitOps`, but this is fine

src/uint/boxed/bits.rs
```diff
@@ -60,55 +51,24 @@ impl BoxedUint {
 
     /// Calculate the number of trailing zeros in the binary representation of this number.
     pub fn trailing_zeros(&self) -> u32 {
-        let mut count = 0;
-        let mut nonzero_limb_not_encountered = Choice::from(1u8);
```
tarcieri, line 64: I guess we're losing `subtle`'s optimization barriers here (and elsewhere).

It might be good to examine the assembly generated by the `ConstChoice` version on x86/ARM and ensure LLVM isn't doing anything clever with it.

src/traits.rs
fjarri: I'll rename it.

src/uint/boxed/bits.rs
fjarri: Honestly the whole constant-time business is a lot of uncertainty and hand-waving. `subtle` quotes https://www.chosenplaintext.ca/open-source/rust-timing-shield/security/ as a basis of its implementation, but the example from there does not produce jumps in the assembly according to Godbolt. Neither do some of the bit-counting functions that use `ConstChoice` I tried. I would like to see a working example of a supposedly constant-time code that generates a non-constant-time assembly.

On a side note, I wonder if it's possible to come up with a trapdoor function that would obfuscate the boolean 0/1 and convert it into a mask, without needing to use `unsafe` and volatile pointers.

fjarri: Also filed #513 for the discussion

src/traits.rs
fjarri: Renamed

src/uint/boxed/bits.rs
tarcieri: Hmm, thought I already responded here but I guess my response was lost.

> subtle quotes https://www.chosenplaintext.ca/open-source/rust-timing-shield/security/ as a basis of its implementation, but the example from there does not produce jumps in the assembly according to Godbolt.

That both that example and the [example from nadeko](https://github.com/klutzy/nadeko?tab=readme-ov-file#why) are cause for concern, and rustc/LLVM is of course free to do so, hence inserting optimization barriers to at least hint to LLVM about what it shouldn't optimize, which I consider a recommended best practice which is [endorsed by LLVM engineers](https://github.com/The-DevX-Initiative/RCIG_Coordination_Repo/issues/55#issuecomment-926996240).

Alternatively we could try to automate analysis of the generated assembly across commonly used platforms to spot the insertion of branches, but that seems much harder.

In the meantime, I'll spot check these and if the generated ASM looks fine on x86/ARM I'm ok to merge.

tarcieri: Here's the relevant functions in godbolt, although I had to remove the `#[inline(always)]` annotations to be able to see the generated assembly:

https://godbolt.org/z/Y9fTjPhc7

The jumps I am seeing all relate to the loops, which the compiler can't elide without knowledge of the length of the slice (which should ideally be resolved via `#[inline(always)]`). There's no way to avoid them with `BoxedUint` though since the slice length is determined at runtime, so this is actually an accurate picture for that use case.

Otherwise I don't see any data-dependent branching. I was surprised to see LLVM making use of the `cmov*` family on x86(-64) for handling conditional assignment as in the past LLVM specifically rewrote cmov with branches in the [X86CmovConversion](https://llvm.org/doxygen/X86CmovConversion_8cpp.html) pass, however [it seems it now emits cmov depending on a new set of performance heuristics](https://discourse.llvm.org/t/rfc-cmov-vs-branch-optimization/6040)

All that said, this seems fine to me, so I'm going to go ahead and merge.

fjarri: Yes, that was my experience as well. I tried on multiple occasions to make the compiler sneakily insert jump instructions where there shouldn't be any, and never succeeded. The example from `nadeko` compiles into jump-free code as well.

But really, of course, this should be handled on LLVM and Rust compiler side, without making the programmer perform magic rituals with `unsafe` and `volatile` hoping that the gods will be merciful this time (or in our case, it's more alike performing a magic ritual helping the sun to rise).

tarcieri: That was the goal of the [secret types RFC](https://github.com/rust-lang/rfcs/pull/2859) however it was closed (and the corresponding LLVM work it would've required never happened either)

PR #510, opened 2023-12-25
Add `RandomBits` trait

fjarri:
Adds a trait for generating a random number of a given bit size.

Can be improved by using masking instead of shift.

- Is there a better way to express this API?
- Should it be constant-time in `bit_length`? 
- Should the inherent `BoxedUint::random()` take `bits_precision` as well?

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -335,17 +335,23 @@ impl std::error::Error for RandomBitsError {}
 #[cfg(feature = "rand_core")]
 pub trait RandomBits: Sized {
     /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`.
+    ///
+    /// A wrapper for [`RandomBits::try_random_bits`] that panics on error.
     fn random_bits(rng: &mut impl CryptoRngCore, bit_length: u32) -> Self {
         Self::try_random_bits(rng, bit_length).expect("try_random_bits() failed")
     }
 
     /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`.
+    ///
+    /// This method is variable time wrt `bit_length`.
     fn try_random_bits(
         rng: &mut impl CryptoRngCore,
         bit_length: u32,
     ) -> Result<Self, RandomBitsError>;
 
-    /// Generate a cryptographically secure random value.
+    /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`,
+    /// returning an integer with the closest available size to `bits_precision`
+    /// (if the implementing type supports runtime sizing).
     ///
     /// A wrapper for [`RandomBits::try_random_bits_with_precision`] that panics on error.
     fn random_bits_with_precision(
@@ -360,6 +366,8 @@ pub trait RandomBits: Sized {
     /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`,
     /// returning an integer with the closest available size to `bits_precision`
     /// (if the implementing type supports runtime sizing).
+    ///
+    /// This method is variable time wrt `bit_length`.
     fn try_random_bits_with_precision(
         rng: &mut impl CryptoRngCore,
         bit_length: u32,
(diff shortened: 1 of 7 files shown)


src/traits.rs
```diff
@@ -288,6 +288,13 @@ pub trait Random: Sized {
     fn random(rng: &mut impl CryptoRngCore) -> Self;
 }
 
+/// Random number generation support.
+#[cfg(feature = "rand_core")]
+pub trait RandomBits: Sized {
+    /// Generate a cryptographically secure random value.
+    fn random_bits(rng: &mut impl CryptoRngCore, bit_length: u32, bits_precision: u32) -> Self;
```
tarcieri, line 295: It seems like there are several failure cases with this API. Perhaps it should return a `Result` or `CtOption`?

In `crypto-common` we migrated to returning e.g. `rand_core::Error` and using `try_fill_bytes` so as to also handle potential RNG failures in various environments

fjarri: I don't know how much good will that do for the user. It's not something you can recover from, or pass it on to a higher level. If one of those panics triggers, it's just a plain bug in the user code.

tarcieri: Returning a `Result` at least gives the caller a choice, versus always crashing the entire program.

(As it were, this crate was originally written with the goal of minimizing panics, but it seems that's gone out the window, especially with `Boxed*`)

tarcieri: How about having a `try_random_bits` method that returns a result, and a provided `random_bits` method which panics?

That follows the same pattern as the [`RngCore` trait](https://docs.rs/rand_core/0.6.4/rand_core/trait.RngCore.html).

fjarri: Good idea, I'll do that. Should it actually propagate the `RngCore` error?

tarcieri: That seems like a reasonable place to start, though you may need to define a separate error type to capture cases like `bit_length` > `bits_precision` (where `rand_core::Error` can be a variant like `RngError`)

fjarri: Yes, that's what I meant. Wrap the inner error and add our error cases.

tarcieri: I guess you could use a custom error in the [`CUSTOM_START`](https://docs.rs/rand_core/latest/rand_core/struct.Error.html#associatedconstant.CUSTOM_START) range and the `From<NonZeroU32>` impl to construct a `rand_core::Error` if you wanted to go that route, though that may not be helpful from a usability perspective

fjarri:
Expanded the API - added fallible methods, and convenience methods allowing one omit `bits_precision` (with behavior depending on whether it's `Uint` or `BoxedUint`). The main question that still remains - should it be constant-time in `bit_length`, or even whether it is expected to be. That is, should I add `_vartime` suffixes? Or redo it to be constant-time?

src/traits.rs
```diff
@@ -274,6 +274,65 @@ pub trait Random: Sized {
     fn random(rng: &mut impl CryptoRngCore) -> Self;
 }
 
+/// Possible errors of [`RandomBits::try_random_bits`].
+#[cfg(feature = "rand_core")]
+#[derive(Debug)]
+pub enum RandomBitsError {
```
tarcieri, line 280: For completeness, this could probably use a `Display` impl along with a `std::error::Error` impl when the `std` feature is enabled

fjarri:
Also the inherent `BoxedUint::random()` is identical in behavior to `BoxedUint::random_bits()` now. Should it be removed?

tarcieri:
> The main question that still remains - should it be constant-time in bit_length, or even whether it is expected to be. 

@fjarri perhaps we should document it, but I consider everything to be implicitly variable-time around the integer size.

I guess "bit length" is a slightly different question, particularly in regard to that versus the precision / capacity of the underlying type.

tarcieri:
> Also the inherent BoxedUint::random() is identical in behavior to BoxedUint::random_bits() now. Should it be removed?

That's fine I guess

src/traits.rs
fjarri: Done

tarcieri (approved):

tarcieri:
Perhaps `Random` could be kept but with a blanket impl for `FixedInteger` + `RandomBits`? (and possibly a different name)

PR #511, opened 2023-12-25
Variable-time division improvements

fjarri:
- Implement the variable-time division algorithm from "Handbook of Applied Cryptography"

TODO:
- The divisor should be normalized first so that the MSB is set (see the notes in the Handbook). But that might put the dividend out of bounds of `Uint`.
- (tangential) `mul_mod_special()` is variable time in `c` - need to either change the name, or to make a constant-time equivalent of `mul_rem_vartime` that it uses. 
- (also tangential) the algorithm in `mul_mod_special()` is not actually the one from "Handbook of Applied Cryptography"; there are some modifications made, and I'm trying to understand why they work. It needs comments.
- The division-by-limb methods can have variable-time equivalents, but we need to see if they are actually faster. The variable time algorithms can be introduced at `div_rem_limb_with_reciprocal()`, `short_div()`, and possibly `div2by1()` (although the latter mostly has additions and multiplications, and only a few constant-time comparisons)

### DIFF
--- a/src/primitives.rs
+++ b/src/primitives.rs
@@ -66,8 +66,18 @@ pub(crate) const fn mac(a: Word, b: Word, c: Word, carry: Word) -> (Word, Word) 
     (ret as Word, (ret >> Word::BITS) as Word)
 }
 
-/// Computes `(a * b) % d`.
+/// Computes `(a * b) % d`. Not constant time in `d`.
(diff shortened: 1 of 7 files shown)


tarcieri:
Closing in favor of #608

PR #518, opened 2023-12-27
Add constant-time `MontyParams::new`

tarcieri:
Adds a constant-time equivalent to `MontyParams::new_vartime`


fjarri:
Is there a big performance difference between vartime and constant-time here? Could we just remove the vartime one?

tarcieri:
See #521. The `new_vartime` constructor is 2.75x faster

PR #523, opened 2023-12-27
Handle even `s` in `Uint::inv_mod`

tarcieri:
This is an oversight from #501 which switched to using Bernstein-Yang inversions for `Uint::inv_odd_mod`.

The new implementation in #501 assumes `s` is always `Odd` but didn't actually check for that and set the `ConstCtOption` to be "none" accordingly. This PR fixes that.

### DIFF
--- a/src/const_choice.rs
+++ b/src/const_choice.rs
@@ -186,7 +186,7 @@ impl PartialEq for ConstChoice {
 #[derive(Debug, Clone)]
 pub struct ConstCtOption<T> {
     value: T,
-    pub(crate) is_some: ConstChoice,
+    is_some: ConstChoice,
 }
 
 impl<T> ConstCtOption<T> {
--- a/src/const_choice.rs
+++ b/src/const_choice.rs
@@ -243,7 +243,7 @@ impl<T> ConstCtOption<T> {
 
     /// Apply an additional [`ConstChoice`] requirement to `is_some`.
     #[inline]
-    pub const fn and_choice(mut self, is_some: ConstChoice) -> Self {
+    pub(crate) const fn and_choice(mut self, is_some: ConstChoice) -> Self {
         self.is_some = self.is_some.and(is_some);
         self
     }
--- a/src/const_choice.rs
+++ b/src/const_choice.rs
@@ -186,7 +186,7 @@ impl PartialEq for ConstChoice {
 #[derive(Debug, Clone)]
 pub struct ConstCtOption<T> {
     value: T,
-    is_some: ConstChoice,
+    pub(crate) is_some: ConstChoice,
 }
 
 impl<T> ConstCtOption<T> {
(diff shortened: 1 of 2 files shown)


tarcieri:
This could probably use a test for this edge case.

src/const_choice.rs
```diff
@@ -186,7 +186,7 @@ impl PartialEq for ConstChoice {
 #[derive(Debug, Clone)]
 pub struct ConstCtOption<T> {
     value: T,
-    is_some: ConstChoice,
+    pub(crate) is_some: ConstChoice,
```
tarcieri, line 189: This isn't great but I'm not sure there's a better option without something like `const_mut_refs`

src/uint/inv_mod.rs
```diff
@@ -106,7 +106,10 @@ impl<const LIMBS: usize> Uint<LIMBS> {
 
         // Decompose `self` into RNS with moduli `2^k` and `s` and calculate the inverses.
         // Using the fact that `(z^{-1} mod (m1 * m2)) mod m1 == z^{-1} mod m1`
-        let maybe_a = self.inv_odd_mod(&Odd(s));
+        let s_is_odd = s.is_odd();
+        let mut maybe_a = self.inv_odd_mod(&Odd(s));
+        maybe_a.is_some = maybe_a.is_some.and(s_is_odd);
```
fjarri, line 111: This seems somewhat like https://doc.rust-lang.org/std/option/enum.Option.html#method.filter

fjarri: What I mean is, I would prefer it to be a method instead of mutating `maybe_a`

tarcieri: Hmm yeah, something like that might be doable in a `const fn` context

tarcieri: Oh maybe not... it would need an actual closure (as opposed to a function pointer) to be able to query `s_is_odd` from the outer scope and still be callable from a `const fn`.

tarcieri: So really it would have to be a method that accepts a `ConstChoice` and applies an `&=` style mutation, but it could consume the original `ConstCtOption` and return a new one.

I guess the question is what to call such a method.

fjarri: To implement `filter`, yes, but we only need a kind of `flatten()` functionality. Some `fn ___(&self, is_some: ConstChoice) -> Self`. `some_if()`?

tarcieri: [222685d](https://github.com/RustCrypto/crypto-bigint/pull/523/commits/222685d6e4d59c028eb5b0e054845ddd7c6842af) adds `ConstCtOption::and_choice` (for lack of a better name) to wrap up the mutation

tarcieri: I'm open to better options for a name

fjarri: Maybe make it `pub(crate)` for now, so that it's easier to change later if we come up with something better

PR #525, opened 2023-12-27
Add "vartime" suffix to Monty::new_params()

fjarri:
The implementation for `Uint` was already vartime. The `BoxedUint` one wasn't, changed to vartime.

It would be logical to have `new_params()` for the constant-time implementation, but implementing it for `Uint` requires additional bounds with `Split`/`Concat` which have to be propagated everywhere. Not sure what the best course of action here is.

One possibility is perhaps splitting out the "convertible to Montgomery" in a separate trait.

### DIFF
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -240,8 +240,8 @@ impl Monty for BoxedMontyForm {
     type Integer = BoxedUint;
     type Params = BoxedMontyParams;
 
-    fn new_params(modulus: Odd<Self::Integer>) -> Self::Params {
-        BoxedMontyParams::new(modulus)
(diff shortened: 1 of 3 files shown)


tarcieri:
>  implementing it for Uint requires additional bounds with Split/Concat which have to be propagated everywhere. Not sure what the best course of action here is.

Elsewhere instead of a wide `Uint`, split lo/hi components are used for the wide multiplication output (e.g. `Uint::rem_wide_vartime`), which avoids the need for the bounds.

fjarri:
I mean, since `MontyParams` are a part of `Integer`, if `new_params()` require this bound, all the `Integer` stuff will, too, even if it technically doesn't use it.

tarcieri:
It wouldn't need the bounds if it used a "split" implementation strategy like `Uint::rem_wide_vartime`

tarcieri (approved):

PR #533, opened 2023-12-29
Additional methods for `Integer` and `Monty`

fjarri:
- Add `Integer::from_limb_like()`, `one_like()`, and `Zero::zero_like()`. This is needed in [#39](https://github.com/entropyxyz/crypto-primes/pull/39)
- Add `Monty::params()` and `as_montgomery()` 

I do not entirely like `from_limb_like`, but that's all I could come up with.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -800,6 +800,12 @@ pub trait Monty:
     /// Returns one in this representation.
     fn one(params: Self::Params) -> Self;
 
+    /// Returns the parameter struct used to initialize this object.
+    fn params(&self) -> &Self::Params;
+
+    /// Access the value in Montgomery form.
+    fn as_montgomery(&self) -> &Self::Integer;
+
     /// Performs division by 2, that is returns `x` such that `x + x = self`.
     fn div_by_2(&self) -> Self;
 }
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -158,11 +158,6 @@ pub trait Integer:
     /// The value `1`.
     fn one() -> Self;
 
-    /// The value `0` with the same precision as `other`.
-    fn zero_like(other: &Self) -> Self {
-        Self::from_limb_like(Limb::ZERO, other)
-    }
-
     /// The value `1` with the same precision as `other`.
     fn one_like(other: &Self) -> Self {
         Self::from_limb_like(Limb::ONE, other)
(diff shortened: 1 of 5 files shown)


src/traits.rs
```diff
@@ -158,6 +158,19 @@ pub trait Integer:
     /// The value `1`.
     fn one() -> Self;
 
+    /// The value `0` with the same precision as `other`.
+    fn zero_like(other: &Self) -> Self {
+        Self::from_limb_like(Limb::ZERO, other)
+    }
```
tarcieri, line 164: Perhaps this should be defined on the `Zero` trait?

src/traits.rs
```diff
@@ -158,6 +158,19 @@ pub trait Integer:
...
+
+    /// The value `1` with the same precision as `other`.
+    fn one_like(other: &Self) -> Self {
+        Self::from_limb_like(Limb::ONE, other)
+    }
```
tarcieri, line 169: I've wondered if we should have a `One` trait, similar to `num_traits::One`.

The only reason we don't use `num_traits::One` is so `is_one` can return `Choice`, as it were.

src/traits.rs
```diff
@@ -787,6 +800,12 @@ pub trait Monty:
     /// Returns one in this representation.
     fn one(params: Self::Params) -> Self;
 
+    /// Returns the parameter struct used to initialize this object.
+    fn params(&self) -> &Self::Params;
+
+    /// Access the value in Montgomery form.
+    fn as_montgomery(&self) -> &Self::Integer;
```
tarcieri, line 807: I'm not wild about this name (or some of the others) but we can bikeshed it in another issue

src/traits.rs
fjarri: I thought about it, it would have to be defined explicitly for every case then, now it's just a default implementation.

src/traits.rs
fjarri: It just mimics the inherent method name we already have.

tarcieri: Yeah, definitely, I just find the existing naming confusing. I should've prefixed it with "sidebar"

src/traits.rs
tarcieri: It could still have a provided impl with the appropriate bounds, e.g. `AsMut<[Limb]> + Clone`

fjarri: Hm, something like this?
```rust
trait Zero
    ...
    fn zero_like(other: &Self) -> Self where Self: Clone {
        let mut ret = other.clone();
        ret.set_zero();
        ret
    }
}
```

A small downside here is that it will do the unnecessary copy just to fill it with zeros later, but perhaps it's not that impactful in most cases.

tarcieri: Yeah, exactly.

It does have that potential disadvantage, however the LLVM optimizer often seems smart enough to elide those writes if it sees they're being subsequently overwritten.

fjarri: Fixed

src/traits.rs
fjarri: Seems logical to have such a trait. If we're redefining `num_traits::Zero`, might as well redefine `One` too.

tarcieri (approved):

PR #538, opened 2023-12-29
Fix dudect build

tarcieri:


tarcieri:
It takes 7 minutes to run dudect, and it just covers `Uint::rem`.

Not sure it's worth it, but I'll leave it in for now.

Issue #283, opened 2023-11-04
improve documentation with usage examples


drcapybara:
I understand that this library is a work in progress, and a substantial amount of effort and work has and continues to go into bringing awesome functionality to life. So this isn't at all a criticism, but the documentation for public-facing apis, particularly in the modular math department, feels lacking. This has made it difficult to get started with the library.

Rust has one of the best documentation systems I've had the pleasure of using so far. If the code for some of the functions in the API is considered stable, and would benefit from documentation, Id like to propose that we begin to add usage examples. This will enhance the library by:

1. Improving readability of the public facing apis
2. speeding up onboarding to those who may be switching from other arbitrary-precision libraries by reducing the work of searching the code base
3. improve safety by documenting best practices in the usage examples
4. increase efficiency for the same reason as above. someone who needs a strong, constant-time library but might not have as much experience with advanced algorithms such as montgomery reduction, would benefit from having the documentation point them directly to the best tool for the job.

I fully acknowledge that this only makes sense if the functions we want to document are stable. If things are still developing and will dramatically change soon, we should we wait for stability.  If it makes sense to do all of this, I volunteer as tribute for this task if no one else feels strongly compelled. [Here is a sample](https://github.com/drcapybara/capyCRYPT/blob/4054f5fa3e2f17705b04d0b06c3cb04795337445/src/ops.rs#L431) of some of my work from my own cryptosystem Ive built as an academic exercise in cryptographic algorithm design:

```rust
/// # Schnorr Signatures
/// Signs a [`Message`] under passphrase pw.
///
/// ## Algorithm:
/// * `s` ← kmac_xof(pw, “”, 512, “K”); s ← 4s
/// * `k` ← kmac_xof(s, m, 512, “N”); k ← 4k
/// * `𝑈` ← k*𝑮;
/// * `ℎ` ← kmac_xof(𝑈ₓ , m, 512, “T”); 𝑍 ← (𝑘 – ℎ𝑠) mod r
///
/// ## Arguments:
/// * key: &[`KeyPair`], : reference to KeyPair.
/// * d: u64: encryption security strength in bits. Can only be 224, 256, 384, or 512.
///
/// ## Assumes:
/// * Some(key.priv_key)
///
/// ## Usage
/// ```
/// use capycrypt::{
///     Signable,
///     KeyPair,
///     Message,
///     sha3::aux_functions::byte_utils::get_random_bytes,
///     curves::EdCurves::E448};
/// // Get random 5mb
[...]

PR #540, opened 2024-01-02
docs: a few usage examples for div

drcapybara:
Adds a few usage examples for div operations. ref issue #283.

Additionally modifies:

```rust
pub const fn wrapping_rem
```

into:
```rust
pub const fn wrapping_rem_vartime
```

since it uses ```rem_vartime``` internally.

### DIFF
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -159,7 +159,6 @@ impl<const LIMBS: usize> Uint<LIMBS> {
 
     /// Computes `self` % 2^k. Faster than reduce since its a power of 2.
     /// Limited to 2^16-1 since Uint doesn't support higher.
-    /// TODO: this is not constant-time.
     ///
     /// ### Usage:
     /// ```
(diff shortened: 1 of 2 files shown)


src/uint/div.rs
```diff
@@ -159,6 +159,19 @@ impl<const LIMBS: usize> Uint<LIMBS> {
 
     /// Computes `self` % 2^k. Faster than reduce since its a power of 2.
     /// Limited to 2^16-1 since Uint doesn't support higher.
+    /// TODO: this is not constant-time.
```
tarcieri, line 162: Sidebar: we should perhaps consider renaming this `rem2k_vartime` /cc @fjarri

src/uint/div.rs
```diff
@@ -203,25 +216,65 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+    ///
+    /// ### Usage:
+    /// ```
+    /// use crypto_bigint::{U448, NonZero, subtle::{Choice,CtOption}};
+    ///
+    /// let a = U448::from(10_u64);
+    /// let remainder_option = NonZero::new(U448::from(3_u64))
+    ///     .map(|b| a.rem(&b));
+    ///
+    /// assert!(bool::from(remainder_option.is_some()), "Reduction by zero");
```
tarcieri, line 271: I'd find this more intuitive without the comment:

```suggestion
    /// assert!(bool::from(remainder_option.is_some()));
```

src/uint/div.rs
fjarri: I imagine making it constant-time would also be pretty straightforward.

Although I am not sure we need it at all in the API.

fjarri:
I just want to reiterate my position (last time, I promise) that usage examples for simple arithmetic operations like these are entirely excessive.

drcapybara:
> I just want to reiterate my position (last time, I promise) that usage examples for simple arithmetic operations like these are entirely excessive.

```wrapping_rem_vartime``` and ```rem2k_vartime``` examples might be excessive, if you would like those removed, Im not opposed to it. The others did take me considerable time to get working without examples, it would be nice to leave them if possible.

tarcieri (approved):

PR #541, opened 2024-01-05
Boxed sqrt reduce cloning

xuganyu96:
Removed one of the inline TODO's for reducing allocation. Also added some benchmarks.

I am not sure if it's more appropriate to use `.clone_from` on `BoxedUint` or the limbs; it seems like alloc-free cloning is done on the limbs elsewhere. Reviewer please advise.

Interestingly the benchmarks did not show statistically significant improvements between the code change. I am not aware of benchmarking on memory usage, either.

### DIFF
--- a/src/uint/boxed/sqrt.rs
+++ b/src/uint/boxed/sqrt.rs
@@ -22,12 +22,13 @@ impl BoxedUint {
 
         // Repeat enough times to guarantee result has stabilized.
         let mut i = 0;
-        // TODO: avoid this clone
         let mut x_prev = x.clone(); // keep the previous iteration in case we need to roll back.
 
         // TODO (#378): the tests indicate that just `Self::LOG2_BITS` may be enough.
         while i < self.log2_bits() + 2 {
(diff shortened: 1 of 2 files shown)


src/uint/boxed/sqrt.rs
```diff
@@ -22,12 +22,13 @@ impl BoxedUint {
...
         let mut i = 0;
-        // TODO: avoid this clone
         let mut x_prev = x.clone(); // keep the previous iteration in case we need to roll back.
 
         // TODO (#378): the tests indicate that just `Self::LOG2_BITS` may be enough.
         while i < self.log2_bits() + 2 {
-            x_prev = x.clone(); // TODO: can we avoid this clone?
+            // Use clone_from to avoid allocation
+            // TODO: check if `x_prev.clone_from(&x)` does the same thing
+            x_prev.limbs.clone_from(&x.limbs);
```
tarcieri, line 31:
```suggestion
            x_prev.limbs.copy_from_slice(&x.limbs);
```

tarcieri (approved):

PR #550, opened 2024-01-10
Fix `Uint::shr(_vartime)` and `Uint::shl(_vartime)`

tarcieri:
The amount to shift was being miscomputed. It should've been a saturating-like operation, where any shift larger than the precision returns zero, but instead a modulus (which isn't constant time) was being applied instead.

This also adds proptests to ensure correct behavior.


tarcieri:
Confirm this fixes the test failures on https://github.com/RustCrypto/elliptic-curves/pull/1011

PR #560, opened 2024-01-16
CI: run tests under miri

tarcieri:
Should help identify potentially problematic unsafe code


tarcieri:
Wow, these take forever to run unfortunately.

I guess we should perhaps recalibrate the size of the boxed tests to be more amenable to this (and users on slower computers).

tarcieri:
Okay, running the minimum viable configuration of `--no-default-features --lib` for the tests, Miri passes in 4m while the cross tests take 3.

That seems OK to me.

PR #562, opened 2024-01-17
demo of const_monty_form

drcapybara:
demonstrates usage of various functionality of the `modular` module

### DIFF
--- a/src/modular.rs
+++ b/src/modular.rs
@@ -19,9 +19,9 @@
 //! # Usage
 //!
 //! Suppose some prover claims to have a valid signature for some message. Perfect zero-knowledge
-//! proof-of-posession of this message-signature pair can be obtained with the 
+//! proof-of-posession of this message-signature pair can be obtained with the
 //! partially-homomorphic properties of RSA:
-//! 
+//!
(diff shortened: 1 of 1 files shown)


tarcieri:
I think having an example is good.

I worry this might be a bit much (i.e. is the demonstration of `const_monty_form!` getting lost in the other details of the algorithm?), and somewhat question the choice of RSA as an algorithm for demonstrating these features, since in real-world settings the RSA modulus is always dynamic as opposed to something you'd hardcode as a constant (as compared to e.g. an elliptic curve field modulus, which is actually constant)

drcapybara:
> I think having an example is good.
> 
> I worry this might be a bit much (i.e. is the demonstration of `const_monty_form!` getting lost in the other details of the algorithm?), and somewhat question the choice of RSA as an algorithm for demonstrating these features, since in real-world settings the RSA modulus is always dynamic as opposed to something you'd hardcode as a constant (as compared to e.g. an elliptic curve field modulus, which is actually constant)

youre totally right this is way too extra, Ill come back with something more focused and a little less dense

PR #566, opened 2024-01-26
In-place cloning of Montgomery form

xuganyu96:
Originally meant to allow boxed Monty to be cloned without additional allocation. Also includes a change to the `Monty` trait to expose `clone_from_montgomery` to generics

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -218,7 +218,7 @@ dependencies = [
 
 [[package]]
 name = "crypto-bigint"
-version = "0.6.0-pre.11"
+version = "0.6.0-pre.12"
 dependencies = [
  "bincode",
  "criterion",
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -220,6 +220,14 @@ impl BoxedMontyForm {
         self.montgomery_form.clone()
     }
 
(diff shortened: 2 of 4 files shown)


tarcieri:
Why not just use `Clone::clone_from`?

xuganyu96:
> Why not just use `Clone::clone_from`?

My bad for not realizing that the option is here.

tarcieri:
Note that the implementation of `clone_from` isn't quite in-place yet the way this PR was.

We'd need to deal with mismatched numbers of limbs to make it work like what you had (in which case panicking is probably fine, so `clone_from` doesn't have weird performance edge cases)

xuganyu96:
> Note that the implementation of `clone_from` isn't quite in-place yet the way this PR was.
> 
> We'd need to deal with mismatched numbers of limbs to make it work like what you had (in which case panicking is probably fine, so `clone_from` doesn't have weird performance edge cases)

I originally raised this PR because of https://github.com/entropyxyz/crypto-primes/issues/41, in which fjarri mentioned the Lucas testing having a lot of `.clone()` calls. After running a `cargo flamegraph`, however, I found that the bulk of the time of generating prime is spent on instantiating MontyForm so maybe allocating is not the bottleneck there.

PR #568, opened 2024-01-29
Enable `clippy::must_use_candidate` lint + autofix

tarcieri:
Many of the computations we perform have results where failure to use them likely constitutes a bug.

`clippy` can autodetect these cases, and `clippy fix` can auto-apply them.

### DIFF
--- a/src/const_choice.rs
+++ b/src/const_choice.rs
@@ -274,6 +274,7 @@ impl<T> From<ConstCtOption<T>> for Option<T> {
 impl<const LIMBS: usize> ConstCtOption<Uint<LIMBS>> {
     /// This returns the underlying value if it is `Some` or the provided value otherwise.
     #[inline]
+    #[must_use]
     pub const fn unwrap_or(self, def: Uint<LIMBS>) -> Uint<LIMBS> {
         Uint::select(&def, &self.value, self.is_some)
     }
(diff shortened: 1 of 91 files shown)


fjarri:
This seems extremely inefficient to me. So any method that doesn't take `&mut self` and returning something (so, like, 95% of all methods at least) must be declared as `must_use`? I thought that attribute is for cases where it's easy to make a mistake and forget to check the result, e.g. `verify(...) -> Result<(), Error>`.

fjarri:
I wish there was a lint where `must_use` was the default (under the conditions above), and you could opt out methods from it.

tarcieri:
@fjarri it's a totally optional lint we can enable. Unfortunately there is no way of changing the default.

Are you concerned about compile times or what?

Since this crate is intended for safety-oriented code, and failure to use the results of these functions pretty much necessarily implies a bug, I think it might be worth enabling.

fjarri:
Not compile times, just general code noise.

tarcieri:
Yeah, it is a lint from the `pedantic` category that carries with it some false positive potential, although looking through what it annotated here, I have a hard time finding anything where I wouldn't agree that "failure to use this return value would constitute a bug"

fjarri:
Any method that returns a value should have that value used, I just don't see why clippy needs an explicit annotation for that. An annotation that applies to almost every method in the codebase is just noise.

tarcieri:
The annotation isn't for clippy: it's so downstream users of `crypto-bigint` get a `must_use` warning, emitted by rustc itself, in the event they fail to use a value they computed in their downstream code.

It does unfortunately cause a lot of churn in `crypto-bigint` itself, but that's just how `must_use` and the Rust compiler work.

tarcieri:
This seems controversial enough I guess I'll just close it.

PR #578, opened 2024-02-29
Some efficiency improvements in `div_rem`

tcoratger:
Improvements are made to the main loop of `div_rem` and `div_rem_vartime`.
- `r` as a result of subtraction with borrow does not need to be mutable because it is not strictly speaking modified afterwards.
- The declaration of `from_word_mask` is done twice in a similar way each time through the loop, a performance gain can be found by making a single declaration for this choice, to be reused in the two calls.


tarcieri (approved):
Thanks!

PR #600, opened 2024-05-30
Make the `Monty::new` borrow the `Uint`

### LINKED ISSUE #599: `Monty::new` takes values, but impl takes ref
The `Monty` trait's [`new` method](https://github.com/RustCrypto/crypto-bigint/blob/6ccbe7412ffbb041e79c520be623af330ee8730f/src/traits.rs#L800) takes its arguments by value, but [the implementation for `MontyForm`](https://github.com/RustCrypto/crypto-bigint/blob/6ccbe7412ffbb041e79c520be623af330ee8730f/src/modular/monty_form.rs#L243) uses references when calling into [`MontyForm::new`](https://github.com/RustCrypto/crypto-bigint/blob/6ccbe7412ffbb041e79c520be623af330ee8730f/src/modular/monty_form.rs#L156).

What are your thoughts on changing the trait in v0.6 to take a reference as well? The `new` method only ever need read access to the data so why impose a clone/copy on users if they need to keep using their `Integer`?

Sort of like so:
```
src/traits.rs --- Rust
799     /// Convert the value into the representation using precomputed data.
800     fn new(value: &Self::Integer, params: Self::Params) -> Self;
801 

src/modular/monty_form.rs --- Rust
243     fn new(value: &Self::Integer, params: Self::Params) -> Self {
244         MontyForm::new(value, params)
245     } 

```

dvdplm:
Change the `Monty` trait's `new` method to take a ref to the `Integer` instead of ownership.

Fixes #599

### DIFF
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -244,8 +244,8 @@ impl Monty for BoxedMontyForm {
         BoxedMontyParams::new_vartime(modulus)
     }
 
-    fn new(value: Self::Integer, params: Self::Params) -> Self {
-        BoxedMontyForm::new(value, params)
+    fn new(value: &Self::Integer, params: Self::Params) -> Self {
+        BoxedMontyForm::new(value.clone(), params)
     }
 
     fn zero(params: Self::Params) -> Self {
--- a/src/modular/monty_form.rs
+++ b/src/modular/monty_form.rs
@@ -240,8 +240,8 @@ impl<const LIMBS: usize> Monty for MontyForm<LIMBS> {
         MontyParams::new_vartime(modulus)
     }
 
-    fn new(value: Self::Integer, params: Self::Params) -> Self {
-        MontyForm::new(&value, params)
+    fn new(value: &Self::Integer, params: Self::Params) -> Self {
+        MontyForm::new(value, params)
     }
 
     fn zero(params: Self::Params) -> Self {
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -789,7 +789,7 @@ pub trait Monty:
     fn new_params_vartime(modulus: Odd<Self::Integer>) -> Self::Params;
 
     /// Convert the value into the representation using precomputed data.
-    fn new(value: Self::Integer, params: Self::Params) -> Self;
+    fn new(value: &Self::Integer, params: Self::Params) -> Self;
 
     /// Returns zero in this representation.
     fn zero(params: Self::Params) -> Self;


tarcieri:
@fjarri I'm guessing this will need changes in `crypto-primes`

src/modular/boxed_monty_form.rs
```diff
@@ -244,8 +244,8 @@ impl Monty for BoxedMontyForm {
         BoxedMontyParams::new_vartime(modulus)
     }
 
-    fn new(value: Self::Integer, params: Self::Params) -> Self {
-        BoxedMontyForm::new(value, params)
+    fn new(value: &Self::Integer, params: Self::Params) -> Self {
+        BoxedMontyForm::new(value.clone(), params)
```
tarcieri, line 248: Hmm, I know I already hit approve but this has me second guessing if it's worth it. `MontyForm` is a `Copy` type but `BoxedMontyForm` is a heap-backed type and this is adding another heap allocation which could otherwise be avoided.

dvdplm: Yeah my heart sank a bit too when I had to do this. Then I figured that 1) a Box is cheap to clone and 2) the unboxed variant is probably what users that care about performance would prefer, so perhaps it makes sense to accommodate them?

tarcieri: The heap allocation is a tangible performance hit, whereas passing a `Copy` type by reference may have no performance impact, particularly with inlining

dvdplm: As a user of the API before this PR I find that I end up having to clone my `Integer`s more often than I'd like, given that `MontyForm` just reads the data. But people instantiating with `BoxedMontyForm` will need the same clones in their code yeah? It's less annoying for them I guess, because if they read the code they see that the `value` is actually mutated in place.

At the end of the day I think it comes down to what the API is trying to communicate. Taking ownership of the arguments signals that the users aren't supposed to re-use the values, but when it comes to "number types" I think the expectation is that you can actually continue using it (which is the case for `Copy` types of course).

Maybe this is tricky to decide because the underlying impls are so different. Would it make sense to try to refactor `BoxedMontyForm::new` to take a ref and then do the minimum requried amount of copying?

tarcieri: Why are you cloning a `Copy` type?

dvdplm: Because I'm using it in generic code, through the `Integer` trait (and the `Monty` associated type`) so I don't know if my users are going to use it with `MontyForm` or `BoxedMontyForm`.

fjarri:
My personal approach is to take by reference the arguments that are going to be processed in some way, and by value the ones that are going to be used verbatim (e.g. saved in a struct). So for this method I would vote for passing the argument by reference. 

But in the boxed case, while the value of the integer is not used verbatim, the heap memory where it's stored is. I can see the value of that to avoid an allocation. Would it be useful to have two `new()` methods - one that takes a reference and one that takes a value? `new(&)` and `new_consuming()` or something?

dvdplm:
I quite like having two separate initializers, reflecting more explicitly what is going on under the hood.

tarcieri:
Two initializers where one just calls `clone()` for you? I think it would be simpler to have one and let the caller `clone` if need be, which I guess is the current state of affairs.

(When I made my original comment about changing this to a reference, I guess I forgot that `BoxedMontyForm::new` doesn't borrow and tries to reuse the memory to avoid an allocation)

src/modular/boxed_monty_form.rs
tarcieri: Aah yeah, that's a good point

dvdplm:
Fair enough! Thank you for taking the time to hash it out. :)

PR #602, opened 2024-06-03
Add a const_new method to NonZero

dvdplm:
Makes it so users can have `NonZero` constants.

### DIFF
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -116,7 +116,7 @@ impl<const LIMBS: usize> NonZero<Uint<LIMBS>> {
     ///
     /// In future versions of Rust it should be possible to replace this with
     /// `NonZero::new(…).unwrap()`
-    // TODO: Remove when `Result::unwrap` is const (https://github.com/rust-lang/rust/issues/82814)
+    // TODO: Remove when `Self::new` and `CtOption::unwrap` support `const fn`
     pub const fn new_unwrap(n: Uint<LIMBS>) -> Self {
         if n.is_nonzero().is_true_vartime() {
             Self(n)
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -35,11 +35,6 @@ impl<T> NonZero<T> {
         CtOption::new(Self(n), !is_zero)
     }
 
-    /// Create a new non-zero integer without zero-checks.
-    pub const fn new_unchecked(n: T) -> Self {
-        Self(n)
-    }
-
     /// Provides access to the contents of `NonZero` in a `const` context.
     pub const fn as_ref(&self) -> &T {
         &self.0
@@ -116,6 +111,18 @@ impl NonZero<Limb> {
 }
 
 impl<const LIMBS: usize> NonZero<Uint<LIMBS>> {
+    /// Creates a new non-zero integer in a const context.
+    /// # Panics
+    ///
+    /// Panics if the value is zero.
+    pub const fn const_new(n: Uint<LIMBS>) -> Self {
+        if n.is_nonzero().is_true_vartime() {
+            Self(n)
+        } else {
+            panic!("Invalid value: zero")
+        }
+    }
+
     /// Create a [`NonZero<Uint>`] from a [`NonZeroU8`] (const-friendly)
     // TODO(tarcieri): replace with `const impl From<NonZeroU8>` when stable
     pub const fn from_u8(n: NonZeroU8) -> Self {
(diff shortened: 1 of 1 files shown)


dvdplm:
There used to be a `const_new` for this purpose but it relied on the custom `CtOption` machinery that is gone now (for good reason!). Maybe this can be a stop-gap solution until we get to a better place with const support?

tarcieri:
The constructor is also what's checking the `T: Zero` bound. This approach with absolutely no checks allows you to make `NonZero<String>` and all sorts of other nonsense type combinations.

Perhaps instead until `const impl` is stable it could be defined directly on concrete types like e.g. `NonZero<Uint<LIMBS>>` which would also enable checking for zero and panicking, rather than allowing for construction of a numerically invalid value.

dvdplm:
> The constructor is also what's checking the T: Zero bound. This approach with absolutely no checks allows you to make NonZero<String> and all sorts of other nonsense type combinations.


I can add the `Zero` bound of course, but that's a bit of a "false safety" no?

dvdplm:
> Perhaps instead until `const impl` is stable it could be defined directly on concrete types like e.g. `NonZero<Uint<LIMBS>>` which would also enable checking for zero and panicking, rather than allowing for construction of a numerically invalid value.

You mean bring back a variant of what 0.5.5 had?

```rust
impl<const LIMBS: usize> NonZero<Uint<LIMBS>> {
    /// Creates a new non-zero integer in a const context.
    /// The second return value is `FALSE` if `n` is zero, `TRUE` otherwise.
    pub const fn const_new(n: Uint<LIMBS>) -> (Self, CtChoice) {
        (Self(n), n.ct_is_nonzero())
    }
}
```

…but without `CtChoice`?

tarcieri:
Yes, I forget offhand why we got rid of that, but for defining constants a `const` constructor that panics is probably preferable until e.g. `.unwrap()`/`.expect()` can work in const contexts

dvdplm:
I found a workaround for my concrete issue using `to_nz()` and `expect` (both `const`):

```
trait Beef {
    const MOO: NonZero<Uint<16>>
}

struct MyBeef;
impl Beef for MyBeef {
    const MOO: U512::from_be_hex("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141")
        .to_nz()
        .expect("Valid by construction");
}
```

I had a stab at making a panicking `const_new`, but ran into issues. I'll have another go later.

dvdplm:
@tarcieri PTAL at https://github.com/RustCrypto/crypto-bigint/pull/602/commits/cf6486715602000b50de25e3a99e696dc7496194 – is that what you had in mind here? :)

tarcieri:
Sure, that works. You could also implement it with `.to_nz().expect` as per above.

dvdplm:
Is there anything left to do here before merging?

src/non_zero.rs
```diff
@@ -111,6 +111,18 @@ impl NonZero<Limb> {
 }
 
 impl<const LIMBS: usize> NonZero<Uint<LIMBS>> {
+    /// Creates a new non-zero integer in a const context.
+    /// # Panics
+    ///
+    /// Panics if the value is zero.
+    pub const fn const_new(n: Uint<LIMBS>) -> Self {
```
tarcieri, line 118: In `const-oid` we call a similar function `new_unwrap`: https://docs.rs/const-oid/latest/const_oid/struct.ObjectIdentifier.html#method.new_unwrap

dvdplm: Renamed in [7f6bbb1](https://github.com/RustCrypto/crypto-bigint/pull/602/commits/7f6bbb12fcef27464186b57cd08fd19358af36c8)

src/non_zero.rs
```diff
@@ -111,6 +111,20 @@ impl NonZero<Limb> {
 }
 
 impl<const LIMBS: usize> NonZero<Uint<LIMBS>> {
+    /// Creates a new non-zero integer in a const context.
+    /// Panics if the value is zero.
+    ///
+    /// In future versions of Rust it should be possible to replace this with
+    /// `NonZero::new(…).unwrap()`
+    // TODO: Remove when `Result::unwrap` is const (https://github.com/rust-lang/rust/issues/82814)
```
tarcieri, line 119:
```suggestion
    // TODO: Remove when `Self::new` and `CtOption::unwrap` support `const fn`
```

tarcieri (approved):

PR #603, opened 2024-06-03
Implement `from_str_radix_vartime` for `Uint`, `BoxedUint`

andrewwhitehead:
This adds fallible parsing for binary, decimal, and hex integers (and any base from 2 to 36).

The format of these numbers generally matches the `from_str_radix` methods of the standard library. One exception is that underscores may used as spacers, as in `+1_000_000` for example. This is consistent with the `num_bigint` crate.

The `DecodeError` enum is promoted from the uint.boxed module and reused.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -9,7 +9,7 @@ pub use num_traits::{
 pub(crate) use sealed::PrecomputeInverterWithAdjuster;
 
 use crate::{Limb, NonZero, Odd, Reciprocal};
-use core::fmt::Debug;
+use core::fmt::{self, Debug};
 use core::ops::{
     Add, AddAssign, BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Div, DivAssign,
     Mul, MulAssign, Neg, Not, Rem, Shl, ShlAssign, Shr, ShrAssign, Sub, SubAssign,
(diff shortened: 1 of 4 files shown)


ndavd:
Is this already fully working or still a WIP?

andrewwhitehead:
It's fully working. I have encoding done too (to_str_radix_vartime), but I didn't want to submit too many changes at once.

ndavd:
That will be a must, was about to start working on a solution for it locally but it's great to know this is being added 🙏🏻

ndavd:
When do you estimate both from and to methods will be available?

tarcieri:
@ndavd I have a lot on my plate right now and higher priority issues to work on, even in this crate. It will also be going into a prerelease, and we likely won't have a final release out for several months.

src/traits.rs
```diff
@@ -541,6 +538,41 @@ pub trait Encoding: Sized {
...
+            Self::Precision => write!(
+                f,
+                "the deserialized number is larger than the given precision"
+            ),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for DecodeError {}
```
tarcieri, line 574: Seems like an odd place for this. Perhaps we should extract an `error.rs`. But I'll save that for another PR.

tarcieri (approved):

PR #604, opened 2024-06-04
Implement serde for `Odd`

dvdplm:
Most other wrapper types implement serde, e.g. `NonZero` and `Wrapping`. This PR implements it for `Odd` as well.

One oddity I noticed when writing the tests was that deserializing a bigger `Uint` into a smaller one does not return an error. Instead the original value seems to be truncated to fit the smaller type. Is that the expected behaviour?

### DIFF
--- a/src/odd.rs
+++ b/src/odd.rs
@@ -268,7 +268,8 @@ mod tests {
         // @reviewers: I expected an error when deserializing an Odd<U128> into an Odd<U64>, instead I get a truncated number. Is this a big or known limitation?
         #[test]
         fn silently_coerces_bigger_type_into_smaller_type() {
-            let three = Odd::new(U128::from_u128(0x33333333333333333333333333333333)).unwrap();
+            let three = Odd::new(U128::from_u128(0x77777777777777773333333333333333)).unwrap();
+
             let three_ser = bincode::serialize(&three).unwrap();
 
             // This doesn't fail, which is unexpected
--- a/src/odd.rs
+++ b/src/odd.rs
@@ -171,15 +171,10 @@ impl<'de, T: Deserialize<'de> + Integer + Zero> Deserialize<'de> for Odd<T> {
 
         if bool::from(value.is_odd()) {
             Ok(Self(value))
-        } else if bool::from(value.is_zero()) {
-            Err(D::Error::invalid_value(
-                Unexpected::Other("zero"),
-                &"a non-zero odd value",
-            ))
         } else {
             Err(D::Error::invalid_value(
                 Unexpected::Other("even"),
-                &"an odd value",
+                &"a non-zero odd value",
             ))
         }
     }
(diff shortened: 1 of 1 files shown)


src/odd.rs
```diff
@@ -153,6 +161,40 @@ impl Odd<BoxedUint> {
...
+            Err(D::Error::invalid_value(
+                Unexpected::Other("zero"),
+                &"a non-zero odd value",
+            ))
+        } else {
+            Err(D::Error::invalid_value(
+                Unexpected::Other("even"),
+                &"an odd value",
+            ))
+        }
```
tarcieri, line 184: Seems like this is largely duplicating `Odd::new`. And why are you checking for zero explicitly here? Zero is even and will fail that way.

dvdplm: `Odd::new`? I copied the `NonZero` deserialization code almost verbatim!

tarcieri: Mathematically it seems weird to special case zero here, IMO.

I guess you can claim it's more helpful, but when is it actually helpful to know the value is zero beyond simply being even?

tarcieri: > I copied the NonZero deserialization code almost verbatim!

Well, that code actually has a reason to worry about zero as a value specifically, and it would probably be good if that were updated to use `NonZero::new` as well.

dvdplm: I removed the superfluous error message. Not sure what you mean by `Odd::new` though. Are you saying I should change this PR to use the `pub fn new(n: T) -> CtOption<Self>` and map the case the `CtOption` is `None` to a Serde error?

tarcieri: It's six of one, half dozen of another, but yes, that would avoid repeating the logic to validate `Odd`'s invariant

src/odd.rs
```diff
@@ -168,9 +168,9 @@ impl<'de, T: Deserialize<'de> + Integer + Zero> Deserialize<'de> for Odd<T> {
         D: Deserializer<'de>,
     {
         let value: T = T::deserialize(deserializer)?;
-
-        if bool::from(value.is_odd()) {
-            Ok(Self(value))
+        let odd = Self::new(value);
```
tarcieri, line 171: If you do:

```suggestion
        let odd = Option::<Self>::from(Self::new(value));
```

...you can handle the rest using `Option`/`Result` combinators

tarcieri (approved):
Looks good now, thanks!

src/odd.rs
```diff
@@ -175,4 +207,68 @@ mod tests {
...
+            let three = Odd::new(U64::from_u64(0x3)).unwrap();
+            let three_ser = bincode::serialize(&three).unwrap();
+            let error_message = bincode::deserialize::<Odd<U128>>(&three_ser)
+                .unwrap_err()
+                .to_string();
+
+            assert_eq!(&error_message, "io error: unexpected end of file");
+        }
+
+        // @reviewers: I expected an error when deserializing an Odd<U128> into an Odd<U64>, instead I get a truncated number. Is this a big or known limitation?
```
dvdplm, line 258: @tarcieri Forgot about this. :/

Is it expected that deserializing bigger into smaller truncates this way?

tarcieri: That does seem like a possible bug. Can you file an issue?

PR #608, opened 2024-06-12
Faster vartime division for Uint

andrewwhitehead:
Implements faster vartime division (vartime with the divisor only) for Uint<L> based on Knuth's TAOCP volume 2, as outlined at https://janmr.com/blog/2014/04/basic-multiple-precision-long-division/

This does not address vartime division for BoxedUint or the other TODOs in https://github.com/RustCrypto/crypto-bigint/pull/511

Relevant benchmarks:

```
wrapping ops/div/rem_vartime, U256/U128, full size
                        time:   [84.091 ns 84.290 ns 84.525 ns]
                        change: [-85.312% -85.155% -84.896%] (p = 0.00 < 0.05)
                        Performance has improved.
wrapping ops/rem_vartime, U256/U128, full size
                        time:   [84.129 ns 84.474 ns 84.890 ns]
                        change: [-84.208% -84.057% -83.898%] (p = 0.00 < 0.05)
                        Performance has improved.
wrapping ops/rem_wide_vartime, U256
                        time:   [166.95 ns 167.27 ns 167.69 ns]
                        change: [-94.750% -94.705% -94.636%] (p = 0.00 < 0.05)
                        Performance has improved.
Dynamic Montgomery arithmetic/MontyParams::new_vartime
                        time:   [459.28 ns 460.48 ns 461.88 ns]
                        change: [-80.512% -80.452% -80.390%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -675,10 +675,8 @@ pub(crate) const fn div_rem_vartime<const N1: usize, const N2: usize>(
                 ((((!rx) & qy) | (((!rx) | qy) & (rx.wrapping_sub(qy)))) >> (WideWord::BITS - 1))
                     as Word,
             );
-            {
-                quo = quo.saturating_sub(diff.select_word(0, 1));
-                rem = rem.saturating_add(diff.select_word(0, y[yc - 1].0));
-            }
+            quo = quo.saturating_sub(diff.select_word(0, 1));
+            rem = rem.saturating_add(diff.select_word(0, y[yc - 1].0));
             i += 1;
         }
 
@@ -814,10 +812,8 @@ pub(crate) const fn rem_wide_vartime<const N: usize>(
                 ((((!rx) & qy) | (((!rx) | qy) & (rx.wrapping_sub(qy)))) >> (WideWord::BITS - 1))
                     as Word,
             );
-            {
-                quo = quo.saturating_sub(diff.select_word(0, 1));
-                rem = rem.saturating_add(diff.select_word(0, y[yc - 1].0));
-            }
+            quo = quo.saturating_sub(diff.select_word(0, 1));
+            rem = rem.saturating_add(diff.select_word(0, y[yc - 1].0));
             i += 1;
         }
 
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -86,7 +86,6 @@ impl<const LIMBS: usize> Uint<LIMBS> {
 
         // Short circuit for small or extra large divisors
         match yc {
-            0 => panic!("divisor must be non-zero"),
             1 => {
                 // If the divisor is a single limb, use limb division
                 let (q, r) = div_rem_limb_with_reciprocal(
@@ -241,19 +240,14 @@ impl<const LIMBS: usize> Uint<LIMBS> {
         let dbits = rhs.0.bits_vartime();
         let yc = ((dbits + Limb::BITS - 1) / Limb::BITS) as usize;
 
-        // Short circuit for small divisors
-        match yc {
-            0 => panic!("divisor must be non-zero"),
-            1 => {
-                // If the divisor is a single limb, use limb division
-                let r = rem_limb_with_reciprocal_wide(
-                    (&lower_upper.0, &lower_upper.1),
-                    &Reciprocal::new(rhs.0.limbs[0].to_nz().expect("zero divisor")),
-                );
-                return Uint::from_word(r.0);
-            }
-            _ => {}
-        };
+        // If the divisor is a single limb, use limb division
+        if yc == 1 {
+            let r = rem_limb_with_reciprocal_wide(
+                (&lower_upper.0, &lower_upper.1),
+                &Reciprocal::new(rhs.0.limbs[0].to_nz().expect("zero divisor")),
+            );
+            return Uint::from_word(r.0);
+        }
 
         let lshift = (Limb::BITS - (dbits % Limb::BITS)) % Limb::BITS;
         let rshift = if lshift == 0 { 0 } else { Limb::BITS - lshift };
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -1,7 +1,12 @@
 //! [`Uint`] division operations.
 
-use super::div_limb::{div_rem_limb_with_reciprocal, rem_limb_with_reciprocal, Reciprocal};
-use crate::{CheckedDiv, ConstChoice, DivRemLimb, Limb, NonZero, RemLimb, Uint, Word, Wrapping};
+use super::div_limb::{
+    div2by1, div_rem_limb_with_reciprocal, rem_limb_with_reciprocal, rem_limb_with_reciprocal_wide,
+    Reciprocal,
+};
+use crate::{
+    CheckedDiv, ConstChoice, DivRemLimb, Limb, NonZero, RemLimb, Uint, WideWord, Word, Wrapping,
+};
 use core::ops::{Div, DivAssign, Rem, RemAssign};
 use subtle::CtOption;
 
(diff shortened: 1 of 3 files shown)


andrewwhitehead:
It looks like `Uint::mul_mod` would also be about 70% faster using this update and `split_mut`/`rem_wide_vartime`. The current implementation uses `MontyParams::new_vartime` so it seems like it is vartime for the modulus, but maybe that should be renamed `mul_mod_vartime` anyway? Accepting a `&NonZero<Uint>` for the modulus would also simplify things.

src/uint/div.rs
```diff
@@ -69,29 +74,8 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
-            if bd == 0 {
-                break;
-            }
-            bd -= 1;
-            c = c.shr1();
-            quo = quo.shl1();
-        }
-
-        (quo, rem)
+        div_rem_vartime(self, &rhs.0)
```
tarcieri, line 78: Is there a particular reason you need a separate free function for `div_rem_vartime`, as opposed to placing the implementation here? They appear to have nearly the same type signature aside from supporting different lengths (which could also be changed here)

andrewwhitehead: I think just that the implementation could support a `div_rem_vartime_mixed` potentially. The signature for `Uint::div_rem_vartime` can't be made more flexible without creating type inference issues.

src/uint/div.rs
```diff
@@ -660,6 +602,283 @@ impl<const LIMBS: usize> RemLimb for Uint<LIMBS> {
...
+        // If the subtraction borrowed, then decrement q and add back the divisor
+        let ct_borrow = ConstChoice::from_word_mask(borrow.0);
+        carry = Limb::ZERO;
+        i = 0;
+        while i < yc {
+            (x[xi + i + 1 - yc], carry) =
+                x[xi + i + 1 - yc].adc(Limb::select(Limb::ZERO, y[i], ct_borrow), carry);
+            i += 1;
+        }
+        quo -= ct_borrow.select_word(0, 1);
```
andrewwhitehead, line 704: I can't actually get this to trigger in practice, removing it seems to have no effect as `quo` is always correct before this. It would be nice to prove that (mathematically).

src/uint/div.rs
tarcieri: > can't be made more flexible without creating type inference issues.

Really? Like what?

Most of the methods in `mul.rs` have a `const RHS_LIMBS: usize` without problems.

andrewwhitehead: Actually it's `rem_vartime` that creates a minor type inference issue in the montgomery params, but that's easily fixed: https://github.com/RustCrypto/crypto-bigint/blob/master/src/modular/monty_form.rs#L78-L80

andrewwhitehead: I moved the implementations inline and adjusted the signature of the vartime division methods, leaving the signature of `rem_vartime` alone for now.

tarcieri:
> The current implementation uses MontyParams::new_vartime so it seems like it is vartime for the modulus, but maybe that should be renamed mul_mod_vartime anyway? 

That's a good point... perhaps there should be both `Uint::mul_mod` as well as `Uint::mul_mod_vartime` with the former using `MontyParams::new`? (looks like it would need some extra bounds)

tarcieri:
@fjarri thoughts on this versus #511?

fjarri:
This is definitely much faster than what I have there, and as for the naming stuff, I'll make a separate PR.

src/uint/div.rs
```diff
@@ -69,29 +74,148 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+    ) -> (Self, Uint<RHS_LIMBS>) {
+        // Based on Section 4.3.1, of The Art of Computer Programming, Volume 2, by Donald E. Knuth.
+        // Further explanation at https://janmr.com/blog/2014/04/basic-multiple-precision-long-division/
+
+        let dbits = rhs.0.bits_vartime();
+        let yc = ((dbits + Limb::BITS - 1) / Limb::BITS) as usize;
+
+        // Short circuit for small or extra large divisors
+        match yc {
+            0 => panic!("divisor must be non-zero"),
```
fjarri, line 89: Can be removed, the divisor is already ensured to be non-zero.

src/uint/div.rs
```diff
@@ -104,22 +228,7 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
-            rem = Self::select(&r, &rem, ConstChoice::from_word_mask(borrow.0));
-            if bd == 0 {
-                break;
-            }
-            bd -= 1;
-            c = c.shr1();
-        }
-
-        rem
+        self.div_rem_vartime(rhs).1
```
fjarri, line 231: I wonder if manually modifying `div_rem_vartime()` with the assumption that the quotient is discarded produces a faster result.

src/uint/div.rs
```diff
@@ -129,32 +238,133 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
-        let mb = rhs.0.bits_vartime();
-
-        // The number of bits to consider is two sets of limbs * BITS - mb (modulus bitcount)
-        let mut bd = (2 * Self::BITS) - mb;
+        let dbits = rhs.0.bits_vartime();
+        let yc = ((dbits + Limb::BITS - 1) / Limb::BITS) as usize;
+
+        // Short circuit for small divisors
+        match yc {
+            0 => panic!("divisor must be non-zero"),
```
fjarri, line 246: Same as above, this will never be reached.

src/uint/div.rs
```diff
@@ -129,32 +238,133 @@ impl<const LIMBS: usize> Uint<LIMBS> {
     /// When used with a fixed `rhs`, this function is constant-time with respect
     /// to `self`.
     pub const fn rem_wide_vartime(lower_upper: (Self, Self), rhs: &NonZero<Self>) -> Self {
```
fjarri, line 240: Could this be reduced to `div_rem_vartime()` by widening `rhs` to double width, and then halving the result? So that you don't have to spell out the whole algorithm again.

andrewwhitehead: I don't think that can be done without extra trait bounds or const evaluation support.

src/uint/div.rs
andrewwhitehead: I don't believe there would be much benefit, it would just avoid storing the quotient words and then shifting them at the end. I think I also tried an inlined version to see if the optimizer would catch anything and it didn't seem to have any impact.

tarcieri:
This splits `Uint::mul_mod` and `Uint::mul_mod_vartime`: https://github.com/RustCrypto/crypto-bigint/pull/623

andrewwhitehead:
I also have vartime division for BoxedUint implemented, which I was planning to add in a separate PR.

tarcieri (approved):

fjarri:
Just a note: I experimented a little with trying to piggyback `rem_wide_vartime()` on `div_rem_vartime()`, and the following kinda works:
```rust
    pub const fn rem_wide_vartime<const R: usize>(lower_upper: (Self, Self), rhs: &NonZero<Self>) -> Self
    where
        Self: Concat<Output=Uint<R>>,
        Uint<R>: Split<Output=Self>,
    {
        let (lo, hi) = lower_upper;
        let wide_self: <Self as Concat>::Output = lo.concat(&hi);
        let (_q, r) = wide_self.div_rem_vartime(rhs);
        r
    }
```
The problem is, it then requires an `<R>` generic parameter in some `Monty` trait methods, which doesn't make sense when they're implemented for `BoxedUint`. So yeah, we'll have to wait until const traits are a thing so that we can say "`Concat::Output` is an `Uint` (or `Integer`)" without spelling out the concrete type.

PR #617, opened 2024-07-11
Add support for computing `Uint::gcd` with even modulus

tarcieri:
Uses a similar method to `Uint::inv_mod` for extending `Uint::gcd` with support for an even modulus, first dividing out `2^k` from both operands where this operation will ensure at least one value is odd, and then applying Bernstein-Yang to the odd modulus, multiplying the result by `2^k`.

### DIFF
--- a/src/uint/gcd.rs
+++ b/src/uint/gcd.rs
@@ -1,6 +1,8 @@
-//! Support for computing greatest common divisor of two `Uint`s.
+//! Support for computing the greatest common divisor of two `Uint`s.
 
-use crate::{modular::BernsteinYangInverter, ConstCtOption, Gcd, Odd, PrecomputeInverter, Uint};
+use crate::{
+    modular::BernsteinYangInverter, ConstChoice, ConstCtOption, Gcd, Odd, PrecomputeInverter, Uint,
+};
 use subtle::CtOption;
 
 impl<const SAT_LIMBS: usize, const UNSAT_LIMBS: usize> Uint<SAT_LIMBS>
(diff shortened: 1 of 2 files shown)


conradludgate (approved):

tarcieri:
Would appreciate any retroactive review, if anyone would like to offer any.

That said, this seems like a reasonable method so long as the underlying primitives are constant time, and it passes proptests.

src/uint/gcd.rs
```diff
@@ -1,18 +1,34 @@
...
-        let ret = <Odd<Self> as PrecomputeInverter>::Inverter::gcd(self, rhs);
-        ConstCtOption::new(ret, self.is_odd())
+        let k1 = self.trailing_zeros();
+        let k2 = rhs.trailing_zeros();
+
+        // Select the smaller of the two `k` values, making 2^k the common even divisor
+        let k = ConstChoice::from_u32_lt(k2, k1).select_u32(k1, k2);
+
+        // Decompose `self` and `rhs` into `s{1, 2} * 2^k` where either `s1` or `s2` is odd
+        let s1 = self.overflowing_shr(k).unwrap_or(Self::ZERO);
```
fjarri, line 21: Probably `wrapping_shr()` would look a little better here (even though it does the same thing), since `k` is always less than `BITS`.

fjarri: Ah actually it is not, it will be `BITS` if `self` and `rhs` are zero.

fjarri:
Looks good to me.

PR #618, opened 2024-07-13
Add support for computing `BoxedUint::gcd` with even modulus

tarcieri:
Edit: it works now! Uses an implementation similar to `Uint::gcd` as added in #617

~~Tests are currently failing due to the handling of zero as one of the values.~~

~~I'm thinking the return value of `CtOption` is largely a vestige of when even numbers weren't supported, and we can change it to be infallible now, similar to `num-bigint`.~~

~~I'm uncertain why we aren't seeing similar failures in the `Uint::gcd` proptests as I expect it to have a similar issue.~~

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -197,12 +197,12 @@ pub trait FixedInteger: Bounded + ConditionallySelectable + Constants + Copy + I
     const LIMBS: usize;
 }
 
-/// Compute greatest common divisor of two integers.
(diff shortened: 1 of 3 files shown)


dignifiedquire:
agreed on removing ctoption 👍

tarcieri:
Opened #619 to make `Uint::gcd` infallible, then when that's merged I'll do a similar treatment here

PR #623, opened 2024-07-27
Split `Uint::mul_mod` and `Uint::mul_mod_vartime`

tarcieri:
The previous `mul_mod` was variable-time with respect to the modulus. This adds a proper constant-time version.

It also changes the `p` argument to be `NonZero<Self>`

### DIFF
--- a/src/uint/mul_mod.rs
+++ b/src/uint/mul_mod.rs
@@ -3,28 +3,40 @@
 use crate::{
     modular::{MontyForm, MontyParams},
     primitives::mul_rem,
-    Limb, MulMod, Uint, WideWord, Word,
+    Concat, Limb, MulMod, Odd, Split, Uint, WideWord, Word,
 };
 
 impl<const LIMBS: usize> Uint<LIMBS> {
     /// Computes `self * rhs mod p` for odd `p`.
-    ///
-    /// Panics if `p` is even.
-    // TODO(tarcieri): support for even `p`?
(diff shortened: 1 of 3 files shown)


src/uint/mul_mod.rs
```diff
@@ -64,7 +76,7 @@ impl<const LIMBS: usize> MulMod for Uint<LIMBS> {
     type Output = Self;
 
     fn mul_mod(&self, rhs: &Self, p: &Self) -> Self {
-        self.mul_mod(rhs, p)
+        self.mul_mod_vartime(rhs, &p.to_odd().expect("only odd moduli supported"))
     }
```
tarcieri, line 80: Perhaps the `MulMod` trait should be changed to accept `&Odd<Self>` for `p`, so that the inherent method matches the trait.

(I do find a type safe `Odd` preferable to panicking here)

andrewwhitehead:
An updated version of `mul_mod_vartime` would not require an odd modulus, but I think requiring a NonZero modulus would make sense.

src/uint/mul_mod.rs
```diff
@@ -64,7 +80,7 @@ impl<const LIMBS: usize> MulMod for Uint<LIMBS> {
     type Output = Self;
 
     fn mul_mod(&self, rhs: &Self, p: &Self) -> Self {
-        self.mul_mod(rhs, p)
+        self.mul_mod_vartime(rhs, &NonZero::new(*p).expect("p should be non-zero"))
```
tarcieri, line 83: This should probably be `mul_mod` but I would need to add the extra bounds

PR #625, opened 2024-08-01
Make `mul_rem()` primitive constant-time

fjarri:
As mentioned in #511, `mul_rem()` is currently variable-time in `d`, which makes `mul_mod_special()` variable-time in `c`. This PR makes `mul_rem()` constant-time with no loss of performance. 

Since `mul_mod` now uses the `div_limb` functionality, it was moved there.

Also added `NonZero::<Limb>::new_unwrap()`.

### DIFF
--- a/src/primitives.rs
+++ b/src/primitives.rs
@@ -65,9 +65,3 @@ pub(crate) const fn mac(a: Word, b: Word, c: Word, carry: Word) -> (Word, Word) 
     let ret = a + (b * c) + carry;
     (ret as Word, (ret >> Word::BITS) as Word)
 }
-
-/// Computes `(a * b) % d`.
-#[inline(always)]
-pub(crate) const fn mul_rem(a: Word, b: Word, d: Word) -> Word {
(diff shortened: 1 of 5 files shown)


tarcieri:
> Whatever happened to a const constructor of `NonZero`? 

#602

src/primitives.rs
```diff
@@ -65,9 +65,3 @@ pub(crate) const fn mac(a: Word, b: Word, c: Word, carry: Word) -> (Word, Word)
     let ret = a + (b * c) + carry;
     (ret as Word, (ret >> Word::BITS) as Word)
 }
-
-/// Computes `(a * b) % d`.
-#[inline(always)]
-pub(crate) const fn mul_rem(a: Word, b: Word, d: Word) -> Word {
-    ((a as WideWord * b as WideWord) % (d as WideWord)) as Word
```
tarcieri, line 72: Perhaps we should enable the `clippy::integer_division_remainder_used` lint

tarcieri (approved):

Issue #627, opened 2024-08-02
Bernstein-Yang: constant-time issues


tarcieri:
This is a tracking issue for various problems with our Bernstein-Yang implementation where it operates in a non-constant-time manner:

- [x] `Int62L::eq` returns `bool` that's branched on in various places ([#630](https://github.com/RustCrypto/crypto-bigint/pull/630), [#631](https://github.com/RustCrypto/crypto-bigint/pull/631))
- [x] `Int62L::is_negative` returns `bool` that's branched on in various places ([#628](https://github.com/RustCrypto/crypto-bigint/pull/628), [#629](https://github.com/RustCrypto/crypto-bigint/pull/628))
- [x] `gcd` uses a data-dependent loop ([#632](https://github.com/RustCrypto/crypto-bigint/pull/632), [#635](https://github.com/RustCrypto/crypto-bigint/pull/635))

Note: as the implementation has never received a stable release, we don't consider this issue security critical.

tarcieri:
All of the timing variability I'm aware of has been removed. Closing this.

PR #628, opened 2024-08-02
Uint: constant-time Bernstein-Yang `is_negative`

tarcieri:
Changes `Int64L::is_negative` to return `ConstChoice` rather than `bool`, also adding a `Int64L::select` method for selecting between two values predicated on a `ConstChoice`.

This eliminates some of the branching in the implementation (#627).

cc @erik-3milabs

### DIFF
--- a/src/const_choice.rs
+++ b/src/const_choice.rs
@@ -22,10 +22,15 @@ impl ConstChoice {
     }
 
     #[inline]
-    #[allow(trivial_numeric_casts)]
-    #[allow(clippy::unnecessary_cast)]
+    #[cfg(target_pointer_width = "32")]
     pub(crate) const fn as_u64_mask(&self) -> u64 {
-        self.0 as u64
+        ((self.0 as u64) << 32) | (self.0 as u64)
+    }
+
+    #[inline]
+    #[cfg(target_pointer_width = "64")]
+    pub(crate) const fn as_u64_mask(&self) -> u64 {
+        self.0
     }
 
     /// Returns the truthy value if `value == Word::MAX`, and the falsy value if `value == 0`.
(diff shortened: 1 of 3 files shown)


tarcieri:
Hmm, this appears to potentially be broken on 32-bit platforms. Will try to reproduce locally.

src/const_choice.rs
```diff
@@ -154,6 +193,12 @@ impl ConstChoice {
     pub(crate) const fn to_u8(self) -> u8 {
         (self.0 as u8) & 1
     }
+
+    /// WARNING: this method should only be used in contexts that aren't constant-time critical!
+    #[inline]
+    pub(crate) const fn to_bool(self) -> bool {
```
fjarri, line 199: Maybe it's worth suffixing with `vartime`

tarcieri: Done in [47c4db3](https://github.com/RustCrypto/crypto-bigint/pull/628/commits/47c4db32624b52933dafa080dc585c76e5deb941)

src/modular/bernstein_yang.rs
```diff
@@ -86,7 +86,7 @@ impl<const SAT_LIMBS: usize, const UNSAT_LIMBS: usize>
         // integer to be inverted and the modulus the inverter was created for.
         // Thus, if "f" is neither 1 nor -1, then the sought inverse does not exist.
         let antiunit = f.eq(&Int62L::MINUS_ONE);
-        let ret = self.norm(d, antiunit);
+        let ret = self.norm(d, ConstChoice::from_word_lsb(antiunit as Word));
```
tarcieri, line 89: It would be nice to make `eq` return `ConstChoice` here but I'm saving that as a followup

PR #639, opened 2024-08-05
Inversion tests

dvdplm:
As a follow-up to #606, this PR extends the set of tests for modular inversion.


tarcieri:
It looks like the tests are breaking on 32-bit platforms because you're hardcoding the number of limbs

tarcieri:
@dvdplm can you remove the tests that overlap with proptests, or convert them to proptests where the existing proptests are lacking?

dvdplm:
Right, this has been lingering, my bad.

dvdplm:
@tarcieri I have converted the relevant parts of my tests to proptests. I have not used proptest before so feedback is welcome. The most annoying part is how it can't work with generics, leading to very verbose code here (had to copy paste the test for each uint size).
The basic idea is to test inversion over random edge cases, where I tweak the boundary bytes randomly (and also the midpoint for good measure).

tarcieri:
@dvdplm rebasing should fix the security audit

tarcieri (approved):
Thanks for bearing with us on this one!

Issue #312, opened 2023-11-26
Determine and document `BoxedUint` widening arithmetic semantics


tarcieri:
Arithmetic with `BoxedUint` needs to deal with values whose number of limbs don't match. This is true of `Uint` as well, but there everything is nicely type safe.

So far the internal `BoxedUint::chain` operation (which addition and subtraction are implemented in terms of) handles widening to the widest of the two inputs.

The `wrapping_*` and `checked_*` functions use the `self`/left-hand side argument as the overflow width.

The `mul_wide` operation widens to as many limbs as are in the sum of the number of limbs in the operands (though should perhaps be renamed to just `BoxedUint::mul` for consistency with `Uint::mul`, which is also a widening multiply.

This seems like a reasonable enough approach, but whatever one we take should be thoroughly documented, both as an overall policy and on the individual functions.

We should also ensure it's consistent with how `Uint` performs widening, where applicable.

tarcieri:
I should probably add some context that these semantics are necessarily different from your typical `num-bigint`-style arbitrary precision library.

Notably in cryptographic cases we want all of the values to be the same precision, or if not that, then known/fixed precisions at every point in the program. In the case of an algorithm like RSA, that would be the number of bits of precision in the RSA modulus. This means preserving leading zero limbs in such cases, so that we don't wind up with value-dependent timing variability with respect to the number of limbs.

xuganyu96:
Automatic widening when instantiating `BoxedMontyForm` should also be considered here. `BoxedMontyForm::new` currently panics if `integer` and `params` have different precision, but I think it makes sense to automatically widen the integer if it is less precise than the params. I am not sure if we should consider widening the params since it has more moving parts.

kayabaNerve:
Probably worth an issue, yet I observed overflows on `BoxedUint::mul` which is documented to inherently widen. I don't see an existing issue other than this one and wanted to clarify if that's considered a bug I should report.

tarcieri:
@kayabaNerve please open an issue, ideally with a repro of the problem

PR #642, opened 2024-08-06
Support `BoxedUint::gcd(_vartime)` where `self` and `rhs` have a different number of limbs

fjarri:
The PR currently has just a testcase illustrating the problem. 

So the question is, what should the behavior be? We can equalize the sizes of the arguments at the top level, but that would make `gcd()` not constant time wrt numbers of limbs. Then again, would users assume that it is?

### DIFF
--- a/src/modular/bernstein_yang/boxed.rs
+++ b/src/modular/bernstein_yang/boxed.rs
@@ -68,6 +68,11 @@ impl Inverter for BoxedBernsteinYangInverter {
 
 /// Returns the greatest common divisor (GCD) of the two given numbers.
 pub(crate) fn gcd(f: &BoxedUint, g: &BoxedUint) -> BoxedUint {
+    // TODO (#312): depending on the decision there, this may be relaxed.
+    if f.nlimbs() != g.nlimbs() {
+        panic!("Both arguments to `gcd()` must have the same number of limbs");
+    }
+
     let bits_precision = f.bits_precision();
     let inverse = inv_mod2_62(f.as_words());
     let f = BoxedUnsatInt::from(f);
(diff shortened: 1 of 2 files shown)


tarcieri:
It's part of the general implicit widening problem: #312. Implicit widening may make sense.

In many places we currently just panic rather than performing implicit widening, relying on the caller to widen appropriately.

tarcieri:
I can try to add implicit widening in a separate PR, unless you want to add it in this one. I think it's appropriate here.

fjarri:
That's fine, it can wait until the resolution on #312. It's just that the error caught me off-guard.

tarcieri:
`BoxedUnsatInt` needs to allocate anyway, so we can implicitly widen there while also converting to `u62`-ish limbs

fjarri:
We need to equalize the number of limbs earlier, when making sure the first argument is odd and conditionally swapping them.

tarcieri:
@fjarri I was suggesting finding the max of the two limb counts, similar to how it finds the max of two bits, then pass a limb count hint in when constructing `BoxedUnsatInt`.

That would be as opposed to calling `BoxedUint::widen` which would add another allocation.

fjarri:
But the difference in limb numbers will already lead to panic before that, [in `BoxedUint::gcd()`](https://github.com/RustCrypto/crypto-bigint/blob/master/src/uint/boxed/gcd.rs#L22)

tarcieri:
Aah! Yes, that is a problem, though there are several paths through Bernstein-Yang that avoid that.

I'm surprised there isn't a better length `assert!` for that function.

fjarri:
Actually, it won't even lead to panic if the first argument is longer, it'll just behave incorrectly.

tarcieri:
Will merge this to have a set of test cases for widening behavior, and then I'll take a look at actually implementing it. It shouldn't be too hard.

PR #643, opened 2024-08-07
Faster constant-time division

andrewwhitehead:
This replaces the `div_rem` method for Uint and BoxedUint with one based on the updated vartime division method.

Unlike the old method, this one panics if `NonZero(0)` is given as the divisor. The `sqrt` method relied on the old behavior and is updated.

The `Default` implementation of `NonZero` was also producing invalid `NonZero(0)` values, which affected the `NonZero::map` usage in `checked_div`. The trait implementation is updated to return `Self::ONE` as the default.

Some relevant benchmarks:

```
wrapping ops/div/rem, U256/U128, full size
                        time:   [312.41 ns 314.29 ns 317.18 ns]
                        change: [-81.412% -79.729% -78.416%] (p = 0.00 < 0.05)
                        Performance has improved.
wrapping ops/rem, U256/U128, full size
                        time:   [310.37 ns 312.08 ns 315.15 ns]
                        change: [-78.497% -78.170% -77.896%] (p = 0.00 < 0.05)
                        Performance has improved.
sqrt/sqrt, U256         time:   [3.1401 µs 3.1769 µs 3.2335 µs]
                        change: [-78.636% -78.245% -77.823%] (p = 0.00 < 0.05)
                        Performance has improved.
wrapping ops/boxed_div_rem
                        time:   [7.6462 µs 7.6800 µs 7.7228 µs]
                        change: [-99.614% -99.609% -99.605%] (p = 0.00 < 0.05)
                        Performance has improved.
wrapping ops/boxed_rem  time:   [7.6416 µs 7.7808 µs 7.9917 µs]
                        change: [-99.615% -99.608% -99.601%] (p = 0.00 < 0.05)
                        Performance has improved.
boxed_sqrt/boxed_sqrt, 4096
                        time:   [110.47 µs 111.58 µs 113.16 µs]
                        change: [-99.601% -99.593% -99.586%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -21,7 +21,7 @@ use serdect::serde::{
 };
 
 /// Wrapper type for non-zero integers.
-#[derive(Clone, Copy, Debug, Default, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
 #[repr(transparent)]
 pub struct NonZero<T>(pub(crate) T);
 
@@ -207,6 +207,15 @@ where
 {
     fn ct_eq(&self, other: &Self) -> Choice {
         self.0.ct_eq(&other.0)
+    }
+}
+
+impl<T> Default for NonZero<T>
+where
+    T: Constants,
+{
+    fn default() -> Self {
+        Self(T::ONE)
     }
 }
 
--- a/src/uint/boxed/div.rs
+++ b/src/uint/boxed/div.rs
@@ -6,7 +6,7 @@ use crate::{
     RemLimb, Wrapping,
 };
 use core::ops::{Div, DivAssign, Rem, RemAssign};
-use subtle::{Choice, ConstantTimeLess, CtOption};
+use subtle::CtOption;
 
 impl BoxedUint {
     /// Computes `self / rhs` using a pre-made reciprocal,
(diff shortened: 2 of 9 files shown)


src/non_zero.rs
```diff
@@ -210,6 +210,15 @@ where
     }
 }
 
+impl<T> Default for NonZero<T>
```
fjarri, line 213: I'm not sure about this one. Seems a little counter-intuitive. Is it necessary?

src/uint/boxed/div.rs
```diff
@@ -118,41 +118,113 @@ impl BoxedUint {
...
+            return (quo, rem);
+        }
+
+        let dbits = rhs.bits();
+        assert!(dbits > 0, "zero divisor");
+        let dwords = (dbits + Limb::BITS - 1) / Limb::BITS;
+        let lshift = (Limb::BITS - (dbits % Limb::BITS)) % Limb::BITS;
+
+        // Shift entire divisor such that the high bit is set
+        let mut y = rhs.shl((size as u32) * Limb::BITS - dbits).to_limbs();
```
fjarri, line 156: Is the argument here not the same as `lshift`?

src/uint/div.rs
```diff
@@ -37,33 +38,96 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+            return (quo, rem);
+        }
+
+        let dbits = rhs.0.bits();
+        assert!(dbits > 0, "zero divisor");
+        let dwords = (dbits + Limb::BITS - 1) / Limb::BITS;
+        let lshift = (Limb::BITS - (dbits % Limb::BITS)) % Limb::BITS;
+
+        // Shift entire divisor such that the high bit is set
+        let mut y = rhs.0.shl(Self::BITS - dbits).to_limbs();
```
fjarri, line 59: Same here, shouldn't it be just `lshift`?

src/non_zero.rs
andrewwhitehead: Deriving Default seems pretty broken to me, since it produces `NonZero(0)`. Having a Default implementation isn't strictly necessary but it does make `NonZero::new(_).map` possible, as used in `Uint::checked_div`.

src/uint/boxed/div.rs
andrewwhitehead: No, it may be shifting by multiple words.

src/non_zero.rs
tarcieri: I think it's OK, although I'd note there aren't corresponding `Default` impls on `core::num::NonZero*`

tarcieri (approved):

andrewwhitehead:
I saw one `dudect` failure after this was merged which had me concerned, but it seems to be passing since then? I might try to run it on an Intel machine today.

PR #644, opened 2024-08-08
Clean up clippy warnings, asserts in proptests

andrewwhitehead:


src/modular/boxed_monty_form.rs
```diff
@@ -182,15 +182,15 @@ impl BoxedMontyForm {
     pub fn zero(params: BoxedMontyParams) -> Self {
         Self {
             montgomery_form: BoxedUint::zero_with_precision(params.bits_precision()),
-            params: params.into(),
+            params,
```
tarcieri, line 185: This conditionally uses `Arc` depending on if `std` is available (really it should be if `alloc` is available), so the clippy warning here is likely a false positive for the case that `Arc` isn't available

tarcieri: Here's a fix which unconditionally uses `Arc`: #645

tarcieri (approved):

PR #646, opened 2024-08-11
Some simplifications of vartime division

fjarri:
- Extract shl/shr by a limb-sized shift into separate methods
- Add a more detailed docstring and some debug assertions in `div3by2()`
- Replaced `saturating_sub` with `wrapping_sub` in several places. While logically it is the same thing (the wrapping/saturation only happens for values that are later selected out), I think there are readability advantages. First, elsewhere in the code we use wrapping ops for selected out values (meaning "perform the subtraction without any checks, since we already have a constant-time condition for that"), so `saturating_sub` indicates that the algorithm actually uses the saturation mechanic. Second, in case of a bug, it will be easier to spot the consequences of a "0xffff..." value than a 0.

Questions:
- should I move the new shl/shr methods to `shl.rs`/`shr.rs`? Are they general enough?
- should `div3by2_vartime()` be moved out of `div_limb.rs`?

### DIFF
--- a/src/uint/boxed/div.rs
+++ b/src/uint/boxed/div.rs
@@ -310,6 +310,42 @@ impl RemLimb for BoxedUint {
     }
 }
 
+/// Computes `limbs << shift` inplace, where `0 <= shift < Limb::BITS`, returning the carry.
+fn shl_limb_vartime(limbs: &mut [Limb], shift: u32) -> Limb {
+    if shift == 0 {
+        return Limb::ZERO;
+    }
+
+    let lshift = shift;
+    let rshift = Limb::BITS - shift;
+    let limbs_num = limbs.len();
+
+    let carry = limbs[limbs_num - 1] >> rshift;
+    for i in (1..limbs_num).rev() {
+        limbs[i] = (limbs[i] << lshift) | (limbs[i - 1] >> rshift);
+    }
+    limbs[0] <<= lshift;
+
+    carry
+}
+
+/// Computes `limbs >> shift` inplace, where `0 <= shift < Limb::BITS`.
+fn shr_limb_vartime(limbs: &mut [Limb], shift: u32) {
+    if shift == 0 {
+        return;
+    }
+
+    let lshift = Limb::BITS - shift;
+    let rshift = shift;
+
+    let limbs_num = limbs.len();
+
+    for i in 0..limbs_num - 1 {
+        limbs[i] = (limbs[i] >> rshift) | (limbs[i + 1] << lshift);
+    }
+    limbs[limbs_num - 1] >>= rshift;
+}
+
 /// Computes `x` / `y`, returning the quotient in `x` and the remainder in `y`.
 ///
 /// This function operates in variable-time. It will panic if the divisor is zero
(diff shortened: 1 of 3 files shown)


src/uint/boxed/div.rs
```diff
@@ -336,51 +372,42 @@ pub(crate) fn div_rem_vartime_in_place(x: &mut [Limb], y: &mut [Limb]) {
...
-        }
-        (_, borrow) = x_hi.sbb(carry, borrow);
+        let borrow = {
+            let mut carry = Limb::ZERO;
+            let mut borrow = Limb::ZERO;
+            let mut tmp;
+            for i in 0..yc {
+                (tmp, carry) = Limb::ZERO.mac(y[i], Limb(quo), carry);
+                (x[xi + i + 1 - yc], borrow) = x[xi + i + 1 - yc].sbb(tmp, borrow);
+            }
```
andrewwhitehead, line 397: Perhaps this bit should be extracted into a method like `mac_by_limb`?

src/uint/boxed/div.rs
```diff
@@ -336,51 +372,42 @@ pub(crate) fn div_rem_vartime_in_place(x: &mut [Limb], y: &mut [Limb]) {
...
+        };
 
         // If the subtraction borrowed, then decrement q and add back the divisor
         // The probability of this being needed is very low, about 2/(Limb::MAX+1)
-        let ct_borrow = ConstChoice::from_word_mask(borrow.0);
-        carry = Limb::ZERO;
-        for i in 0..yc {
-            (x[xi + i + 1 - yc], carry) =
-                x[xi + i + 1 - yc].adc(Limb::select(Limb::ZERO, y[i], ct_borrow), carry);
+        if borrow.0 > 0 {
```
andrewwhitehead, line 404: The old method was intended to be constant-time for the dividend, but I don't think that this would be.

src/uint/div_limb.rs
```diff
@@ -145,43 +145,49 @@ pub(crate) const fn div2by1(u1: Word, u0: Word, reciprocal: &Reciprocal) -> (Wor
...
-/// divisor word `v0`. The dividend and the lower divisor word must be left-shifted
-/// according to the `shift` value  of the reciprocal.
+/// Given two long integers `u = (..., u0, u1, u2)` and `v = (..., v0, v1)`
+/// (where `u2` and `v1` are the most significant limbs), where `floor(u / v) <= Limb::MAX`,
+/// calculates `q` such that `q - 1 <= floor(u / v) <= q`.
+/// In place of `v1` takes its reciprocal, and assumes that `v` was already pre-shifted
+/// so that v1 has its most significant bit set (that is, the reciprocal's `shift` is 0).
 #[inline(always)]
-pub(crate) const fn div3by2(
+pub(crate) const fn div3by2_vartime(
```
andrewwhitehead, line 154: I'm also worried about the runtime of this method depending on the dividend value. Replacing it completely would also mess with my PR #643, do you want to do a quick review of that one?

src/uint/boxed/div.rs
fjarri: Fair point, I think the new `div3by2` is also not constant-time wrt the dividend. What's more important, 15-35% performance increase, or the partial constant-timeness property?

src/uint/div_limb.rs
fjarri: The original `div3by2()` can be preserved, of course, but I am worried about the connection of its contents and the original algorithm from the article. Could you add some explanations about why it actually does the same thing?

src/uint/boxed/div.rs
andrewwhitehead: It depends on the use case of course, but this version couldn't be used in `Uint::mul_mod_vartime` for example, without changing the guarantees there as well.

src/uint/div_limb.rs
andrewwhitehead: I added a PR to better align that method with the definition, as it doesn't look like overflow was being handled properly in the remainder.

tarcieri:
It seems like we should probably get #643 merged first since it might affect this

src/uint/boxed/div.rs
fjarri: I don't think such generalization will be easy without mutable refs support in constant-time functions

tarcieri: FWIW, I believe this fixes one of the last blockers on stabilizing `const_mut_refs`: https://github.com/rust-lang/rust/pull/128543

src/uint/div_limb.rs
tarcieri: @andrewwhitehead can you take a look again now that #643 is merged?

src/uint/div_limb.rs
```diff
@@ -174,12 +178,12 @@ pub(crate) const fn div3by2(
...
             .or(ConstChoice::from_wide_word_le(qy, rx));
-        quo = done.select_word(quo.saturating_sub(1), quo);
-        rem = done.select_wide_word(rem + (reciprocal.divisor_normalized as WideWord), rem);
+        quo = done.select_word(quo.wrapping_sub(1), quo);
+        rem = done.select_wide_word(rem + (v1_reciprocal.divisor_normalized as WideWord), rem);
         i += 1;
     }
 
-    (quo, rem)
+    quo
```
andrewwhitehead, line 186: The constant-time division does use this remainder

andrewwhitehead:
Looks good to me.

I think at some point it would make sense to turn functions like `shl_limb_vartime` into methods on a struct like `LimbsMut(&mut [Limb])`, which might be a convenient way to collect and document them. Then once there is language support they could be const-ified and used by both `Uint` and `BoxedUint`.

tarcieri (approved):

src/uint/div_limb.rs
tarcieri: Oops, I now see this broke the build.

tarcieri:
Going to revert this merge as it broke the build.

PR #648, opened 2024-08-13
Move schoolbook squaring implementation into a macro

andrewwhitehead:
Applying this to BoxedUint results in an almost 50% speedup at 4096 bits.

Also adding benchmarks for multiplication.


tarcieri:
Hmm, swear I tried this and it didn't effect a performance increase with `BoxedUint`, but this looks like it's worth a shot

Issue #585, opened 2024-03-25
Easier working with native numbers


dignifiedquire:
In my usage with `BoxedUint` I am really missing comparison and calculations with `u32` and `u64` at least, currently it is quite inefficient as I have to do the following

```rust
let x = make_boxed_uint();
let y = x - BoxedUint::from(4); // today
let y = x - 1; // should be easily possible
```

tarcieri:
That would require an impl with `i32` as the argument. We don't support `i32` for anything except `shl`/`shr` currently.

I guess specifically for the case of `Add`/`Sub` we could take the absolute value of the argument and if it's negative, switch from `Add` to `Sub` or `Sub` to `Add` accordingly, before converting the absolute value into a `BoxedUint`.

Do you want this just for `i32` or other types of arguments?

dignifiedquire:
Sorry my example is bad, I am happy with just `u32` and `u64` support for this. (Updated the example)

TitouanReal:
Would manually writing implementations be the best way to go about this? This is a lot of repetitive implementations.

I just did it locally for multiplying `ConstMontyForm` with u8, u16, u32, u64 and u128, and it is already 40 `Mul` implementations, if you want to make it work commutatively and with references (ie &{integer}).

If you do not make it work with &{integer}, the number of implementations is divided by 2.

tarcieri:
Here's a start which impls `Add` for `BoxedUint`: https://github.com/RustCrypto/crypto-bigint/pull/650

If that pattern seems acceptable I can do it for `Sub` and `Mul` as well.

TitouanReal:
Would it be possible to have this for ConstMontyForm too? Right now this is even more cumbersome than for BoxedUint, because one has to go from native numbers to Uint and then to ConstMontyForm.

```rust
let a: ConstMontyForm<...> = ...;
let b = a * ConstMontyForm::new(&Uint::from(3u32)); // today
let b = a * 3u32; // should be easily possible
```

tarcieri:
@dignifiedquire did you need anything else here for `rsa` before we cut a release?

PR #650, opened 2024-08-15
BoxedUint: `Add` impls for stack allocated types

tarcieri:
Adds `Add` impls for `BoxedUint` for the following `rhs` types:

- `Uint<LIMBS>`
- `u8`
- `u16`
- `u32`
- `u64`
- `u128`

The primitive integer types convert to `Uint<LIMBS>` then leverate the `Add` impl for that.

See also: #585


tarcieri:
Some tests here would be good

PR #652, opened 2024-08-16
Bernstein-Yang: support boxed GCD with different size inputs

tarcieri:
Implicitly widens inputs to the size of the largest before computing Bernstein-Yang for GCD.

### DIFF
--- a/src/modular/bernstein_yang/boxed.rs
+++ b/src/modular/bernstein_yang/boxed.rs
@@ -5,8 +5,11 @@
 
 use super::{inv_mod2_62, iterations, jump, Matrix};
 use crate::{BoxedUint, Inverter, Limb, Odd, Word};
-use alloc::{boxed::Box, vec::Vec};
-use core::ops::{AddAssign, Mul};
+use alloc::boxed::Box;
+use core::{
+    cmp::max,
+    ops::{AddAssign, Mul},
+};
 use subtle::{Choice, ConditionallySelectable, ConstantTimeEq, ConstantTimeGreater, CtOption};
 
 /// Modular multiplicative inverter based on the Bernstein-Yang method.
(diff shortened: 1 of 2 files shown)


src/modular/bernstein_yang/boxed.rs
```diff
@@ -68,17 +71,14 @@ impl Inverter for BoxedBernsteinYangInverter {
...
+    let nlimbs = bernstein_yang_nlimbs!(max(f.nlimbs(), g.nlimbs()) * Limb::BITS as usize);
 
     let bits_precision = f.bits_precision();
     let inverse = inv_mod2_62(f.as_words());
-    let f = BoxedUnsatInt::from(f);
-    let mut g = BoxedUnsatInt::from(g);
-    let mut d = BoxedUnsatInt::zero(f.nlimbs());
-    let e = BoxedUnsatInt::one(f.nlimbs());
+    let f = BoxedUnsatInt::from(f).widen(nlimbs);
+    let mut g = BoxedUnsatInt::from(g).widen(nlimbs);
```
tarcieri, line 79: This pattern could probably benefit from a conversion that allocates `nlimbs` in the first place, but that will be easier to add after this change.

src/modular/bernstein_yang/boxed.rs
```diff
@@ -380,9 +377,9 @@ impl BoxedUnsatInt {
 
     /// Widen self to the given number of limbs.
     pub fn widen(self, nlimbs: usize) -> Self {
-        let mut limbs = Vec::from(self.0);
-        debug_assert!(nlimbs >= limbs.len());
-        limbs.truncate(nlimbs);
+        debug_assert!(nlimbs >= self.nlimbs(),);
+        let mut limbs = self.0.into_vec();
+        limbs.resize(nlimbs, 0);
```
tarcieri, line 382: Seems the original `widen` implementation was buggy

PR #659, opened 2024-08-21
Add radix string encoding for Uint, BoxedUint

andrewwhitehead:


src/uint/encoding.rs
```diff
@@ -476,6 +498,252 @@ fn decode_str_radix_aligned_digits<D: DecodeByLimb>(
...
+    } else {
+        vec_buf.extend_from_slice(limbs);
+        &mut vec_buf[..limb_count]
+    };
+    radix_encode_limbs_mut_to_string(radix, buf)
+}
+
+#[cfg(feature = "alloc")]
+pub(crate) fn radix_encode_limbs_mut_to_string(radix: u32, limbs: &mut [Limb]) -> String {
+    if !(2..=26).contains(&radix) {
```
fjarri, line 518: The comment above says "Panics if `radix` is not in the range from 2 to 36."

andrewwhitehead: Oops, apparently I forgot halfway through what the upper limit was meant to be. Fixed now.

fjarri: Perhaps it could be made a named constant?

fjarri:
I'm not sure if this PR is in finished state now, or still WIP, but I think it would be useful to some doc comments for the new structures and methods - it's pretty hard to guess the exact contracts right now.

andrewwhitehead:
Added a few doc comments. I think the only question I have is whether the methods should panic on an invalid radix (consistent with the stdlib `from_str_radix` methods) or return a Result.

fjarri:
I'd say `from_*` is fine with panicking, and if we need `Result`-returning methods, we can add `try_from_*` ones.

tarcieri (approved):

PR #661, opened 2024-08-21
Some simplifications of vartime division (second attempt)

fjarri:
This is a fix of reverted #646 

- Extract shl/shr by a limb-sized shift into separate methods
- Add a more detailed docstring and some debug assertions in `div3by2()`
- Replaced `saturating_sub` with `wrapping_sub` in several places. While logically it is the same thing (the wrapping/saturation only happens for values that are later selected out), I think there are readability advantages. First, elsewhere in the code we use wrapping ops for selected out values (meaning "perform the subtraction without any checks, since we already have a constant-time condition for that"), so `saturating_sub` indicates that the algorithm actually uses the saturation mechanic. Second, in case of a bug, it will be easier to spot the consequences of a "0xffff..." value than a 0.

Update:
- Use `div2by1` instead of `div3by2` with zero lower limbs. That's why I was wondering if `div3by2` should be moved out of `div_limb.rs` — when it's there it's perceived as a general-purpose method analogous to `div2by1`, but it's not - it may return the quotient **or** the quotient plus 1. It is only supposed to be used in a specific place in the division algorithm. It returns a `WideWord` as a remainder, which was truncated to `Word` by the caller without any checks or explanations why it's supposed to be correct.

### DIFF
--- a/src/uint/boxed/div.rs
+++ b/src/uint/boxed/div.rs
@@ -1,9 +1,12 @@
 //! [`BoxedUint`] division operations.
 
 use crate::{
-    uint::{boxed, div_limb::div3by2},
+    uint::{
+        boxed,
+        div_limb::{div2by1, div3by2},
+    },
     BoxedUint, CheckedDiv, ConstChoice, ConstantTimeSelect, DivRemLimb, Limb, NonZero, Reciprocal,
-    RemLimb, Word, Wrapping,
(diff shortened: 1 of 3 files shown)


tarcieri:
cc @andrewwhitehead

tarcieri (approved):

PR #663, opened 2024-08-22
Derive Hash for a bunch of types

dvdplm:
Some of the types in the crate do not derive `Hash` which make them difficult to use in hashmaps or sets.

It would be nice to make `Checked` also be `Hash`, but that requires https://github.com/dalek-cryptography/subtle/pull/138 so perhaps that can be a separate PR.

### DIFF
--- a/src/modular/const_monty_form.rs
+++ b/src/modular/const_monty_form.rs
@@ -68,7 +68,7 @@ pub trait ConstMontyParams<const LIMBS: usize>:
 /// The modulus is constant, so it cannot be set at runtime.
 ///
 /// Internally, the value is stored in Montgomery form (multiplied by MOD::ONE) until it is retrieved.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct ConstMontyForm<MOD: ConstMontyParams<LIMBS>, const LIMBS: usize> {
     montgomery_form: Uint<LIMBS>,
     phantom: PhantomData<MOD>,
--- a/src/modular/monty_form.rs
+++ b/src/modular/monty_form.rs
@@ -17,7 +17,7 @@ use crate::{Concat, Limb, Monty, NonZero, Odd, Split, Uint, Word};
 use subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
 
 /// Parameters to efficiently go to/from the Montgomery form for an odd modulus provided at runtime.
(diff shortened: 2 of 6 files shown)


src/modular/const_monty_form.rs
```diff
@@ -68,7 +68,7 @@ pub trait ConstMontyParams<const LIMBS: usize>:
 /// The modulus is constant, so it cannot be set at runtime.
 ///
 /// Internally, the value is stored in Montgomery form (multiplied by MOD::ONE) until it is retrieved.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
tarcieri, line 71: I'm a little concerned about deriving it for potentially secret-containing types. Though the default `Hasher` is `SipHash` which is fine when properly keyed, other hashers do not necessarily prevent preimage attacks

src/modular/monty_form.rs
```diff
@@ -145,7 +145,7 @@ impl<const LIMBS: usize> ConstantTimeEq for MontyParams<LIMBS> {
 
 /// An integer in Montgomery form represented using `LIMBS` limbs.
 /// The odd modulus is set at runtime.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
tarcieri, line 148: Ditto here re: potentially secret-containing types

src/modular/const_monty_form.rs
dvdplm: Do you think documenting the caveats properly is enough? A bigint library that blocks all potential security misuses risks crippling itself a fair bit and it’s not obvious where to draw the line. I’d argue that in this case it’s expected that users know about the security hazards.

dvdplm:
> This library is documented as being [constant-time by default](https://docs.rs/crypto-bigint/0.6.0-rc.3/crypto_bigint/index.html#goals) and we are careful to treat each integer as potentially containing a secret, with separate *_vartime methods which act in variable-time so as to make those sort of secret accesses easier to audit.

Thinking about this further I now think you are absolutely right. If `crypto-bigint`'s stance is that its integers are all potentially secrets, then its types should not be (easily) usable in hash maps or sets. Users would have to have deep knowledge of how the hasher used in their code behaves wrt CT and pre-image resistance etc (or pick a cryptographic hash, with the perf penalty).

This was a dumb idea. Closing. :)

PR #664, opened 2024-08-28
Montgomery form arithmetic improvements

andrewwhitehead:
- Adds benchmarks for Montgomery arithmetic methods
- Adds a `double` method for Montgomery forms and `double_mod` method for Uint/BoxedUint, using a left shift instead of adding `self`
- Uses the existing (faster) `neg_mod` methods in Montgomery form `neg` methods
- Optimizes `add_mod` and `sub_mod` for `BoxedMontyForm` to remove allocations


tarcieri (approved):
These all look like good improvements

PR #666, opened 2024-09-04
Add efficient linear combination for Montgomery forms

andrewwhitehead:
This PR implements Algorithm 2 (for B=1) from Efficient Algorithms for Large Prime Characteristic Fields and Their Application to Bilinear Pairings by Patrick Longa: https://eprint.iacr.org/2022/367

This algorithm interleaves schoolbook multiplication and accumulation of multiple terms with the Montgomery reduction. For moduli with one or more leading zeros, this helps to reduce the number of reductions performed. In the `BoxedUint` case this also reduces the number of allocations required. For larger moduli it may be useful to explore the use of larger values of B along with the existing Karatsuba multiplication.

As a concrete example, the calculation of `(a•b + c•d) mod m` using `ConstMontyForm`, where `m` is a `U256` with at least one leading zero, is reduced from 50 to 32ns in my tests.

### DIFF
--- a/src/modular/lincomb.rs
+++ b/src/modular/lincomb.rs
@@ -1,0 +1,171 @@
+use crate::{modular::MontyForm, Limb, Odd, Uint};
+
+use super::{ConstMontyForm, ConstMontyParams};
+
+#[cfg(feature = "alloc")]
+use super::BoxedMontyForm;
+#[cfg(feature = "alloc")]
+use crate::BoxedUint;
+
+/// Implement the coarse interleaved sum of products (Algorithm 2 for B=1) from
+/// Efficient Algorithms for Large Prime Characteristic Fields and Their Application
+/// to Bilinear Pairings by Patrick Longa. https://eprint.iacr.org/2022/367
+///
+/// For correct results, the un-reduced sum of products must not exceed `p•R`  where `p`
+/// is the modulus. Given a list of pairs `(a_1, b_1)..(a_k, b_k)` in Montgomery form,
+/// where each `a_i < p` and `b_i < p`, we have `sum(a_i•b_i) < k•p^2` and so up to
+/// `k = floor(R/p)` pairs may be safely accumulated per call.
+///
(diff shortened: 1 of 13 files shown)


src/modular/lincomb.rs
```diff
@@ -0,0 +1,171 @@
...
+///
+/// For correct results, the un-reduced sum of products must not exceed `p•R`  where `p`
+/// is the modulus. Given a list of pairs `(a_1, b_1)..(a_k, b_k)` in Montgomery form,
+/// where each `a_i < p` and `b_i < p`, we have `sum(a_i•b_i) < k•p^2` and so up to
+/// `k = floor(R/p)` pairs may be safely accumulated per call.
+///
+/// This is implemented as a macro to abstract over `const fn` and boxed use cases, since the latter
+/// needs mutable references and thus the unstable `const_mut_refs` feature (rust-lang/rust#57349).
+///
+// TODO: change this into a `const fn` when `const_mut_refs` is stable
```
tarcieri, line 22: Soon! #667 

I guess we can add one more macro in the meantime.

src/modular/monty_form.rs
```diff
@@ -30,6 +31,8 @@ pub struct MontyParams<const LIMBS: usize> {
     /// The lowest limbs of -(MODULUS^-1) mod R
     /// We only need the LSB because during reduction this value is multiplied modulo 2**Limb::BITS.
     mod_neg_inv: Limb,
+    /// Leading zeros in the modulus, used to choose optimized algorithms
+    mod_leading_zeros: u32,
```
tarcieri, line 35: Hmm, I get the use cases, but this does seem like a potential sharp edge

tarcieri (approved):

src/modular/monty_form.rs
andrewwhitehead: Vartime only!

tarcieri: Yeah, I'm just worried someone down the road might be used in constant-time use cases

PR #670, opened 2024-09-10
Bump `serdect` to v0.3.0-rc.0

tarcieri:


tarcieri:
Hmm:

```
failures:

---- odd::tests::serde_tests::silently_coerces_bigger_type_into_smaller_type stdout ----
thread 'odd::tests::serde_tests::silently_coerces_bigger_type_into_smaller_type' panicked at src/odd.rs:331:13:
assertion `left == right` failed
  left: "invalid length 16, expected an array of length 8"
 right: "Slice had bytes remaining after deserialization"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- odd::tests::serde_tests::cannot_deserialize_into_bigger_type stdout ----
thread 'odd::tests::serde_tests::cannot_deserialize_into_bigger_type' panicked at src/odd.rs:314:13:
assertion `left == right` failed
  left: "invalid length 8, expected an array of length 16"
 right: "io error: unexpected end of file"


failures:
    odd::tests::serde_tests::cannot_deserialize_into_bigger_type
    odd::tests::serde_tests::silently_coerces_bigger_type_into_smaller_type
```

tarcieri:
@daxpedda I'm curious if the new length prefixes actually address the problems you were running into with `bincode` and we can just remove these failing tests

daxpedda:
Sorry, what new length prefixes?

In any case, this tests where only there for sanity, its not a bug in my opinion if users are able to decode into the wrong type.

tarcieri:
@daxpedda `serdect` v0.3 prereleases now include a length prefix, even when encoding arrays whose sizes should be known in advance, which is probably catching things which weren't being caught before regarding `bincode` length mismatches

daxpedda:
Ah, yes, that makes sense!
It would indeed address the original issue!

Thanks for the ping and the pointer!

PR #688, opened 2024-10-08
Leverage `div_ceil`

tarcieri:
Fixes clippy warnings about `div_ceil` which are appearing on recent nightlies

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -5,7 +5,7 @@
 #[macro_export]
 macro_rules! nlimbs {
     ($bits:expr) => {
-        (($bits + $crate::Limb::BITS - 1) / $crate::Limb::BITS) as usize
+        u32::div_ceil($bits, $crate::Limb::BITS) as usize
     };
 }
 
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -5,7 +5,7 @@
 #[macro_export]
 macro_rules! nlimbs {
     ($bits:expr) => {
-        (($bits + $crate::Limb::BITS - 1) / $crate::Limb::BITS) as usize
+        u32::div_ceil($bits, $crate::Limb::BITS) as usize
```
tarcieri, line 8: Note: can't use the following due to type inference problems:

```rust
$bits.div_ceil($crate::Limb::BITS) as usize
```

It causes the following error:

```
error[E0689]: can't call method `div_ceil` on ambiguous numeric type `{integer}`
   --> src/macros.rs:8:15
    |
8   |           $bits.div_ceil($crate::Limb::BITS) as usize
    |                 ^^^^^^^^
```

Issue #624, opened 2024-07-30
Supporting Signed Integer Arithmetic


erik-3milabs:
I'm interested in performing cryptographic arithmetic involving negative (signed) integers. As is made evident in #1, this crate does not yet provide support for this. 

I'm exploring what writing the necessary code to support this would take. To progress my exploration, I'd like to learn what additions would be valued/accepted to this library. In particular, what would be considered a "minimal useful addition"? What functionality _must_ be included?  Adding support for all operations (right off the bat) might be beyond my current capacity.

Looking forward to your input!

tarcieri:
As I don't happen to work on cryptographic algorithms that require signed arithmetic (instead we use modular arithmetic where negative values wrap around a modulus, e.g. DSA, RSA, and elliptic curves) I can't answer these questions for you.

What cryptographic algorithm are you implementing and how does it leverage signed arithmetic?

erik-3milabs:
I'm working on cryptography involving class groups. In these groups, each element can be represented as a tuple `(a, b, c)` with `a, b, c \in \mathbb{Z}` (called 'form representation'). Each tuple can be reduced to a representative with non-negative `a` and `c`. However,  `b` can still be negative, and intermediate results involve negative coefficients too.

The most prevalent operations on these tuples are normalization, reduction, and composition. These operations use a combination of addition, subtraction, multiplication, and (floored)division on the tuple's elements.

Currently, I'm investigating whether it is "best" to make my 'math fit the code' or make the 'code fit the math'. Here, "best" is a trade-off between the code's development cost, understandability/modifiability, and execution time. It is quite a complex process ;-)

burdges:
Just curious, do any class group protocols involve secrets or otherwise benefit from constant time arithmetic for other reasons?  VDFs need pure speed, but afaik do not benefit from being constant time, but those are likely broken anyways.  An accumulator or such does probably.

erik-3milabs:
The Additive Homomorphic Encryption (AHE) scheme I'm building on top of Class Groups involves operations that I think should be constant time. Two examples:
1. `encrypt(m, pk) -> ct` involves (among other things) raising some predetermined group element `f` to the power `m`. An efficient method would be to do this through the repeated squaring (compose with self) of `f`. If the running time of this composition function is non-constant, one could potentially uncover (parts of) the bit pattern of `m` (which is supposed to be secret).
[...]

PR #694, opened 2024-10-14
[WIP] `Int` - sign-and-magnitude

erik-3milabs:
**What is this?**
This PR provides basic signed integer support into `crypto_bigint`. This is a not-yet-implemented feature (see #1).
This basic implementation involves the following operations: 
- `checked_add`, 
- `checked_sub`, 
- `checked_mul`, and 
- `checked_div`

**Related**
Discussion started in #624.

**About**
This PR puts the first code on the board. The decision was made to implement `Int` as [sign-and-magnitude](https://en.wikipedia.org/wiki/Signed_number_representations#Sign%E2%80%93magnitude) (over, say, [two's complement](https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement)) since it is easier to scale to flexible-size implementations and, in my opinion, less error-prone. Moreover, multiplications yield hardly any overhead compared to operating on `Uint`s; most of the overhead is incurred in the addition operation.

**Open question:**
It was mentioned in #624 to use `Choice` to represent the sign. Sadly, this struct does not seem to have a `const` constructor. If this is indeed the case, we will be unable to make the constants `Int::ZERO`, `::MINUS_ONE`, `::ONE`, `::MIN` and `::MAX` unless we resolve this.

**Area for improvement:**
The current addition/subtraction operations involve
- one `Uint`-addition,
- one `Uint`-subtraction,
- three `CtOption::ct_selects`,
- one `Uint::eq` call, and
- one `Uint::ct_gt` call.

As such, this operation is ~7x more expensive than a `Uint` addition/subtraction. There is at least one clear area for improvement here: add an `is_zero` flag to the `struct`. Doing so will get allow us to get rid of the `ct_eq` call.
In my understanding, converting to and from two's complement will not be an option; the overhead in doing most likely exceeds the 6x performance penalty.

Any feedback is greatly appreciated!


ycscaly:
@erik-3milabs crypto-bigint has their own `const_choice`, which is what I believe @tarcieri meant for you to use.

src/int.rs
```diff
@@ -0,0 +1,167 @@
...
+
+mod add;
+mod cmp;
+mod div;
+mod mul;
+mod neg;
+mod sub;
+
+#[derive(Copy, Clone, Hash, Debug, PartialEq)]
+pub struct Int<const LIMBS: usize> {
```
mkramarczyk, line 16: this does not seam to be u2 complement but absolute value + sign, right?
If so, Why not represent Int as enum of Positive(Uint<L>), Negative(Uint<L>)?

mkramarczyk:
> The decision was made to implement Int as [sign-and-magnitude (https://en.wikipedia.org/wiki/Signed_number_representations#Sign%E2%80%93magnitude) (over, say, [two's complement (https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement))

This does not make sense, it's either U2 or sign + magnitude (absolute value). The latter is used here, not U2.

lleoha:
> The decision was made to implement Int as [sign-and-magnitude](https://en.wikipedia.org/wiki/Signed_number_representations#Sign%E2%80%93magnitude)
My impression was that in discussion it was suggested to rather use U2 not sign + absolute value. Constant operations would be easier in that representation and IMHO it's more suitable for fixed sized integers.

src/int/div.rs
```diff
@@ -0,0 +1,21 @@
+//! [`Int`] division operations.
+
+use subtle::{Choice, ConstantTimeEq, CtOption};
+
+use crate::int::Int;
+use crate::{CheckedDiv, Uint};
+
+impl<const LIMBS: usize> CheckedDiv for Int<LIMBS> {
```
lleoha, line 8: the zero has two representation, I think we should handle this as well, i.e "-0"/whatever should be "-0" not "+0".

src/int/mul.rs
```diff
@@ -0,0 +1,15 @@
+//! [`Int`] multiplication operations.
+
+use subtle::{ConstantTimeEq, CtOption};
+
+use crate::int::Int;
+use crate::{CheckedMul, Uint, Zero};
+
+impl<const LIMBS: usize, const RHS_LIMBS: usize> CheckedMul<Int<RHS_LIMBS>> for Int<LIMBS> {
+    #[inline]
+    fn checked_mul(&self, rhs: &Int<RHS_LIMBS>) -> CtOption<Self> {
```
lleoha, line 10: same as for div, since there are two zero representation the sign should be preserved.
i.e. 
* -0 * -0 = +0
* +0 * +0 = +0
* -0 * +0 = -0
* +0 * -0 = -0

src/int/neg.rs
```diff
@@ -0,0 +1,21 @@
+//! [`Int`] negation operation.
+
+use crate::int::Int;
+use crate::WrappingNeg;
+
+impl<const LIMBS: usize> Int<LIMBS> {
```
lleoha, line 6: same here, there are two representation of zero so sign should behave correctly i.e.:
-(-0) = +0
-(+0) = -0

src/int.rs
```diff
@@ -0,0 +1,167 @@
...
+    // TODO(erik): replace with Choice once I've figured how to const-construct those.
+    is_negative: bool,
```
tarcieri, line 18: `bool` anywhere in constant-time code is worrisome because it's an invitation for e.g. LLVM's optimizer to insert branches based on the value, even in cases where the original code deliberately avoids them

lleoha:
> There is at least one clear area for improvement here: add an is_zero
I am strongly against that for the following reasons. Currently zero has two representations (-0 and +0) adding is_zero will make it from 2 to ~2^(limbs * limb_bits).

src/int.rs
```diff
@@ -0,0 +1,167 @@
...
+mod add;
+mod cmp;
+mod div;
+mod mul;
+mod neg;
+mod sub;
+
+#[derive(Copy, Clone, Hash, Debug, PartialEq)]
+pub struct Int<const LIMBS: usize> {
+    // TODO(erik): replace with Choice once I've figured how to const-construct those.
```
lleoha, line 17: no problem here, worst case scenario, just use u8, CtChoice is more or less that.

tarcieri:
> The decision was made to implement Int as [sign-and-magnitude](https://en.wikipedia.org/wiki/Signed_number_representations#Sign%E2%80%93magnitude) (over, say, [two's complement](https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement)) since it is easier to scale to flexible-size implementations

I'd like to reiterate my [previous comments](https://github.com/RustCrypto/crypto-bigint/issues/624#issuecomment-2372425489) that the heap-allocated `Boxed*` types are generally a much lower priority than the const-generic stack-allocated types.

The `Boxed*` types largely exist to support RSA and DSA, which are algorithms where it might be necessary to support a very high cardinality of potential key sizes.

Pretty much all other algorithms use a very small, fixed number of key sizes.

src/int.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    fn is_zero(&self) -> bool {
+        self.ct_eq(&Self::ZERO).into()
+    }
+}
+
+#[cfg(target_pointer_width = "64")]
+#[allow(dead_code)]
+type I128 = Int<2>;
```
lleoha, line 114: this is where it's tricky, if we want to relate this to builtin i128 type, this is actually closer to I129 because of using sign and magnitude instead of two's complements.

lleoha:
> @erik-3milabs forgot to mention you in my response :).

Issue #693, opened 2024-10-12
Adopt `core` naming conventions for bigint methods


tarcieri:
See also: #537 

The upcoming [`bigint_helper_methods`](https://github.com/rust-lang/rust/issues/85532) feature of `core` defines method names and type signatures which it would be good to adopt here as well where it makes sense.

Here's a name bikeshedding thread, if anyone has any opinions on these names: https://internals.rust-lang.org/t/naming-bigint-helper-methods-bike-shedding/21688

tarcieri:
I'd like to punt on this for the v0.6 release. We can always add the new names incrementally, deprecate the old ones, and remove them in a v0.7 release.

While I'm here though, I would also like to note some inconsistencies in modular inverse naming: we have `inv`, `invert`, and `inv_mod`, along with various `inv_*` and `inv_*_mod` names.

The `*_mod` is typically omitted when working with types that are implicitly modular like `*Monty*`, so that much seems okay, but the use of `inv` and `invert` seems a bit inconsistent at least.

fjarri:
Is this one going to be delayed until v0.7?

tarcieri:
Yes, per my comment immediately above yours

fjarri:
Sorry, I wasn't sure what "punt on this" meant.

tarcieri:
Aah sorry, that is perhaps a little jargony. But yes, the plan is to introduce new names some time after the stable v0.6 release and deprecate the old ones.

PR #695, opened 2024-10-16
`Int` - two's complement

erik-3milabs:
**Topic**
Adding `Int` support (see #624/#700)

To decide on sign-and-magnitude (#694) vs. two's complement, I whipped up a quick two's complement implementation of the basic add/sub/mul/div operations. It's not perfect yet; I'm just curious to see how it compares to #694.


erik-3milabs:
FYI: local benchmark results @ b23483da960a8476b95ab3372166bbac51b2bda5
```
wrapping ops/split_mul, 
I128xI128:    26.042 ns
I256xI256:    26.104 ns
I512xI512:    68.643 ns
I1024xI1024: 239.23  ns
I2048xI2048: 799.26  ns
I4096xI4096: 2.5102  µs

wrapping ops/div, full size
I256/I128: 351.69   ns
I512/I256: 554.63   ns
I1024/I512:  1.0663 µs
I2048/I1024: 2.5858 µs
I4096/I2048: 7.2846 µs

wrapping ops/add, 
I128+I128:     1.1915 ns
I256+I256:     2.4108 ns
I512+I512:    10.374  ns
I1024+I1024:  20.704  ns
I2048+I2048:  65.318  ns
I4096+I4096: 103.10   ns

wrapping ops/sub,
I128-I128:     1.1967 ns
I256-I256:     2.3008 ns
I512-I512:    10.985  ns
I1024-I1024:  32.041  ns
I2048-I2048:  70.872  ns
I4096-I4096: 128.86   ns
```

src/int.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// map `self` to `(self ^ 1111...1111) + 0000...0001`
+    ///
+    /// Returns
+    /// - the result, as well as
+    /// - whether the addition overflowed (indicating `self` is zero).
+    ///
+    /// Warning: this operation is unsafe; when `self == Self::MIN`, the negation fails.
+    #[inline]
+    fn negate_unsafe(&self) -> (Self, Choice) {
+        let inverted = self.0.bitxor(&Self::FULL_MASK.0);
```
lleoha, line 173: `self.0.bitxor(&Self::FULL_MASK.0)`
is the same as `self.0.wrapping_neg()`

src/int.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// Perform the two's complement "negate" operation on this [`Int`]:
+    /// map `self` to `(self ^ 1111...1111) + 0000...0001`
+    ///
+    /// Returns
+    /// - the result, as well as
+    /// - whether the addition overflowed (indicating `self` is zero).
+    ///
+    /// Warning: this operation is unsafe; when `self == Self::MIN`, the negation fails.
+    #[inline]
+    fn negate_unsafe(&self) -> (Self, Choice) {
```
lleoha, line 172: it does what is the same as `self.0.wrapping_neg()` does.

erik-3milabs: Thanks for pointing this out! I adapted `wrapping_neg()` into `wrapping_negc()` to include the carry and still get the `is_zero` for free. In addition to having no duplicate code, this saves anywhere from 8% (`I256`) to 1% (`I4096`) in the multiplication benchmark :ok_hand:

src/uint/sub.rs
```diff
@@ -1,4 +1,4 @@
-//! [`Uint`] addition operations.
+//! [`Uint`] subtraction operations.
```
tarcieri, line 1: Oh my, some copypasta

erik-3milabs: I couldn't help myself :see_no_evil:

tarcieri:
As a general comment, I feel better about the constant-time operation of this approach in terms of not being so trivial for the optimizer to introspect the sign bit as in #694 and introduce a sidechannel ala https://eprint.iacr.org/2015/1141.pdf

src/int.rs
```diff
@@ -0,0 +1,470 @@
...
+        ConstCtOption::new(Odd(self), self.0.is_odd())
+    }
+
+    /// Get the sign bit of this [`Int`] as `Choice`.
+    pub fn sign_bit(&self) -> Choice {
+        ConstChoice::from_word_msb(self.0.to_words()[LIMBS - 1]).into()
+    }
+
+    /// View the data in this type as an [`Uint`] instead.
+    pub const fn as_uint(&self) -> Uint<LIMBS> {
```
lleoha, line 149: usually as_* methods return borrow value (i.e. in this case &Uint<LIMBS>.
So either rename it to to_uint or change return type.

src/int.rs
```diff
@@ -0,0 +1,470 @@
...
+            Self(Uint::ct_select(&self.0, &negated.0, negate)),
+            is_zero.into(),
+        )
+    }
+
+    /// Map this [`Int`] to `-self` if possible.
+    ///
+    /// Yields `None` when `self == Self::MIN`, since an [`Int`] cannot represent the positive
+    /// equivalent of that.
+    pub fn neg(&self) -> CtOption<Self> {
```
lleoha, line 199: move it from here to neg.rs (all other negation related methods). Similarly to what Uint has.

src/int.rs
```diff
@@ -0,0 +1,470 @@
...
+
+    /// Const-friendly [`Int`] constructor.
+    pub const fn new(limbs: [Limb; LIMBS]) -> Self {
+        Self(Uint::new(limbs))
+    }
+
+    /// Const-friendly [`Int`] constructor.
+    /// Note: interprets the `value` as an `Int`. For a proper conversion,
+    /// see [`Int::new_from_sign_and_magnitude`].
+    pub const fn new_from_uint(value: Uint<LIMBS>) -> Self {
```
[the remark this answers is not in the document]
tarcieri: The name is a bit tricky... this is effectively a casting/transmutation operation that reinterprets the bits of a `Uint` as an `Int`.

src/int/mul.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use subtle::{Choice, ConstantTimeEq, CtOption};
+
+use crate::{Checked, CheckedMul, Int, Uint, Zero};
+
+impl<const LIMBS: usize> Int<LIMBS> {
+    /// Compute "wide" multiplication as a 3-tuple `(lo, hi, sgn)`.
+    /// The `(lo, hi)` components contain the _magnitude of the product_, with sizes
+    /// correspond to the sizes of the operands; the `sgn` indicates the sign of the result.
+    pub fn split_mul<const RHS_LIMBS: usize>(
```
lleoha, line 13: it would be nice to have it const fn

src/int/mul.rs
```diff
@@ -0,0 +1,181 @@
...
+        &self,
+        rhs: &Int<RHS_LIMBS>,
+    ) -> (Uint<{ LIMBS }>, Uint<{ RHS_LIMBS }>, Choice) {
```
lleoha, line 16: it is actually a way to make is slightly easier.
Here's as an example my implementation of widening_mul
```rust
    pub const fn widening_mul<const RHS_LIMBS: usize, const WIDE_LIMBS: usize>(
        &self,
        rhs: &Int<RHS_LIMBS>,
    ) -> Int<WIDE_LIMBS>
    where
        Uint<LIMBS>: ConcatMixed<Uint<RHS_LIMBS>, MixedOutput=Uint<WIDE_LIMBS>>,
    {
        let l_sign = self.is_negative();
        let l_pos = Uint::select(&self.repr, &self.repr.wrapping_neg(), l_sign);
        let r_sign = self.is_negative();
        let r_pos = Uint::select(&rhs.repr, &rhs.repr.wrapping_neg(), r_sign);
        let p = l_pos.widening_mul(&r_pos);
        
        Int { repr: Uint::select(&p, &p.wrapping_neg(), l_sign.xor(r_sign)) }
    }
}
```

src/int.rs
```diff
@@ -0,0 +1,470 @@
...
+
+    /// Convert to a [`Odd<Int<LIMBS>>`].
+    ///
+    /// Returns some if the original value is odd, and false otherwise.
+    pub const fn to_odd(self) -> ConstCtOption<Odd<Self>> {
+        ConstCtOption::new(Odd(self), self.0.is_odd())
+    }
+
+    /// Get the sign bit of this [`Int`] as `Choice`.
+    pub fn sign_bit(&self) -> Choice {
```
lleoha, line 144: make it const fn returning ConstChoice. nit: I would rename it to `is_negative()`

src/int/add.rs
```diff
@@ -0,0 +1,206 @@
...
+use core::ops::{Add, AddAssign};
+
+use num_traits::WrappingAdd;
+use subtle::{ConstantTimeEq, CtOption};
+
+use crate::{Checked, CheckedAdd, Int, Wrapping};
+
+impl<const LIMBS: usize> Int<LIMBS> {
+    /// Perform checked addition.
+    fn checked_add_internal(&self, rhs: &Self) -> CtOption<Self> {
```
lleoha, line 12: make it const fn

lleoha, line 12: nit: remove _internal prefix to be consistent with Uint

src/int/expand.rs
```diff
@@ -0,0 +1,17 @@
+use crate::{ConstantTimeSelect, Int, Uint};
+
+impl<const LIMBS: usize> Int<LIMBS> {
+    /// Expand the representation of `self` to an `Int<T>`.
+    /// Assumes `T >= LIMBS`; panics otherwise.
+    #[inline(always)]
+    pub fn expand<const T: usize>(&self) -> Int<T> {
```
lleoha, line 7: Uint has resize, maybe we can keep consistency and name it the same?

src/int/expand.rs
```diff
@@ -0,0 +1,17 @@
...
+        assert!(T >= LIMBS);
```
lleoha, line 8: make it debug_assert (the same as Uint does)

src/int.rs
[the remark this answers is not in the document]
lleoha: yeah, I was having similar problems with naming and couldn't come up with anything better. you have `as` operator in rust that does this for primitive types so maybe we can mimic that and implement as<Uint> as<Int> (or cast<...>?) to be consistent with at least that.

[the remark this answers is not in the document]
tarcieri: The only thing that comes to mind for me is [`from_bits`](https://doc.rust-lang.org/std/primitive.f32.html#method.from_bits) which reinterprets an integer as a floating point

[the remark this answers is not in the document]
erik-3milabs: Perhaps [`transmute`](https://doc.rust-lang.org/std/mem/fn.transmute.html) is the way to go?

> Reinterprets the bits of a value of one type as another type.
> Both types must have the same size. Compilation will fail if this is not guaranteed.

 I is mentioned in the `from_bits` doc that it is identical to `transmute`.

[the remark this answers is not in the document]
erik-3milabs: @lleoha wait, one cannot implement `as`, right? Because, I agree, this definitely sounds like an `as` operation. I would steer clear from using `from` / `into`, as those names suggests a "proper" conversion, which is definitely not the case here.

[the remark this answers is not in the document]
lleoha: @erik-3milabs Yes, you cannot overload `as` operator but I am big fan of consistency so somewhat see no reason why you shouldn't be able to do something like:
`Int::<2>::ONE.as_uint()` or `Int::<2>::ONE.as::<Uint>()`
when you can do `1_i128 as u128`.
I am also completely fine with `from_bits` and/or `transmute`.

src/uint/neg.rs
```diff
@@ -3,6 +3,11 @@ use crate::{Limb, Uint, WideWord, Word, WrappingNeg};
 impl<const LIMBS: usize> Uint<LIMBS> {
     /// Perform wrapping negation.
     pub const fn wrapping_neg(&self) -> Self {
+        self.wrapping_negc().0
+    }
+
+    /// Perform wrapping negation; returns the resulting carry.
+    pub const fn wrapping_negc(&self) -> (Self, Word) {
```
lleoha, line 10: The name of it is misleading.
The operation is either wrapping or carrying (for comparison see add_wrapping/adc) hence I would rename it to just `negc`.

src/int.rs
```diff
@@ -0,0 +1,470 @@
...
+        ConstChoice::from_word_msb(self.0.to_words()[LIMBS - 1]).into()
+    }
+
+    /// View the data in this type as an [`Uint`] instead.
+    pub const fn as_uint(&self) -> Uint<LIMBS> {
+        self.0
+    }
+
+    /// Whether this [`Int`] is equal to `Self::MIN`.
+    pub fn is_minimal(&self) -> Choice {
```
lleoha, line 154: for consistency if we define `Int::MIN` I would rename it to `is_min()`, similarly for `is_max()`.
Also make then const fn return `ConstChoice`.

src/uint/neg.rs
```diff
@@ -12,7 +17,7 @@ impl<const LIMBS: usize> Uint<LIMBS> {
             carry = r >> Limb::BITS;
             i += 1;
         }
-        Uint::new(ret)
+        (Uint::new(ret), carry as Word)
```
lleoha, line 20: btw. to fix that you would have to return `!0 + carry` as carry (which again is consistent with how you compute every limb value). I wanted to just explain above why is that a case.

src/int/mul.rs
erik-3milabs: Fixed it.

src/int.rs
lleoha: btw. why do you need is_zero in the first place?

src/int.rs
lleoha: btw. here's the same story as transmute discussion here: https://github.com/RustCrypto/crypto-bigint/pull/695#discussion_r1803457240

src/int/add.rs
erik-3milabs: I did this because, indeed, it makes things consistent. However (!), there are now two `checked_add`s for `Int`:
- The `ConstCtOption`, and
- The `CtOption` (trait) version.

src/int/mul.rs
lleoha: we already have `Uint::select`

src/int/expand.rs
erik-3milabs: When developing, I initially used `resize` as well. However, it felt weird to downsize an `Int`, knowing that you could be losing sign information :thinking: 

But then again, I suppose the user should know that this is an "unsafe" operation that could lead to data loss :man_shrugging: 

I'll make it happen.

src/int.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub const MINUS_ONE: Self = Self::FULL_MASK; // Bit sequence (be): 1111....1111
+
+    /// Smallest value this [`Int`] can express.
+    pub const MIN: Self = Self(Uint::MAX.bitxor(&Uint::MAX.shr(1u32))); // Bit sequence (be): 1000....0000
+
+    /// Maximum value this [`Int`] can express.
+    pub const MAX: Self = Self(Uint::MAX.shr(1u32)); // Bit sequence (be): 0111....1111
+
+    /// Bit mask for the sign bit of this [`Int`].
+    pub const SIGN_BIT_MASK: Self = Self::MIN; // Bit sequence (be): 1000....0000
```
lleoha, line 52: for consistency either rename it to `SIGN_MASK` or rename `FULL_MASK` to `FULL_BIT_MASK`. I prefer the former.

src/int/expand.rs
erik-3milabs: was removed.

src/uint/neg.rs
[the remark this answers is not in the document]
erik-3milabs: hmmm, interesting :thinking:

src/int/add.rs
lleoha: The same thing is for Uint there are two versions of the same method (one from trait impl one from struct impl) take bitand as an example:
https://github.com/RustCrypto/crypto-bigint/blob/master/src/uint/bit_and.rs#L11
https://github.com/RustCrypto/crypto-bigint/blob/master/src/uint/bit_and.rs#L55

The only difference is that one is const fn, the other is not (I believe due to lack of stability of this feature yet).

src/int/expand.rs
lleoha: you should not be losing sign information, resize(Int) is not the same as Int(resize(Uint)).
e.g. resize 8-bit `-1` which is `11111111` is perfectly fine resizable to 4 bits as `1111` which is still `-1`, because `-1` has only one bit of information.
In summary when -2^k <= x < 2^k, then you should be able to downsize Int to k-bits without losing it's value (without overflowing or whatever we call it).

src/uint/neg.rs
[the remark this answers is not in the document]
lleoha: The other solution is to name whatever overflows as not carry nor borrow, because these are well defined; but that would be somewhat not standard abuse :)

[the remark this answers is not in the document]
erik-3milabs: Done!

erik-3milabs: done!

src/int.rs
erik-3milabs: done!

src/int.rs
erik-3milabs: done!

src/int.rs
```diff
@@ -0,0 +1,470 @@
...
+    /// Perform the two's complement "negate" operation on this [`Int`]:
+    /// map `self` to `(self ^ 1111...1111) + 0000...0001`
+    ///
+    /// Returns
+    /// - the result, as well as
+    /// - whether the addition overflowed (indicating `self` is zero).
+    ///
+    /// Warning: this operation is unsafe; when `self == Self::MIN`, the negation fails.
+    #[inline]
+    pub const fn negate_unsafe(&self) -> (Self, ConstChoice) {
```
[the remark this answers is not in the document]
erik-3milabs: done!

src/uint/neg.rs
erik-3milabs: done!

src/int/add.rs
erik-3milabs: done!

src/int.rs
erik-3milabs: Well, I _needed_ it for `mul` and `div`; the fix you pointed out there makes knowing whether something is zero unnecessary (for now).

src/int.rs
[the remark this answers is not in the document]
erik-3milabs: Ahh. That is a curious, but elegant solution! Never thought about that :thinking:

src/int.rs
erik-3milabs: done!

src/int/mul.rs
erik-3milabs: Well, guess I'm blind! 
Let me know what you think of this new solution.

src/int.rs
erik-3milabs: done!

src/int/add.rs
erik-3milabs: done!

lleoha:
@tarcieri would it be possible to create something to resemble "feature" branch for this effort? I would like to contribute but I want to avoid merging many branches and avoid conflicts as much as possible.
@erik-3milabs WDYT?

erik-3milabs:
Agreed. This PR is growing beyond something that can be reasonably reviewed. Moreover, I would still need `gcd`, `xgcd` and various `mod` features, so there is enough I also want to add. Having one/more separate branches would help!

src/uint/neg.rs
```diff
@@ -12,6 +20,21 @@ impl<const LIMBS: usize> Uint<LIMBS> {
             carry = r >> Limb::BITS;
             i += 1;
         }
+        (Uint::new(ret), carry.wrapping_add(!0) as Word)
+    }
+
+    /// Perform negation, if `negate` is truthy.
+    pub const fn wrapping_neg_if(&self, negate: ConstChoice) -> Self {
+        let mut ret = [Limb::ZERO; LIMBS];
```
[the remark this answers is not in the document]
erik-3milabs: 1) I wrote this before learning about `Uint::select`
2) This is significantly faster. 

[ref: previous comment](https://github.com/RustCrypto/crypto-bigint/pull/695#discussion_r1804770156)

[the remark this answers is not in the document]
erik-3milabs: My hunch for why it is faster: this implementation iterates over the limb-array only once, whereas the solution you're proposing does so twice: once for the `wrapping_neg` and once for `select`

lleoha:
@erik-3milabs please see https://github.com/erik-3milabs/crypto-bigint/pull/1 I did some work there, see descriptions for rationale for these changes.

ycscaly:
Really good work and @lleoha thanks for working with Erik about this. @tarcieri, what are your thoughts about the deliverable of this PR? is it too much for a single PR? and are there any missing functionality or the only thing we need it to pass correctness review? excited for this to be merged :)

erik-3milabs:
@lleoha could you please resolve the conversations that have been wrapped up? It allows us to focus on what still needs to be done :D

erik-3milabs:
@tarcieri in implementing `from` for `Int` I noticed that there are some
> // TODO(tarcieri): replace with `const impl From<uXX> ` when stable

comments in `Uint::from`. Would there be value in setting up a tracking issue for that and replacing them by `//TODO(#XXX)` ?

tarcieri:
> Would there be value in setting up a tracking issue for that and replacing them by //TODO(#XXX) ?

@erik-3milabs we could do something like that, yeah

tarcieri:
> @tarcieri would it be possible to create something to resemble "feature" branch for this effort? I would like to contribute but I want to avoid merging many branches and avoid conflicts as much as possible.

Sure, I can make a `signed-int` branch or thereabouts. Though we'll need to merge this initial spike like this to it for it to have value, and it would be good to get that properly reviewed (i.e. this PR can be changed to target the new branch)

tarcieri:
`signed-int` has been branched from `master`

lleoha:
> `signed-int` has been branched from `master`

Thanks @tarcieri

lleoha (approved):
Let's make it into the feature branch so we have a baseline to work on.
LGTM to initial commit.

tarcieri:
Will do another pass tomorrow, but I agree this is already a lot

src/int/from.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn from_i32(n: i32) -> Self {
+        assert!(LIMBS >= 1, "number of limbs must be greater than zero");
+        Int::new_from_uint(Uint::new([Limb(n as Word)])).resize()
+    }
+
+    /// Create a [`Int`] from an `i64` (const-friendly)
+    // TODO(tarcieri): replace with `const impl From<i64>` when stable
+    #[cfg(target_pointer_width = "32")]
+    pub const fn from_i64(n: i64) -> Self {
+        Int::new_from_uint(Uint::<{ I64::LIMBS }>::from_u64(n as u64)).resize()
```
lleoha, line 31: this should work and is a little simpler:
Int::new_from_uint(Uint::from_u64(n as u 64))

lleoha:
@tarcieri @erik-3milabs 
What do you think about introducing Positive, NonPositive, Negative, NonNegative newtype structs similarly to Zero, NonZero, Odd etc.?
Also I was thinking about introducing Signed and Unsigned traits (not sure if these would be a'la marker traits) or should contains any methods.

src/int/div.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// assert_eq!(
+    ///     I128::from(8).checked_div_floor(&I128::from(-3)).unwrap(),
+    ///     I128::from(-3)
+    /// );
+    /// assert_eq!(
+    ///     I128::from(-8).checked_div_floor(&I128::from(-3)).unwrap(),
+    ///     I128::from(2)
+    /// )
+    /// ```
+    pub fn checked_div_floor(&self, rhs: &Self) -> CtOption<Self> {
```
lleoha, line 67: do we need this method?
I think one method that rounds towards zero it sufficient (so it works like regular primitive int division).

tarcieri:
@lleoha the wrappers idea sounds possibly good (although I'm unclear on use cases), however something probably best tackled after getting this PR over the finish line

src/int/div.rs
```diff
@@ -0,0 +1,314 @@
...
+use subtle::{Choice, CtOption};
+
+use crate::{CheckedDiv, ConstChoice, Int, NonZero, Uint};
+
+impl<const LIMBS: usize> Int<LIMBS> {
+    #[inline]
+    /// Base checked_div_mod operation.
+    /// Given `(a, b)` computes the quotient and remainder of the absolute
+    /// Note: this operation rounds towards zero, truncating any fractional part of the exact result.
+    fn checked_div_mod(
```
lleoha, line 14: can we instead have a method that is consistent with primitive div/rem i.e. return quotient and remainder with quotient round towards and both being signed integer as they both can be negative e.g. -5 / 2 = (-2, -1).
Also to be consistent with Uint can we rename it to div_rem instead of div_mod?

lleoha: Sorry, I was a little chaotic,
What I'm trying to say is that returning two Uints and one bit is not enough to cover all the cases:
```
-5/-2 = (2, -1)
-5/2 = (-2, -1)
5/-2 = (-2, 1)
5/2 = (2, 1)
```

src/int/from.rs
erik-3milabs: I don't think that would work. Let me explain why:

In scaling up a negative integer, we have to make sure that all new limbs are initialized with `Limb::MAX` instead of `Limb::ZERO` (as is the default for `Uint`). So, what I'm doing here is:
- cast `i64` to `u64`,
- ingest this `u64` into a `Uint` that is as small as possible,
- wrap the `Uint` into an `Int`, and
- scale up the `int` (which I wrote to make sure it uses the right `Limb` type).

IIUC, `Int::new_from_uint(Uint::from_u64(n as u 64))` does
- cast `i64` to `u64`,
- ingest this `u64` into a `Uint<LIMBS>`, filling any missing limbs with `Limb::ZERO`, and
- wrapping the `Uint` into an `Int`.

This would not work for negative values.

Would you agree?

src/int/div.rs
erik-3milabs: Do _we_ need it? No, probably not. Do _I_ need it? Yes, yes I do :)

I need this operation as well as `div_mod_floor` for a project I'm working on (that op is one of the reasons I'm implementing this right now :see_no_evil:). IIUC, the definition I am using here aligns with `num_bigint::BigInt::div_mod_floor`, which I'm trying to replace to make my algorithm constant time.

This function, of course, differs slightly from  `Int::checked_div` in that it rounds down, instead of to zero.

src/int/div.rs
erik-3milabs: Ahh, didn't think of that. Let me fix this later today!

erik-3milabs:
Failing CI should be fixed by https://github.com/erik-3milabs/crypto-bigint/pull/1 which is slated to be merged.

src/int/div.rs
erik-3milabs: I updated the code. Let me know what you think!

erik-3milabs:
Ok, this PR has gotten out of hand... Oops...

With all checks passing, I'm proposing to freeze it here; I'll create a new PR for new features I intend to introduce.

erik-3milabs:
@tarcieri, could you please re-review this and let me know what still needs to be done to merge this?

src/int/div.rs
lleoha: @erik-3milabs out of curiosity, if this is not a secret, what do you need the `floor_div` for?

src/int/from.rs
lleoha: ahh, right,
I forgot that you can call from_i64 to Int with more than two limbs :)

src/int/div.rs
lleoha: LGTM

src/int/div.rs
erik-3milabs: I need `Int` support to implement class-group cryptography. In one or two of the sub-algorithms, I specifically need floor-div.

lleoha: This is the reason I am asking because you mentioned CL before.
I have implemented Class Groups crypto (BICYCL paper and original Castagnos Laguillaumie paper) and didn't need this method so this got me thinking maybe I am doing something wrong :/.
BTW. Elements of group in my impl are represented in both binary quadratic forms and ideals (for compactness).

erik-3milabs: Also, I'm curious: what did you do with the `xgcd` s in the algorithms? This repository does not provide an implementation for that...

lleoha: @erik-3milabs nice :).
As for gcdx I don't have a const time implementation yet, but I was hoping to use Berstein Yung for that. If it can be done for polynomials I so no reason why it wouldn't work for integers. But fair enough gcdx is still not constant time in my implementation :(.

lleoha: Ohh, I also believe that binary extended gcd can be adopted to be const time.

lleoha: Take a look at iacr 2020 972.
I plan to try do that constant time I Berstein Yang redness unusable for some reasons.

.gitignore
```diff
@@ -1,2 +1,4 @@
 target
 proptest-regressions
+
+.idea/
```
tarcieri, line 4: Speaking as a fellow IntelliJ user, these sorts of editor-specific ignore directives are best kept in a local `excludesfile` so everyone doesn't have to add these to every repo they interact with

src/int/cmp.rs
```diff
@@ -0,0 +1,265 @@
+//! [`Int`] comparisons.
+//!
+//! By default, these are all constant-time and use the `subtle` crate.
```
tarcieri, line 3: The `Const*` stuff doesn't use `subtle`. If this comment is copypasta from `uint/cmp.rs`, the comment there is probably out-of-date.

src/int/div.rs
```diff
@@ -0,0 +1,335 @@
...
+use core::ops::Div;
+
+use subtle::{Choice, CtOption};
+
+use crate::{CheckedDiv, ConstChoice, ConstCtOption, Int, NonZero, Uint};
+
+impl<const LIMBS: usize> Int<LIMBS> {
+    #[inline]
+    /// Base div_rem operation.
+    /// Given `(a, b)`, computes the quotient and remainder of their absolute values. Furthermore,
```
tarcieri, line 12:
```suggestion
    /// Base div_rem operation.
    ///
    /// Given `(a, b)`, computes the quotient and remainder of their absolute values. Furthermore,
```

This will look better in the rustdoc, as it doesn't include the explanative comment in one-line summaries.

src/int/div.rs
```diff
@@ -0,0 +1,335 @@
...
+        &self,
+        rhs: &NonZero<Self>,
+    ) -> (Uint<{ LIMBS }>, Uint<{ LIMBS }>, ConstChoice, ConstChoice) {
+        // Step 1: split operands into signs and magnitudes.
+        let (lhs_mag, lhs_sgn) = self.abs_sign();
+        let (rhs_mag, rhs_sgn) = rhs.0.abs_sign();
+
+        // Step 2. Divide magnitudes
+        // safe to unwrap since rhs is NonZero.
+        let (quotient, remainder) = lhs_mag.div_rem(&NonZero::<Uint<LIMBS>>::new_unwrap(rhs_mag));
```
tarcieri, line 24: Perhaps you could define `abs_sign` on `NonZero<Int<LIMBS>>` and have it return `NonZero`?

src/int/encoding.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use crate::{Int, Uint};
+
+impl<const LIMBS: usize> Int<LIMBS> {
+    /// Create a new [`Int`] from the provided big endian bytes.
+    ///
+    /// See [`Uint::from_be_slice`] for more details.
+    pub const fn from_be_slice(bytes: &[u8]) -> Self {
+        Self(Uint::from_be_slice(bytes))
+    }
```
tarcieri, line 11: Should this (and potentially the other methods) handle signed two's complement wire encodings, which are somewhat commonly used? e.g. ASN.1 BER `INTEGER` and SSH `mpint`

src/int/neg.rs
```diff
@@ -0,0 +1,112 @@
...
+impl<const LIMBS: usize> Int<LIMBS> {
+    /// Perform the two's complement "negate" operation on this [`Int`]:
+    /// map `self` to `(self ^ 1111...1111) + 0000...0001` and return the carry.
+    ///
+    /// Note: a zero carry indicates `self == Self::ZERO`.
+    ///
+    /// Warning: this operation is unsafe to use as negation; the negation is incorrect when
+    /// `self == Self::MIN`.
+    #[inline]
+    pub(crate) const fn negc(&self) -> (Self, Word) {
```
tarcieri, line 16: Per the conventions currently used by `core` (see #693), perhaps this should be `carrying_neg`?

```suggestion
    /// Perform the two's complement "negate" operation on this [`Int`]:
    /// map `self` to `(self ^ 1111...1111) + 0000...0001` and return the carry.
    ///
    /// Note: a zero carry indicates `self == Self::ZERO`.
    ///
    /// Warning: this operation is unsafe to use as negation; the negation is incorrect when
    /// `self == Self::MIN`.
    #[inline]
    pub(crate) const fn carrying_neg(&self) -> (Self, Word) {
```

src/uint/neg.rs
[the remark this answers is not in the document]
tarcieri: The subsequent code below looks worryingly similar to the code that was vulnerable here: https://github.com/dalek-cryptography/curve25519-dalek/pull/659

tarcieri:
Some proptests against `num-bigint` that mirror the existing ones for unsigned integers would definitely help instill confidence, though I'd be okay with saving that for a followup PR, given how large this is.

src/int/div.rs
tarcieri: Binary GCD may be of interest in `crypto-primes` as well (although possibly not a constant time implementation) /cc @fjarri @dvdplm

fjarri: We have a somewhat specific case where one argument is always small, so the general one provided by `crypto-bigint` is an overkill. @dvdplm added a binary GCD manually recently in https://github.com/entropyxyz/crypto-primes/pull/54

.gitignore
lleoha: This is me. @erik-3milabs feel free to remove it. I don't want to create a PR just for that.

erik-3milabs: Done.

src/int/cmp.rs
```diff
@@ -0,0 +1,265 @@
...
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
erik-3milabs: Done.

src/int/cmp.rs
erik-3milabs: Done. Fixed it for `uint::cmp` as well.

src/int/div.rs
erik-3milabs: Done.

src/int/div.rs
erik-3milabs: Done.

src/int/neg.rs
erik-3milabs: Done.

erik-3milabs:
@tarcieri Re: proptest
Given the scope of this project, perhaps we should make a tracking PR for these and other missing features?

src/int/encoding.rs
erik-3milabs: Sounds like a good idea. Let me look into this to see what this is about/how it works; I have not worked with encodings before.

src/uint/neg.rs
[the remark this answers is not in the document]
erik-3milabs: Hmm, good point. I'm not yet at a place where I understand how to look into compiler code and determine whether this is vulnerable or not. I, therefore, simplified the implementation to @lleoha's earlier proposal. In my understanding, this should remove the (potential) vulnerability.

src/int.rs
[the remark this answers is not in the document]
erik-3milabs: I renamed `new_from_uint` to `pub(crate) const fn from_bits` and introduced `Uint::as_int`; `Int::as_uint` already existed.

src/int/div.rs
erik-3milabs: @lleoha I was eyeing the 2020/972 paper as well :smirk: Still experimenting with it.

erik-3milabs:
@tarcieri responded to your new and some old remarks. IIUC, all open discussions can be resolved except the one about encoding. I'll need some time to look into that. @lleoha I took the liberty to resolve some of the earlier remarks which I dealt with at the time to clean this PR up a bit.

src/int/neg.rs
```diff
@@ -0,0 +1,113 @@
...
+
+use subtle::CtOption;
+
+use crate::{ConstChoice, Int, Word};
+
+impl<const LIMBS: usize> Int<LIMBS> {
+    /// Perform the two's complement "negate" operation on this [`Int`]:
+    /// map `self` to `(self ^ 1111...1111) + 0000...0001` and return the carry.
+    ///
+    /// Note: a zero carry indicates `self == Self::ZERO`.
```
mkramarczyk, line 11: I don't think it's true. When self is negative and not MIN, the carry would be zero as well, right?

mkramarczyk: When we treat a carry as implicit "`limbs[LIMBS]`" then we would have:
* for positive signed numbers carry is -1 (i.e. 1111...1111)
* for negative (not MIN) carry would be 0 (i.e. 0000...0000)
* for MIN carry is 1 (i.e. 0000....0001).
Not sure if my understanding is inlined with what the carry for negation means, it's more like a convention I guess the same as what carry for add means (as opposed to borrow for example).

mkramarczyk: @tarcieri opinion would be appreciated :)

src/uint/neg.rs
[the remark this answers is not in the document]
tarcieri: FWIW a couple ways of inspecting the generated assembly:

- `cargo-show-asm`
- https://godbolt.org

tarcieri:
> Given the scope of this project, perhaps we should make a tracking PR/issue for these and other missing features?

@erik-3milabs if you make such an issue yourself, you can manage the task list in the description. Otherwise you'll be blocked on me to do updates.

src/int/encoding.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// Create a new [`Int`] from the provided big endian hex string.
+    ///
+    /// Panics if the hex is malformed or not zero-padded accordingly for the size.
+    ///
+    /// See [`Uint::from_be_hex`] for more details.
+    pub const fn from_be_hex(hex: &str) -> Self {
+        Self(Uint::from_be_hex(hex))
+    }
```
tarcieri, line 20: I think this should work for adding two's complement encoding support?

```suggestion
    pub const fn from_be_hex(hex: &str) -> Self {
        Self::from_bits(Uint::from_be_hex(hex))
    }
```

src/int/encoding.rs
tarcieri: I'd suggest doing that because making `from_be_slice` work with two's complement is trickier than `from_be_hex`, since you have to deal with non-`Word`/`Limb`-aligned sizes.

src/uint/neg.rs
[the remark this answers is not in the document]
erik-3milabs: It took me a second, but I was able to spin up `cargo-show-asm`. In release-mode, it cannot find the `wrapping_neg_if` function. In `--dev` mode it does the following:

image

The call to `carrying_neg` is somewhat surprising: their functions are indeed very similar. Is there anything problematic you're seeing about this? Any clue how to get the asm when working with `--release`?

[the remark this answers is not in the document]
erik-3milabs: Note: I'm just curious to learn about all of this, since I enjoy creating _fast_ code.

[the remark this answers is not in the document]
erik-3milabs: nvm, this is `wrapping_neg`, not `wrapping_neg_if`. I cannot find `wrapping_neg_if` at all.

src/int/encoding.rs
erik-3milabs: This wouldn't change things:

```rust
/// Const-friendly [`Int`] constructor.
///
/// Reinterprets the bits of a value of type [`Uint`] as an [`Int`].
/// For a proper conversion, see [`Int::new_from_abs_sign`];
/// the public interface of this function is available at [`Uint::as_int`].
pub(crate) const fn from_bits(value: Uint<LIMBS>) -> Self {
    Self(value)
}
```

erik-3milabs:
@lleoha I found your remark regarding `neg` in my inbox, but cannot find it anywhere in this thread for some reason, so I'm responding here.

You were right: the carry is **non** -zero iff `self = Self::ZERO`. I attempted to fix it in 7688826. Let me know what you think.

src/int/encoding.rs
tarcieri: Aah, right

src/uint/neg.rs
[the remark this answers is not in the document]
tarcieri: I'm surprised by `#[inline(never)]` not working.

Godbolt tends to be a little bit easier for these sorts of things, IMO, but requires you copy enough context into the browser for the code to compile.

I've also meant to investigate `cargo-checkct` which uses BINSEC: https://github.com/Ledger-Donjon/cargo-checkct

tarcieri:
You'll likely need to expand the nearly hundred comments in the middle of the thread, which is all the more reason to get this merged.

I'll try to do another pass soon.

lleoha:
> You were right: the carry is **non** -zero iff `self = Self::ZERO`. I attempted to fix it in [7688826](https://github.com/RustCrypto/crypto-bigint/commit/7688826c7773b85dcc5bef204b074ba1b947d2a7). Let me know what you think.

@erik-3milabs The carry is non-zero also when Int::MIN is being negated, it's Word::ONE in this case. Or at least it should be, right?

erik-3milabs:
@lleoha IIRC, the negation map is $x \to (x \oplus 1111...1111) + 0000...0001$

Applying this to `Int::MIN` $= 1000...0000$ we get

$1000...0000 \oplus 1111...1111 = 0111...1111$
and
$0111...1111 + 0000...0001 = 1000...0000$ with a zero-carry.

What am I missing / what is your reasoning for saying the carry should equal 1?

lleoha:
> @lleoha IIRC, the negation map is x → ( x ⊕ 1111. . .1111 ) + 0000. . .0001
> 
> Applying this to `Int::MIN` = 1000. . .0000 we get
> 
> 1000. . .0000 ⊕ 1111. . .1111 = 0111. . .1111 and 0111. . .1111 + 0000. . .0001 = 1000. . .0000 with a zero-carry.
> 
> What am I missing / what is your reasoning for saying the carry should equal 1?

Let's say the limb is 8 bit. So the min is -128. Then neg(-128) = 128, which can only be represented with two limbs in signed arithmetic.
Also with you logic !(-128) + 1 = 127 + 1. 127 + 1 overflows in signed arithmetics hence carry.

erik-3milabs:
_Fourth attempt to write this._

I investigated this issue a bit, and I think I figured out we were both mistaken.

As it turns out, the devil is in the tiniest of details:
> Also with you[r] logic !(-128) + 1 = 127 + 1. 127 + 1 overflows in signed arithmetics _hence_ carry.

You are correct that that addition _overflows_, but I don't think your "hence" is correct. When operating on signed integers, the `carry` flag is irrelevant. Instead, the `overflow` flag is set. [Source](https://en.wikipedia.org/wiki/Carry_flag#Uses). 

IMU, `carrying_negation` is simply not a thing in signed integer arithmetic. It should be `overflowing_negation`, where the `overflow` flag is raised when an addition overflows. This now also makes more sense: the overflow flag now tells you that negation fails.

For unsigned integers, the whole concept of "negation" is vague anyway. There, I decided to stick with "carrying_neg" and have it _carry_ when "negating" zero, as that seems to make the most sense.

Thank you for persisting! Let me know what you think of the patch I wrote.

erik-3milabs:
@lleoha thanks for the elaboration! I think we are (very close) to (finally :smile:) being on the same page.

For unsigned (`Uint`) arithmetic, I propose we stick with `carrying_neg`; the carry "flag" would be set if and only if you're attempting to negate `Uint::ZERO`.

For signed (`Int`) arithmetic, I am ambivalent about naming the function `extended_neg` or `overflowing_neg`. Perhaps @tarcieri has a deciding opinion on this? Regardless of which is chosen, I completely agree that we should be consistent in using it! As long as the function's "flag" is set when attempting to negate `Int::MIN`, I'm all for it ;-)

tarcieri (approved):

erik-3milabs:
@lleoha if you're interested to stay part of this project, you can have a look at the next PR that is open for review: #697

lleoha:
> @lleoha if you're interested to stay part of this project, you can have a look at the next PR that is open for review: #697

Sure, will do

PR #697, opened 2024-10-22
`Int` - Bit ops

erik-3milabs:
Bit operations on `Int`, in preparation for implementing `Integer` trait.


src/const_choice.rs
```diff
@@ -435,6 +441,26 @@ impl<const LIMBS: usize> ConstCtOption<Odd<Uint<LIMBS>>> {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value is none with a custom panic message provided by
+    /// `msg`.
+    #[inline]
+    pub const fn expect(self, msg: &str) -> Int<LIMBS> {
+        assert!(self.is_some.is_true_vartime(), "{}", msg);
+        self.value
+    }
```
erik-3milabs, line 461: I copy-pasted this from `ConstCtOption<Uint>`.

In doing so, I did notice that, although it uses `is_true_vartime`, `ConstCtOption<Uint>::expect` was not annotated as being vartime. Moreover, it is used for `Uint::shr` and `Uint::shl`, among others, which are not vartime either. Is this OK because unwraps are vartime anyway, or do we need to do something about this @tarcieri ?

erik-3milabs:
@tarcieri this is the next PR ready for review. I am looking forward to your input!

erik-3milabs:
@tarcieri _ping_ :smile:

erik-3milabs:
@lleoha IMHO, I see no added value in supporting logical (=unsigned) right shift on a _signed_ integer: it could be confusing for users, and the difference is quite subtle.

Moreover, when logically shifting-right a _negative_ value, you get some bizarre behaviour. To illustrate with an `i8`:
`-122i8.logical_shr(1) = 60`. It is not even `abs_div_by_2` (which should be `61`). I see no reason to support it, but please let me know if you have a use case in which it does make sense!

Moreover, if for some reason you really need the logical shift right, the solution `x.as_uint().shr(1).as_int()` is still on the table, and it immediately illustrates to the reader something weird is up and that they should pay attention.

lleoha:
> @lleoha IMHO, I see no added value in supporting logical (=unsigned) right shift on a _signed_ integer

I'm sold :).
Let's not add it then.

src/const_choice.rs
tarcieri: They're not really "vartime" in that if the condition is violated, it will crash the entire program, as opposed to completing successfully in a variable number of CPU cycles

tarcieri (approved):

PR #698, opened 2024-10-22
`Int` - gcd

erik-3milabs:
Implement `gcd` for `Int`.

Awaiting #695 and #697


tarcieri:
@erik-3milabs are you going to reopen this one?

erik-3milabs:
Well, I wasn't planning to (just yet), but I might as well. I couldn't reopen this MR, so I replaced it with #748.

PR #699, opened 2024-10-24
`Int` - cross compatibility with `Uint`

erik-3milabs:
Implement
- `Mul<Uint<LIMBS>> for Int<LIMBS>`
- `Div<Uint<LIMBS>> for Int<LIMBS>`


erik-3milabs:
Closed in favor of #726

PR #703, opened 2024-11-19
Faster random_mod 

dvdplm:
@tarcieri I noticed that generating random numbers using `random_mod` is slower than using `random_bits` repeatedly until the random number is below the given bound, like so:

```rust
            let bound = U1024::from_be_hex(b"DEADBEEFDEADBEEFDEADBEEFDEADBEEF");
            let bound_bits = bound.bits_vartime();

            let mut r = U1024::random_bits(&mut rng, bound_bits);
            while r >= bound {
                r = U1024::random_bits(&mut rng, bound_bits);
            }
```
The benchmarks show that `random_mod` is ~4x slower than `random_bits`.

I think the reason it is slow is because how it handles the last limb. The code takes the bit length of the modulus, figures out how many limbs are needed, then fills all limbs except the last limb (“high limb”). This last limb is then randomized up to the number of bytes needed. Say the modulus if 258 bit long, then we need 5 limbs total (320 bits). We fill the first 4 with random data, easy. 

For the last limb we have a single byte to randomize (just 2 bits for my example); the code picks a random number that fits in the relevant bytes and then checks if the resulting uint is smaller than the modulus. The problem is that for moduli that have high bytes expressing very small numbers, the chance is very high that we pick a random number that make the whole uint be bigger than the modulus. 
When this happens, all limbs are thrown out and we do it all over again. Sometimes it has to re-do the whole process more than a hundred times (153 times for the modulus in the example).

This PR addresses this like so:
- Flip the order and start with the high limb and use rejection sampling until it is smaller than the modulus' high limb. Now it's much more likely to make the whole uint be smaller than the modulus.
- Use a mask to filter out the parts of a high limb candidate that we do not care about (i.e. the leading zeros of the modulus)
- Use `next_u64`/`next_u32` instead of `fill_bytes`

With this PR, using `random_mod` is consistently faster than using `random_bits` and much faster than the original code.

Benchmarks comparing `random_mod` (mine), `random_mod` (original) and `random_bits`:

```
bounded random/random_mod, U1024
                        time:   [87.013 ns 87.966 ns 88.918 ns]
                        change: [-0.5835% +0.5767% +1.6716%] (p = 0.32 > 0.05)
bounded random/random_mod, U1024, original
                        time:   [423.46 ns 712.46 ns 1.2063 µs]
                        change: [-17.694% +32.387% +102.40%] (p = 0.33 > 0.05)
Found 12 outliers among 100 measurements (12.00%)
  6 (6.00%) high mild
  6 (6.00%) high severe
bounded random/random_bits, U1024
                        time:   [173.92 ns 183.49 ns 193.78 ns]
                        change: [-7.3823% -1.5494% +4.6582%] (p = 0.61 > 0.05)
bounded random/random_mod, U1024, small bound
                        time:   [23.040 ns 23.803 ns 24.633 ns]
                        change: [-5.9600% -2.3197% +1.4043%] (p = 0.22 > 0.05)
bounded random/random_mod, U1024, small bound, original
                        time:   [67.153 ns 111.79 ns 181.27 ns]
                        change: [-28.850% +14.330% +80.119%] (p = 0.63 > 0.05)
  3 (3.00%) high mild
  5 (5.00%) high severe
bounded random/random_bits, U1024, small bound
                        time:   [25.330 ns 27.807 ns 30.383 ns]
                        change: [-12.220% -3.8687% +5.4944%] (p = 0.40 > 0.05)
bounded random/random_mod, U1024, 512 bit bound low
                        time:   [52.970 ns 53.916 ns 54.920 ns]
                        change: [-1.0751% +0.6102% +2.3705%] (p = 0.50 > 0.05)
bounded random/random_mod, U1024, 512 bit bound low, original
                        time:   [251.80 ns 690.10 ns 1.5121 µs]
                        change: [-11.781% +53.085% +160.33%] (p = 0.28 > 0.05)
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) high mild
  9 (9.00%) high severe
bounded random/random_bits, U1024, 512 bit bound low
                        time:   [94.127 ns 99.391 ns 105.01 ns]
                        change: [-4.6947% +1.6697% +8.0796%] (p = 0.60 > 0.05)
bounded random/random_mod, U1024, 512 bit bound hi
                        time:   [87.048 ns 87.954 ns 88.866 ns]
                        change: [-1.0527% +0.0481% +1.0498%] (p = 0.93 > 0.05)
bounded random/random_mod, U1024, 512 bit bound hi, original
                        time:   [364.47 ns 606.81 ns 905.07 ns]
                        change: [-37.306% +6.4865% +79.619%] (p = 0.83 > 0.05)
Found 12 outliers among 100 measurements (12.00%)
  5 (5.00%) high mild
  7 (7.00%) high severe
bounded random/random_bits, U1024, 512 bit bound hi
                        time:   [182.33 ns 192.86 ns 203.37 ns]
                        change: [-5.0127% +0.8204% +7.0139%] (p = 0.79 > 0.05)
bounded random/random_mod, U1024, tiny high limb
                        time:   [51.661 ns 51.838 ns 52.087 ns]
                        change: [+0.0050% +0.4002% +0.8536%] (p = 0.07 > 0.05)
Found 14 outliers among 100 measurements (14.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  12 (12.00%) high severe
bounded random/random_mod, U1024, tiny high limb, original
                        time:   [4.0430 µs 4.0527 µs 4.0621 µs]
                        change: [-1.0595% -0.5710% -0.0700%] (p = 0.03 < 0.05)
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) low mild
  3 (3.00%) high mild
  3 (3.00%) high severe
bounded random/random_bits, U1024, tiny high limb
                        time:   [75.058 ns 75.214 ns 75.453 ns]
                        change: [-0.1149% +0.1039% +0.3474%] (p = 0.39 > 0.05)
Found 17 outliers among 100 measurements (17.00%)
  7 (7.00%) low mild
  2 (2.00%) high mild
  8 (8.00%) high severe
```


src/modular/monty_form/lincomb.rs
```diff
@@ -29,6 +29,7 @@ impl<const LIMBS: usize> MontyForm<LIMBS> {
 #[cfg(test)]
 mod tests {
     #[cfg(feature = "rand")]
+    #[ignore = "Issue #707"]
```
tarcieri, line 32: Can merge this after #713

tarcieri: #713 is merged now if you can rebase and remove this `ignore`

src/uint/rand.rs
```diff
@@ -108,42 +108,50 @@ pub(super) fn random_mod_core<T>(
     modulus: &NonZero<T>,
     n_bits: u32,
 ) where
-    T: AsMut<[Limb]> + ConstantTimeLess + Zero,
+    T: AsMut<[Limb]> + AsRef<[Limb]> + ConstantTimeLess + Zero + core::fmt::Debug,
```
tarcieri, line 111: What is this bound change for?

dvdplm: You mean `AsRef<[Limb]>`?

It comes from a few lines below:

```rust
let hi_word_modulus = modulus.as_ref().as_ref()[n_limbs - 1].0;
```

Here, `modulus` is a `&NonZero<T>` so I can't use `as_mut()` to get to the limbs. Do you have a better suggestion?

Removed the `Debug` bound, that is leftover debug code.

tarcieri: I actually meant `Debug`

tarcieri (approved):

PR #705, opened 2024-12-02
Fix workflow YAML syntax

AaronFeickert:
My IDE reports that certain branch and path specifications in the GitHub actions configurations are not strictly valid YAML. This PR fixes accordingly.

### DIFF
--- a/.github/workflows/crypto-bigint.yml
+++ b/.github/workflows/crypto-bigint.yml
@@ -5,7 +5,8 @@ on:
     paths-ignore:
       - README.md
   push:
-    branches: master
+    branches:
+      - master
     paths-ignore:
       - README.md
 
--- a/.github/workflows/dudect.yml
+++ b/.github/workflows/dudect.yml
@@ -3,7 +3,8 @@ name: dudect
 
 on:
   push:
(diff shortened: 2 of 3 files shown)


tarcieri:
FWIW we use standard templates for our CI YAML everywhere. This would be a notable divergence.

I guess it's more technically correct? Obviously GitHub Actions promotes a single string to an array of strings or it wouldn't work, though I suppose technically that violates lints which are written to the precise letter of GitHub's schema (which is still somewhat ill-specified given the actual semantics, but I digress)

AaronFeickert:
Feel free to close if the resulting inconsistency with other repositories isn't worth it.

tarcieri:
Maybe we can open some common issue to track updates to the boilerplate we should make everywhere

AaronFeickert:
Not sure where that would be opened. Really this is just because my (apparently very picky) IDE was giving errors on this. Regardless, I'd still advocate to merge it to make the YAML gods happy unless there are likely to be headaches bringing in changes from other repositories.

tarcieri (approved):

tarcieri:
Fine

PR #706, opened 2024-12-02
Add zeroizing support for `MontyParams`

### LINKED ISSUE #704: Zeroizing `MontyParams`
In my application there are `MontyParams` objects created from a secret modulus, so I would like to be able to zeroize them. Currently it is not available, with the comment in `monty_form.rs` `impl Zeroize for MontyForm` saying "NOTE: this does _not_ zeroize the parameters, in order to maintain some form of type consistency". I assume it's because `BoxedMontyForm` uses an `Arc` for parameters, and it's unclear how to zeroize in that case? 

What should we do here? Could we, for example, have an optional `zeroize_on_drop` argument when creating Montgomery parameters, which will be kept in the `Params` structure and checked in the `Drop` implementation? To preserve compatibility with the cases where `zeroize` feature is not enabled, `MontyParams` can have the normal `new_vartime()` method, and a `zeroize`-only `new_vartime_secret()` or something?

AaronFeickert:
After discussion in #704, it seemed prudent to add zeroizing support for `MontyParams`, and to use this when zeroizing `MontyForm` as well. This PR adds such support.

Closes #704.

### DIFF
--- a/src/modular/monty_form.rs
+++ b/src/modular/monty_form.rs
@@ -158,6 +158,18 @@ impl<const LIMBS: usize> ConstantTimeEq for MontyParams<LIMBS> {
     }
 }
 
+#[cfg(feature = "zeroize")]
+impl<const LIMBS: usize> zeroize::Zeroize for MontyParams<LIMBS> {
+    fn zeroize(&mut self) {
+        self.modulus.zeroize();
+        self.one.zeroize();
+        self.r2.zeroize();
+        self.r3.zeroize();
+        self.mod_neg_inv.zeroize();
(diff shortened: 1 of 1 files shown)


tarcieri (approved):

fjarri:
Thanks for the PR! Perhaps it's also worth mentioning in the docs that the params are not zeroized for the boxed version.

AaronFeickert:
> Thanks for the PR! Perhaps it's also worth mentioning in the docs that the params are not zeroized for the boxed version.

Neither `BoxedMontyParams` nor `BoxedMontyForm` currently implement `Zeroize` at all. But they could.

PR #710, opened 2024-12-07
Require `RngCore` instead of `CryptoRngCore` for various random methods

### LINKED ISSUE #137: Is there a reason `random()` and `random_mod()` require `CryptoRng`?
I would argue that it's the user's decision whether they want a cryptographically safe RNG used or not; there can be applications which do not require that restriction, and `CryptoRng` is just a marker trait and does not provide any additional functionality.

fjarri:
Relaxes `CryptoRngCore` requirement to `RngCore`. Fixes #137

### DIFF
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -232,7 +232,6 @@ impl<T> Random for NonZero<T>
 where
     T: Random + Zero,
 {
-    /// Generate a random `NonZero<T>`.
     fn random(mut rng: &mut impl RngCore) -> Self {
         // Use rejection sampling to eliminate zero values.
         // While this method isn't constant-time, the attacker shouldn't learn
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -12,7 +12,7 @@ use subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
 use crate::{ArrayEncoding, ByteArray};
 
 #[cfg(feature = "rand_core")]
-use {crate::Random, rand_core::CryptoRngCore};
+use {crate::Random, rand_core::RngCore};
 
 #[cfg(feature = "serde")]
 use serdect::serde::{
@@ -233,7 +233,7 @@ where
     T: Random + Zero,
 {
     /// Generate a random `NonZero<T>`.
-    fn random(mut rng: &mut impl CryptoRngCore) -> Self {
+    fn random(mut rng: &mut impl RngCore) -> Self {
         // Use rejection sampling to eliminate zero values.
         // While this method isn't constant-time, the attacker shouldn't learn
         // anything about unrelated outputs so long as `rng` is a CSRNG.
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -282,7 +282,9 @@ pub trait Constants: ConstZero {
 /// Random number generation support.
 #[cfg(feature = "rand_core")]
 pub trait Random: Sized {
-    /// Generate a cryptographically secure random value.
+    /// Generate a random value.
+    ///
+    /// If `rng` is a CSRNG, the generation is cryptographically secure as well.
     fn random(rng: &mut impl RngCore) -> Self;
 }
 
@@ -343,19 +345,21 @@ impl core::error::Error for RandomBitsError {}
 /// Random bits generation support.
 #[cfg(feature = "rand_core")]
 pub trait RandomBits: Sized {
-    /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`.
+    /// Generate a random value in range `[0, 2^bit_length)`.
     ///
     /// A wrapper for [`RandomBits::try_random_bits`] that panics on error.
     fn random_bits(rng: &mut impl RngCore, bit_length: u32) -> Self {
         Self::try_random_bits(rng, bit_length).expect("try_random_bits() failed")
     }
 
-    /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`.
+    /// Generate a random value in range `[0, 2^bit_length)`.
     ///
     /// This method is variable time wrt `bit_length`.
+    ///
+    /// If `rng` is a CSRNG, the generation is cryptographically secure as well.
     fn try_random_bits(rng: &mut impl RngCore, bit_length: u32) -> Result<Self, RandomBitsError>;
 
-    /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`,
+    /// Generate a random value in range `[0, 2^bit_length)`,
     /// returning an integer with the closest available size to `bits_precision`
     /// (if the implementing type supports runtime sizing).
     ///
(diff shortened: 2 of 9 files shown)


src/non_zero.rs
```diff
@@ -233,7 +233,7 @@ where
     T: Random + Zero,
 {
     /// Generate a random `NonZero<T>`.
-    fn random(mut rng: &mut impl CryptoRngCore) -> Self {
+    fn random(mut rng: &mut impl RngCore) -> Self {
```
AaronFeickert, line 236: It may be useful to move aspects of the internal comment to the doc comment now that a CSPRNG is no longer guaranteed.

src/traits.rs
```diff
@@ -283,7 +283,7 @@ pub trait Constants: ConstZero {
 #[cfg(feature = "rand_core")]
 pub trait Random: Sized {
     /// Generate a cryptographically secure random value.
-    fn random(rng: &mut impl CryptoRngCore) -> Self;
+    fn random(rng: &mut impl RngCore) -> Self;
```
AaronFeickert, line 286: The doc comment is no longer accurate.

src/traits.rs
```diff
@@ -346,25 +346,22 @@ pub trait RandomBits: Sized {
     /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`.
     ///
     /// A wrapper for [`RandomBits::try_random_bits`] that panics on error.
-    fn random_bits(rng: &mut impl CryptoRngCore, bit_length: u32) -> Self {
+    fn random_bits(rng: &mut impl RngCore, bit_length: u32) -> Self {
```
AaronFeickert, line 349: The doc comment is no longer accurate.

src/traits.rs
```diff
@@ -346,25 +346,22 @@ pub trait RandomBits: Sized {
...
     }
 
     /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`.
     ///
     /// This method is variable time wrt `bit_length`.
-    fn try_random_bits(
-        rng: &mut impl CryptoRngCore,
-        bit_length: u32,
-    ) -> Result<Self, RandomBitsError>;
+    fn try_random_bits(rng: &mut impl RngCore, bit_length: u32) -> Result<Self, RandomBitsError>;
```
AaronFeickert, line 356: The doc comment is no longer accurate.

src/traits.rs
```diff
@@ -346,25 +346,22 @@ pub trait RandomBits: Sized {
...
 
     /// Generate a cryptographically secure random value in range `[0, 2^bit_length)`,
     /// returning an integer with the closest available size to `bits_precision`
     /// (if the implementing type supports runtime sizing).
     ///
     /// A wrapper for [`RandomBits::try_random_bits_with_precision`] that panics on error.
     fn random_bits_with_precision(
-        rng: &mut impl CryptoRngCore,
+        rng: &mut impl RngCore,
```
AaronFeickert, line 364: The doc comment is no longer accurate.

src/traits.rs
```diff
@@ -398,7 +395,7 @@ pub trait RandomMod: Sized + Zero {
     /// issue so long as the underlying random number generator is truly a
     /// CSRNG, where previous outputs are unrelated to subsequent
     /// outputs and do not reveal information about the RNG's internal state.
-    fn random_mod(rng: &mut impl CryptoRngCore, modulus: &NonZero<Self>) -> Self;
+    fn random_mod(rng: &mut impl RngCore, modulus: &NonZero<Self>) -> Self;
```
AaronFeickert, line 398: I would be very careful about this doc comment, as the user may not be aware of what constitutes a CSPRNG.

AaronFeickert: At any rate, the top-line comment is no longer accurate.

src/uint/boxed/rand.rs
```diff
@@ -43,7 +40,7 @@ impl RandomMod for BoxedUint {
     /// The variable-time nature of the algorithm should not pose a security issue so long as the
     /// underlying random number generator is truly a CSRNG, where previous outputs are unrelated to
     /// subsequent outputs and do not reveal information about the RNG's internal state.
-    fn random_mod(rng: &mut impl CryptoRngCore, modulus: &NonZero<Self>) -> Self {
+    fn random_mod(rng: &mut impl RngCore, modulus: &NonZero<Self>) -> Self {
```
AaronFeickert, line 43: I would be very careful about this doc comment, as the user may not be aware of what constitutes a CSPRNG.

AaronFeickert: At any rate, the top-line comment is no longer accurate.

src/uint/rand.rs
```diff
@@ -2,12 +2,12 @@
...
 use super::{Uint, Word};
 use crate::{Encoding, Limb, NonZero, Random, RandomBits, RandomBitsError, RandomMod, Zero};
-use rand_core::CryptoRngCore;
+use rand_core::RngCore;
 use subtle::ConstantTimeLess;
 
 impl<const LIMBS: usize> Random for Uint<LIMBS> {
     /// Generate a cryptographically secure random [`Uint`].
-    fn random(mut rng: &mut impl CryptoRngCore) -> Self {
+    fn random(mut rng: &mut impl RngCore) -> Self {
```
AaronFeickert, line 10: The doc comment is no longer accurate.

src/uint/rand.rs
```diff
@@ -93,7 +90,7 @@ impl<const LIMBS: usize> RandomMod for Uint<LIMBS> {
     /// issue so long as the underlying random number generator is truly a
     /// CSRNG, where previous outputs are unrelated to subsequent
     /// outputs and do not reveal information about the RNG's internal state.
-    fn random_mod(rng: &mut impl CryptoRngCore, modulus: &NonZero<Self>) -> Self {
+    fn random_mod(rng: &mut impl RngCore, modulus: &NonZero<Self>) -> Self {
```
AaronFeickert, line 93: See earlier comments.

AaronFeickert (changes requested):
There are many doc comments that reference cryptographically-secure generation, and are no longer accurate with the relaxed trait bound.

fjarri:
Thanks for checking, that's what happens when you make a PR late at night

src/non_zero.rs
fjarri: I removed this doc comment and allowed the `Random::random()` one take over.

src/traits.rs
fjarri: Added a note

src/traits.rs
fjarri: Added notes for the methods in this trait

tarcieri (approved):

fjarri:
I was in the process of adding a little more stuff to the docs... sorry, should have switched it to the draft form

fjarri:
Basically just the additional lines
```
/// To find a CSRNG, look for RNGs implementing the marker trait [`rand_core::CryptoRngCore`].
```
Maybe it would be too much hand-holding anyway.

tarcieri:
That seems okay if you want to submit a followup

AaronFeickert:
Does this change mean that random number generation functions using rejection sampling should be explicitly tagged with a `_vartime` suffix? Previously, their variable-time nature was guaranteed not to leak anything about output values. This is no longer a guarantee given the relaxed trait bound. The alternative would be to retain the existing naming, but update the documentation to carefully note this.

tarcieri:
Ugh, I mean it was always "vartime". I agree that now there are more footguns.

AaronFeickert:
Made some documentation updates in #711 toward this.

PR #713, opened 2024-12-10
Fix modulus leading zeros calculation for (Boxed)`MontyForm`

### LINKED ISSUE #707: lincomb: unexpected failure at index 1309
It seems like [this test](https://github.com/RustCrypto/crypto-bigint/blob/5cab3fa67040f63f82d36a8e4112a3488d3b02aa/src/modular/monty_form/lincomb.rs#L33) fails at a certain point. The current number of loops, 1000, all work. But if you continue checking, say with 10k iterations, it starts to fail. With the seed `1` I get a failure on iteration 1309; with seed `2` it gets to 1705; with `123` all the way to 3787...

Not sure what is going on tbh, but possibly it's an off-by-one problem:

```
assertion `left == right` failed: n=1309, a=6720607A13DC211AFFD8784C89EE1A9AB5109FAC6F9F4FF62385E13BD7E72CDC, b=07972E2B2F380A1C88DA269D8F8BBE80996DA24B701BC69663CFC1720DFB90DE, c=0D5B579740D5DA9326B4CC9C4095382F2549B30657676C3D793B49A44E2F38E7, d=83BBDD566DCDB03EB1A522D35A005ED5AF14BC7CD8FFEFF2A5BE24424377945C
  left: Uint(0x4A64CD5E88E9411302A59BE48EBF856284231F0E37F12D676C29484DDD5B44BE)
 right: Uint(0x4A64CD5E88E9411302A59BE48EBF856284231F0E37F12D676C29484DDD5B44BD)
```

or

```
assertion `left == right` failed: n=1705, a=C634C58CC0098EB4B8E75B0CEF4BB547AF0175A450BFA0952F062CF7DD2492A6, b=6998ED576B47B8D77FBE1E62847FB9326D1A1B7C134F4B39161CFB51C73780B5, c=AFE47096A3C51D7280CF43610ADC68D413059922709AB97326FA005F68A10DB4, d=B9341A8943CDCBCBB53F4A222817C5B6D4808101443A25131A842F90040741BB
  left: Uint(0x0BA83BFE9A86FFCC193403EE7F21DF76C47465EDAC403DA0982640CCF3079A38)
 right: Uint(0x0BA83BFE9A86FFCC193403EE7F21DF76C47465EDAC403DA0982640CCF3079A37)
```

This is on `master` branch. I ran into this working on unrelated code that caused the test to fail at index 975.

andrewwhitehead:
Fixes #707

### DIFF
--- a/src/modular/monty_form/lincomb.rs
+++ b/src/modular/monty_form/lincomb.rs
@@ -39,7 +39,7 @@ mod tests {
         use rand_core::SeedableRng;
 
         let mut rng = rand_chacha::ChaCha8Rng::seed_from_u64(1);
-        for n in 0..1000 {
+        for n in 0..1500 {
             let modulus = Odd::<U256>::random(&mut rng);
             let params = MontyParams::new_vartime(modulus);
             let m = modulus.as_nz_ref();
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -97,7 +97,7 @@ impl BoxedMontyParams {
 
         let mod_neg_inv = Limb(Word::MIN.wrapping_sub(inv_mod_limb.limbs[0].0));
 
-        let mod_leading_zeros = modulus.as_ref().leading_zeros().max(Word::BITS - 1);
+        let mod_leading_zeros = modulus.as_ref().leading_zeros().min(Word::BITS - 1);
 
         let r3 = montgomery_reduction_boxed(&mut r2.square(), &modulus, mod_neg_inv);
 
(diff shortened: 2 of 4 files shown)


src/modular/monty_form/lincomb.rs
```diff
@@ -39,7 +39,7 @@ mod tests {
         use rand_core::SeedableRng;
 
         let mut rng = rand_chacha::ChaCha8Rng::seed_from_u64(1);
-        for n in 0..1000 {
+        for n in 0..1500 {
```
fjarri, line 42: Probably would be more reliable to have a specialized regression test. And a proptest instead of iterating a 1000 times.

Issue #700, opened 2024-10-29
`Int` feature tracking ticket


erik-3milabs:
### Related
  - #1 (feature request), and 
  - #624 (initial discussion).
  - #694 (abandoned attempt involving sign-magnitude representation),

### Features
- [x] `add`, `sub`, `mul`, `div` #695
- [x] `and`, `or`, `xor`, `not`, `shl`, `shr` #697
- [x] `Int::square` operations #734 
- [x] `vartime` division #735
- [x] `mul_uint`, `div_uint` #736 
- [ ] Encoding support (`from_be_slice`, `from_le_slice`)
- [x] `inv_mod` #737 
- [x] `gcd` #698   
- [ ] `xgcd`
- [ ] prop-tests (see also https://github.com/RustCrypto/crypto-bigint/pull/695#issuecomment-2438256609)

ycscaly:
@tarcieri with #695 merged to signed-int, whats the plan for this feature? Specifically, will it be included in the next release? Could be helpful esp. because of  comparability other crates like elliptic-curves depending on crypto-bigint

tarcieri:
@ycscaly yeah, we should be able to get at least a preliminary implementation into the release and then we can additively improve it.

I went ahead and added it to the list in #681

tarcieri:
I'm planning on trying to wrap up `crypto-bigint` v0.6.0 stable, and I'd like to save this for `crypto-bigint` v0.7, FYI.

Fortunately, all work is happening on the [`signed-int`](https://github.com/RustCrypto/crypto-bigint/tree/signed-int) branch.

ycscaly:
> I'm planning on trying to wrap up `crypto-bigint` v0.6.0 stable, and I'd like to save this for `crypto-bigint` v0.7, FYI.
> 
> Fortunately, all work is happening on the [`signed-int`](https://github.com/RustCrypto/crypto-bigint/tree/signed-int) branch.

Why can't we have the signed integers on 0.6? There's quite a bit already implemented here, and we are actively using it.
We are planning to lunch our product very soon, and would be better to have this as part of a release if possible.

Thanks!

tarcieri:
Aah, well then. We can potentially merge the `signed-int` branch, although there are still a few outstanding PRs and work seems to have stalled.

If it doesn't make it in, we'll be able to merge it in a v0.7 release, which I would expect some time in Q2 2025.

fjarri:
What breaking changes do signed integers introduce?

[...]

PR #721, opened 2025-01-07
`Int` - sync `signed-int` branch

erik-3milabs:


tarcieri:
Is this just merging `master`? If so, perhaps we could go the other way around and merge `signed-int` into `master`

erik-3milabs:
> Is this just merging `master`? If so, perhaps we could go the other way around and merge `signed-int` into `master`

I was of the impression that that was off the table :sweat_smile:

erik-3milabs:
but yes, it is just merging `master` into `signed-int`.

tarcieri:
We were just discussing it here actually:

https://github.com/RustCrypto/crypto-bigint/issues/700#issuecomment-2574073756

Note that the goal is a 0.6 release which would effectively freeze it in whatever form we ship, though we'll probably do a followup 0.7 shortly after that.

erik-3milabs:
`signed-int` seems to add features only (but it is hard to verify with `master` and  `signed-int` out of sync).

My `int-dev` branch still has a bunch of stuff related to interoperability between `Uint` and `Int` that makes `Int` a lot more useful. IIRC, these changes are mostly additive, but some will change the interface to some `Int` functions a bit, so, ideally they are included before we launch 0.6.0. However, I'm having a hard time determining exactly which functions will change with `signed-int` and `master` out of sync.

Do you think we could sync `signed-int` and `master`? That way I can create a (hopefully tiny) PR with all the interface changes I'd ideally include in 0.6.

erik-3milabs:
Closed in favor of #723

PR #727, opened 2025-01-08
`Int` - `inv_mod`

erik-3milabs:
Introduces the `Int::inv_mod` operation.

Status: awaiting the merge of #722.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1,26 +1,26 @@
 //! Traits provided by this crate
 
-mod sealed;
-
-pub use num_traits::{
-    ConstZero, WrappingAdd, WrappingMul, WrappingNeg, WrappingShl, WrappingShr, WrappingSub,
-};
-
-pub(crate) use sealed::PrecomputeInverterWithAdjuster;
-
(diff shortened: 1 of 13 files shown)


src/traits.rs
```diff
@@ -443,9 +443,9 @@ pub trait MulMod<Rhs = Self> {
 }
 
 /// Compute `1 / self mod p`.
-pub trait InvMod: Sized {
+pub trait InvMod<Rhs = Self, Output = Self>: Sized {
     /// Compute `1 / self mod p`.
-    fn inv_mod(&self, p: &Self) -> CtOption<Self>;
+    fn inv_mod(&self, p: &Rhs) -> CtOption<Output>;
 }
```
erik-3milabs, line 449: This solution should (in theory?) just be an addition, not a breaking change.

However, perhaps the following solution might be neater:
```rs
pub trait InvMod: Sized {

    type Modulus;
    type Output;

    /// Compute `1 / self mod p`.
    fn inv_mod(&self, p: &Self::Modulus) -> Self::Output;
}
```

PR #731, opened 2025-01-09
Variable-time modular inversion support

### LINKED ISSUE #728: Add an invert_vartime method to ConstMontyForm
Now that the inv method is constant time, it could be useful to have a vartime alternative.

This issue is a feature request for a method like this:

```rust
let x: ConstMontyForm<MOD, SAT_LIMBS> = ...
x.invert_vartime()
```

that computes the inverse of x in the field of cardinal MOD, constant time in regards x but vartime in regards to MOD.

tarcieri:
Adds (back) support for computing modular inversions in variable-time with respect to the value being inverted, which computes the specific number of safegcd divsteps to perform based on the input, as opposed to using a worst case number based on the bit length.

Closes #728

### DIFF
--- a/src/modular/boxed_monty_form/inv.rs
+++ b/src/modular/boxed_monty_form/inv.rs
@@ -10,18 +10,33 @@ use core::fmt;
 use subtle::CtOption;
 
 impl BoxedMontyForm {
-    /// Computes `self^-1` representing the multiplicative inverse of `self`.
-    /// I.e. `self * self^-1 = 1`.
+    /// Computes `self^-1` representing the multiplicative inverse of `self`,
+    /// i.e. `self * self^-1 = 1`.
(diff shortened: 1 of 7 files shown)


tarcieri:
This is unfortunately quite a bit of boilerplate. I'm not sure if there's any refactoring we could do to potentially reduce it. Abstracting against the boxed and `const fn`-capable unboxed versions is quite difficult.

tarcieri:
Note: wanted to add some equivalence tests between the two versions prior to merging

tarcieri:
It seems like for each of `ConstMontyForm`, `MontyForm`, and `BoxedMontyForm` it might be possible to DRY out `gcd`/`gcd_vartime`/`inv(ert)`/`inv(ert)_vartime` with some conversion helpers that all four could call (though they would be slightly different for each form)

PR #734, opened 2025-01-10
`Int` - square operations

erik-3milabs:
- Expand the `Uint::square` operation set 
- Implement `Int::square` operations. 

Replaces #724.


erik-3milabs:
@tarcieri please have a look :)

tarcieri (approved):

PR #737, opened 2025-01-10
`Int` - `inv_mod`

erik-3milabs:
Introduces the `Int::inv_mod` operation. 

Replaces #727

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -20,6 +20,7 @@ mod cmp;
 mod div;
 mod encoding;
 mod from;
+mod inv_mod;
 mod mul;
 mod neg;
 mod resize;
--- a/src/int/inv_mod.rs
+++ b/src/int/inv_mod.rs
@@ -32,7 +32,7 @@ where
 
     fn inv_mod(&self, modulus: &NonZero<Uint<LIMBS>>) -> CtOption<Self::Output> {
         let (abs, sgn) = self.abs_sign();
-        let abs_inv = abs.inv_mod(modulus).into();
+        let abs_inv = abs.inv_mod(modulus);
 
         // Note: when `self` is negative and modulus is non-zero, then
         // self^{-1} % modulus = modulus - |self|^{-1} % modulus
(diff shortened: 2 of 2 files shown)


src/traits.rs
```diff
@@ -443,9 +443,9 @@ pub trait MulMod<Rhs = Self> {
 }
 
 /// Compute `1 / self mod p`.
-pub trait InvMod: Sized {
+pub trait InvMod<Rhs = Self, Output = Self>: Sized {
     /// Compute `1 / self mod p`.
-    fn inv_mod(&self, p: &Self) -> CtOption<Self>;
+    fn inv_mod(&self, p: &Rhs) -> CtOption<Output>;
 }
```
erik-3milabs, line 449: @tarcieri this is the non-breaking type approach for this trait. I think a more readable solution would be to add the `type Modulus;` and `type Output;` as trait types. Wdyt?

tarcieri: making `Output` an associated type makes sense, however `Rhs` is typically a generic parameter in core arithmetic traits, which allows overlapping impls potentially (though I'm not sure that would be of use here)

erik-3milabs: IIUC, you're proposing the following:

```rs
pub trait InvMod<Rhs = Self>: Sized {
    
    type Output;
    
    /// Compute `1 / self mod p`.
    fn inv_mod(&self, p: &Rhs) -> CtOption<Output>;
}
```

Is that correct?

tarcieri: Yes

tarcieri: Went ahead and impl'd this in #741 if you can rebase

erik-3milabs: Brilliant, thanks.

erik-3milabs:
@tarcieri, I rebased and think this PR is ready to be merged. Please have a look. 

Note: I could not make `Int::odd_inv_mod` a `const` function since I don't see how I can manipulate a `ConstCtOption` to apply a `wrapping_sub` operation. Please let me know if there is any way so I can adjust things appropriately.

tarcieri (approved):
Will merge this for now with potential followup for making it `const fn`

tarcieri:
> I don't see how I can manipulate a ConstCtOption to apply a wrapping_sub operation.

You'd need to add a `map` operation to `ConstCtOption`

PR #746, opened 2025-01-20
Add a RemMixed trait and impl for Uints

### LINKED ISSUE #739: Access to `rem_wide_vartime` through a trait
I have recently found myself in a situation where I want to reduce an `Integer` of one size by another `Integer` of a different size. There is a [`rem_wide_vartime()`]() method on `Uint` that fits the bill, but there is no way to call it when using traits.

One way to solve this would be adding a new public trait that expose the method, e.g.:

```rust
pub trait RemWide<Rhs = Self>: Sized {
    type Output;
    fn rem_wide_vartime(lower_upper: (Rhs, Rhs), rhs: &NonZero<Rhs>) -> Self::Output;
}
```

That would allow this test to pass:

```rust
    #[test]
    fn rem_wide_vartime_on_trait() {
        struct A<T, U> {
            x: T,
            y: U,
        }
        impl<T, U> A<T, U>
        where
            T: Clone + Zero + RemWide<Output = T>,
            U: Split<Output = T>,
        {
            fn reduce_u_by_t(&self) -> T {
                let rhs = &NonZero::new(self.x.clone()).unwrap();
                T::rem_wide_vartime(self.y.split(), rhs)
            }
        }

        let a = A {
            x: U64::from(123u64),
            y: U128::from(456u64),
        };
        assert_eq!(a.reduce_u_by_t(), U64::from(87u64));
    }
```

1. Is there a better way to do the above that I have missed?
2. If not, would you be open to adding a `RemWide` trait (or variant thereof)?
3. The above example works for types that are exactly twice the size, but wouldn't help for, say, a `U1024` reduced by a `U128`. Can you think of any way to achieve that?

dvdplm:
Add a `RemMixed` trait in the same vein as `SplitMixed` and `ConcatMixed`, to calculate the remainder of differently sized integers.

Fixes #739.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -627,7 +627,7 @@ pub trait DivRemLimb: Sized {
 }
 
 /// Support for calculating the remainder of two differently sized integers.
-pub trait RemMixed<Reductor, Tail>: Sized {
+pub trait RemMixed<Reductor>: Sized {
     /// Calculate the remainder of `self` by the `reductor`.
     fn rem_mixed(&self, reductor: &NonZero<Reductor>) -> Reductor;
 }
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -626,6 +626,12 @@ pub trait DivRemLimb: Sized {
     fn div_rem_limb_with_reciprocal(&self, reciprocal: &Reciprocal) -> (Self, Limb);
 }
 
+/// Support for calculating the remainder of two differently sized integers.
+pub trait RemMixed<Reductor, Tail>: Sized {
+    /// Calculate the remainder of `self` by the `reductor`.
+    fn rem_mixed(&self, reductor: &NonZero<Reductor>) -> Reductor;
+}
+
 /// Support for optimized division by a single limb.
 pub trait RemLimb: Sized {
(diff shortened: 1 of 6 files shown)


tarcieri:
This looks fine as a general direction

dvdplm:
What else would you like to see done here, can you give me a rough idea?

tarcieri:
@dvdplm if you think it's ready, can you remove draft?

dvdplm:
In https://github.com/RustCrypto/crypto-bigint/pull/746/commits/f6caf928e38880e8ebac8c164ef15120fd8eebd2 I added an impl for `BoxedUint` but not sure if that actually makes sense.

As a rule, should all traits be impl'd for `Int` as well as `Uint`, or are you handling that on a case-by-case basis?

tarcieri:
FWIW we don't impl other `*Mixed` traits on `Boxed*` types.

Re: trait parity for `Uint` and `Int`, since the latter is something of a relatively new addition and not fully featured yet, we haven't been worrying about that, but it would probably be a good eventual goal.

tarcieri (approved):

PR #753, opened 2025-01-24
Speedup `Uint::shr`

erik-3milabs:
Refactor `Uint::shr`, achieving a >60% speed up for `U2048`.

## Benchmarks (run on commit b2dcb20)

```
right shift/shr, U2048  time:   [292.04 ns 292.16 ns 292.30 ns]
                        change: [-0.2073% -0.0666% +0.0669%] (p = 0.34 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe
right shift/new shr, U2048
                        time:   [114.34 ns 114.48 ns 114.62 ns]
                        change: [+0.8375% +1.0410% +1.2401%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
```

PR #755, opened 2025-01-30
Binary GCD

erik-3milabs:
This PR introduces an implementation of Optimized Binary GCD. Ref: [Pornin, Algorithm 2](https://eprint.iacr.org/2020/972.pdf).

### _Upsides to this technique:_
- it is up to 27x faster than the `gcd` algorithm currently implemented in `crypto_bigint` (see below) (really, it just has a different complexity bound).
- does not need `UNSAT_LIMBS`
- it is actually constant time (unlike the current implementation, which is sneakily vartime in the maximum of the bitsizes of the two operands).

### _Benchmark results_
`Word = u64`

limbs | gcd (vt) | gcd (ct) | new_gcd (ct)
---|---|---|---
2 |  10.687 µs | 20.619 µs | **3.6090 µs**
4 |  29.121 µs | 56.433 µs | **7.1124 µs**
8 |  99.819 µs | 195.02 µs | **16.184 µs**
16 |  359.39 µs | 710.26 µs | **44.294 µs**
32 |  1.6804 ms | 3.3097 ms | **136.49 µs**
64 |  6.9717 ms | 13.028 ms | **494.16 µs**
128 |  29.099 ms | 57.325 ms | **2.3335 ms**
256 |  143.22 ms | 244.89 ms | **8.7722 ms**


erik-3milabs:
@tarcieri To the best of my understanding the code is written in constant time, but I will need some assistance verifying that 
- the compiler really constructs something that executes in constant time, and 
- that a CPU truly executes this in constant time (and does not have some tricks up its sleeve that I don't yet know about...)

src/uint/new_gcd.rs
```diff
@@ -0,0 +1,398 @@
...
+            f10 = Int::select(&f10, &f10.shl_vartime(1), do_apply);
+            f11 = Int::select(&f11, &f11.shl_vartime(1), do_apply);
+            log_upper_bound = do_apply.select_u32(log_upper_bound, log_upper_bound + 1);
+        }
+
+        (IntMatrix([[f00, f10], [f01, f11]]), log_upper_bound)
+    }
+
+    /// Compute the greatest common divisor of `self` and `rhs`.
+    pub const fn new_gcd(&self, rhs: &Self) -> Self {
```
ycscaly, line 274: Same for function names, probably `pornin_gcd` is better.

ycscaly: @tarcieri alt is to completely replace the old gcd, and just name this `gcd`

src/uint/new_gcd.rs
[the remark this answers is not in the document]
erik-3milabs: O, no doubt about that. This is just some code I developed to the point that it can be reviewed. There is no point in discussing naming when it turns out the algorithm is bogus ;-)

src/uint/new_gcd.rs
erik-3milabs: Adding to this: I'm currently working on `xgcd` and should also be able to squeeze out `inv_odd_mod`.

erik-3milabs:
After not noticing the issue in #753, I reran the benchmarks here too. The only thing I noticed was that removing the old vartime benchmark improved the old const-time benchmarks, primarily for `Uint`s with few limbs. The table with the new results is presented below.

Moreover, I ran each benchmark thrice, and observed that
- the `new_gcd` benchmark results are as stable as the old gcd results, and
- the confidence interval of the new results are of the same order of magnitude as the old results.

Based on this, I'm quite confident that the bingcd satisfies the constant time requirement at least as well as the current `gcd` implementation.

limbs | gcd (ct) | new_gcd (ct) | factor |
---|---|---|---
2 | 6.9073 µs | **3.6055 µs** |1.9x 
4 | 42.950 µs | **7.1356 µs** |6.0x
8 | 63.511 µs | **16.244 µs** |3.9x
16 | 633.55 µs | **47.897 µs** |13.2x
32 | 3.2049 ms | **136.21 µs**|23.5x
64 | 12.787 ms | **498.34 µs**|25.7x
128 | 56.710 ms | **2.3501 ms**|24.1x
256 | 244.89 ms | **8.7418 ms** |28.0x

Evidently, this technique predominantly shines for larger Uints. I would not be surprised if, with some tweaks, the bingcd performance for smaller cases could be further improved.

src/uint/new_gcd.rs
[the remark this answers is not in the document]
tarcieri: Why not both?

I would suggest putting this under `src/modular` next to the existing `safegcd.rs`

`bingcd` or `binarygcd`/`binary_gcd` would make sense to me.

src/uint/bingcd/gcd.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl<const LIMBS: usize> Odd<Uint<LIMBS>> {
+    const BITS: u32 = Uint::<LIMBS>::BITS;
+
+    /// Compute the greatest common divisor of `self` and `rhs`.
+    #[inline(always)]
+    pub const fn bingcd(&self, rhs: &Uint<LIMBS>) -> Uint<LIMBS> {
+        // Todo: tweak this threshold
+        if LIMBS < 8 {
```
erik-3milabs, line 31: What technique is commonly used to determine a proper threshold?

erik-3milabs: Benchmarks run on my machine indicate this to be a good spot. But who is to say that the same holds for others / 32-bit architectures?

erik-3milabs:
Update: 
- I refactored the code, improving readability
- I introduced the "small" version of the algorithm that achieves slightly faster results for smaller instances (i.e., uints with fewer limbs) than the original "large" version.

Benchmarks below.
Note:
- The original `gcd` algorithm is unexpectedly slow for 4 limbs. Any clue what is up with that?
- Based on these results, I have `bingcd` select the "small" variant up to 7 limbs and the "large" beyond that. Anything else I can do to improve this?

limbs | gcd | bingcd | factor | bingcd (small) | bingcd (large)
-|-|-|-|-|-
1 | 2.6475 µs | 363.93 ns | 7.3x | 342.17 ns| 1.1385 µs
2 | 6.9068 µs | 1.3533 µs | 5.1x | 1.2378 µs| 3.2917 µs
3 | 12.677 µs | 2.7281 µs | 4.6x | 2.5826 µs| 5.5247 µs
4 | **42.896 µs** | 4.7267 µs | 9.1x | 4.5286 µs| 6.6132 µs
5 | 27.987 µs | 7.1246 µs | 3.9x | 6.8353 µs| 10.100 µs
6 | 38.466 µs | 10.136 µs | 3.8x | 9.9228 µs| 11.977 µs
7 | 49.654 µs | 13.621 µs | 3.6x | 13.614 µs| 14.921 µs
8 | 63.467 µs | 15.749 µs | 4.0x | 17.757 µs| 15.545 µs
--|--|--|--|--|--
16 | 632.80 µs | 44.272 µs | 14.3x | 76.077 µs| 44.593 µs
32 | 3.1699 ms | 137.51 µs | 23.1x | 447.11 µs| 137.29 µs
64 | 12.827 ms | 489.57 µs | 26.2x | 1.7868 ms| 499.83 µs
128 | 56.750 ms | 2.3381 ms | 24.3x | 8.7670 ms| 2.3186 ms
256 | 243.97 ms | 8.7289 ms | 27.9x | 34.823 ms| 8.7274 ms

erik-3milabs:
There is still significant room for improvement for the "large" variant. In each of its `Uint::BITS/62` outer loop iterations, it multiplies `a` and `b` by a `I64` matrix that shrinks their size by at least 62 bits. After iteration X, we thus know that either
- `a` and `b` are at most $BITS - 62X$ bits long, or
- the matrix is the unit matrix.

Assuming the first case, we can thus improve this multiplication by not considering some of the top limbs of `a` and `b`. Over the full algorithm, this would equal a ~1.8x reduction for time spent on those multiplications. 

Aside from these multiplications, the "large" variant is linear in `Uint::BITS`. Interpolating from the benchmarks, the fraction of the time spent on these multiplications is _roughly_ estimated as $$L/(L+6.85)$$ for $L$ limbs, which is already equal to ~70% for L=16 (U1024). The benefit would already be ~~30% for U1024, and increasingly more for larger limb sizes.

This would require the development of a "bounded" multiplication feature. Would this project be interested in that?

erik-3milabs:
@tarcieri @ycscaly, I think this PR has stabilized for review (truly now :see_no_evil:). Looking forward to your reviews!

src/uint/bingcd/xgcd.rs
```diff
@@ -0,0 +1,77 @@
...
+        let mut log_upper_bound = 0;
+        let mut j = 0;
+        while j < iterations {
+            j += 1;
+
+            let b_odd = b.is_odd();
+            let a_gt_b = Uint::gt(&a, &b);
+
+            // swap if b odd and a > b
+            let do_swap = b_odd.and(a_gt_b);
```
dolevmu, line 33: The roles of a and b are switched compared to Algorithm 2. Of course, this should work as GCD(a,b)=GCD(b,a), but this is confusing and make everything hard to follow. Any reason for this? This are local variables introduced in this inner function only.

src/uint/bingcd/xgcd.rs
```diff
@@ -0,0 +1,77 @@
...
+            Uint::conditional_swap(&mut a, &mut b, do_swap);
+            matrix.conditional_swap_rows(do_swap);
+
+            // subtract a from b when b is odd
+            b = Uint::select(&b, &b.wrapping_sub(&a), b_odd);
+            matrix.conditional_subtract_top_row_from_bottom(b_odd);
+
+            // Div b by two and double the top row of the matrix when a, b ≠ 0.
```
dolevmu, line 41: According to Algorithm 2, you multiply by 2 always, but I suppose you know what you are doing and he is wrong on this edge case.

src/uint/bingcd/xgcd.rs
```diff
@@ -0,0 +1,77 @@
...
+            let do_apply = a.is_nonzero().and(b.is_nonzero());
+            b = Uint::select(&b, &b.shr_vartime(1), do_apply);
+            matrix.conditional_double_top_row(do_apply);
+            log_upper_bound = do_apply.select_u32(log_upper_bound, log_upper_bound + 1);
+        }
+
+        (matrix, log_upper_bound)
+    }
+}
```
dolevmu, line 50: The algorithm appears constant time. However, this chunk of code if I understand correctly is the bottleneck inner loop that should be as optimized as possible. I believe the guy who wrote Algorithm 2 did not put enough effort for that matter, and considered swaps comparisons and ifs are free, but in const time this can make a difference. It is scary to do that but worth considering.

src/uint/bingcd/xgcd.rs
dolevmu: Also, it multiplies the second row and not the first one (f1,g1) is multiplied by two.

src/uint/bingcd/xgcd.rs
```diff
@@ -0,0 +1,77 @@
...
+            matrix.conditional_subtract_top_row_from_bottom(b_odd);
+
+            // Div b by two and double the top row of the matrix when a, b ≠ 0.
+            let do_apply = a.is_nonzero().and(b.is_nonzero());
+            b = Uint::select(&b, &b.shr_vartime(1), do_apply);
+            matrix.conditional_double_top_row(do_apply);
+            log_upper_bound = do_apply.select_u32(log_upper_bound, log_upper_bound + 1);
+        }
+
+        (matrix, log_upper_bound)
```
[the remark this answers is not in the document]
dolevmu: I also see you avoid the checks on the sign of a and b (lines 18,20) and conditional negation. Nice.

src/uint/bingcd/xgcd.rs
dolevmu: Concretely, it appears there are 4 cases (where in your code a and b switch roles), depending on whether a<b or a>=b and whether a is odd or even, matrix has four options that can be hard-coded, namely [-1,1;2,0], [1,0;0,2], [1,0;0,2] (if a even second does not matter), and [1,-1;0,2].

I might have gotten wrong the values but you can figure it out, and get M1,M2,M3=M4 hard-coded. Then you should only compute the two flags and set matrix=select(M3, select(M1,M2,flag1), flag2). Which I believe could be faster. On the other hand looks like you can only select between integers and not matrices, so this is 8 selects (and computing 2 flags). I didn't count or evaluate if it is faster overall, but I think this is important to do specifically in this function.

dolevmu: In fact, not only M3=M4 is the same when a is even, when a is odd, M1 and M2 are the swap columns of one another. So actually

Matrix= select(M3, conditional_swap_columns(M1, a<b), a is even). This is therefor 2 boolean flags, 2 conditional swaps, and one matrix conditional select.

src/uint/cmp.rs
```diff
@@ -25,6 +25,12 @@ impl<const LIMBS: usize> Uint<LIMBS> {
         Uint { limbs }
     }
 
+    /// Swap `a` and `b` if `c` is truthy, otherwise, do nothing.
+    #[inline]
+    pub(crate) const fn conditional_swap(a: &mut Self, b: &mut Self, c: ConstChoice) {
```
dolevmu, line 30: In light of my previous comment, it is worth to consider if we can efficiently select between two vectors or matrices, and similarly swap columns/rows efficiently (which you did implement).

src/uint/bingcd/gcd.rs
```diff
@@ -0,0 +1,255 @@
...
+
+            let a_odd = a.is_odd();
+
+            // swap if a odd and a < b
+            let a_lt_b = Uint::lt(&a, &b);
+            let do_swap = a_odd.and(a_lt_b);
+            Uint::conditional_swap(&mut a, &mut b, do_swap);
+
+            // subtract b from a when a is odd
+            a = a.wrapping_sub(&Uint::select(&Uint::ZERO, &b, a_odd));
```
erik-3milabs, line 71: @tarcieri what do you think of this compared to how I had it previously:

```suggestion
            a = Uint::select(&a, &a.wrapping_sub(&b), a_odd);
```

it yields 25-10% improvement for Uints with few limbs

src/uint/bingcd/xgcd.rs
erik-3milabs: Turns out, this is wrong: this leads to buggy behavior for specific pairs of numbers. I'm refactoring some code to bypass this issue.

src/uint/bingcd/gcd.rs
tarcieri: I'm surprised there's that much of a difference. Are you sure it's always faster or is it faster depending on `a`?

erik-3milabs: I'm pretty confident it is always faster.

The reason I think this is faster, is that we are now selecting between a constant and a variable, instead of two variables. Given that `select` is `const` and loves to be inlined, the compiler can now optimize the `select` operation. 

Recall, `Uint::select` calls `Limb::select`, which in turn calls

```rust
impl ConstChoice {
    /// Return `b` if `self` is truthy, otherwise return `a`.
    #[inline]
    pub(crate) const fn select_word(&self, a: Word, b: Word) -> Word {
        a ^ (self.0 & (a ^ b))
    }
}
```

When `a` is the constant ZERO, this can be optimized as:

```rust
        self.0 & b
```

saving two XOR operations, or 2/3's of this operation.

Returning to the `gcd` subroutine, this `select` is in the hot loop of this algorithm. In total, the loop executes:
- `Uint::is_odd` (1 op)
- `Uint::lt` (2 ops/word), 
- `ConstChoice::and` (1 op), 
- `Uint::wrapping_sub` (2 ops/word),
- `Uint::select` (3 ops/word -> 1 ops/word)
- `Uint::shr` (3 ops/word)

So, there is a reduction from 10 to 8 ops/word, or a 20% improvement.

tarcieri: > When a is the constant ZERO

Aah, ok

src/uint/cmp.rs
erik-3milabs: @tarcieri do you have any ideas on improving the performance of this and other swapping operations?

src/uint/bingcd/xgcd.rs
erik-3milabs: Fixed in #761

tarcieri:
@erik-3milabs I just set `master` to v0.7.0-pre in #765

This means you can now make breaking changes, such as removing the existing safegcd implementation and changing trait impls like `Gcd` and `InvMod` to use `bingcd` instead

erik-3milabs:
> @erik-3milabs I just set `master` to v0.7.0-pre in #765
> 
> This means you can now make breaking changes, such as removing the existing safegcd implementation and changing trait impls like `Gcd` and `InvMod` to use `bingcd` instead

@tarcieri While I could still modify the `Gcd` trait to use this algorithm, this PR does not yet introduce the tools necessary to replace `InvMod`.

Aside from that, what else would be required to see this PR merged?

tarcieri:
Aah, lack of invmod support would definitely be a problem. Is it something you plan on addressing eventually? My understanding is, like safegcd, that invmod is a big part of its intended usage.

It seems a little weird to have multiple implementations of GCD algorithms which effectively do the same thing, though to completely replace safegcd in addition to invmod you'd also need support for boxed types (though with `const_mut_refs` stable it should be a lot easier to share an implementation).

Also seems it needs a rebase due to upstream changes.

erik-3milabs:
> Aah, lack of invmod support would definitely be a problem. Is it something you plan on addressing eventually? My understanding is, like safegcd, that invmod is a big part of binary GCD's intended usage.

You're right. This PR only introduces the `gcd` algorithm; PR #761 extends the algorithm into `xgcd`. Stripping some things from the `xgcd` algorithm gives `invmod`. Given that I don't need `invmod` myself, I am not too keen on implementing it :see_no_evil:

> It seems a little weird to have multiple implementations of GCD algorithms which effectively do the same thing, though to completely replace safegcd in addition to invmod you'd also need support for boxed types (though with `const_mut_refs` stable it should be a lot easier to share an implementation).

I agree that having two algorithms is overkill. Let me see about implementing this for `Boxed<X>` as well.

> Also seems it needs a rebase due to upstream changes.

Yeah, you're right. Let me address that right away.

kayabaNerve:
I'm incredibly interested in this PR as someone who:

1) Needs a constant-time GCD
2) Has >50% of my runtime spent on `safegcd::divsteps` alone right now

I actually need an `xgcd` (which #761 solves), it's just trivial to implement `xgcd` given `gcd` (by manually calculating `((a / g) % (b / g))**-1` for the first coefficient and solving from there for the second). I'd love to see this merged ASAP accordingly and want to ask if anything is a current blocker I can help with.

tarcieri:
Whoops, fat fingered the close button trying to reply.

While this is probably in OK shape as is I would preferably like to see it as the only GCD implementation, rather than having two.

I think it’s fine to retain `safegcd` for inversions until this can be extended to support inversions as well, but I’d really prefer for there to be one algorithm used for GCD for both boxed and unboxed types.

kayabaNerve:
Heard, especially if the goal of `0.7.0` is a long-lived release without continued maintenance of `safegcd`.

modinv should be trivial as just the `x` coefficient from `xgcd`, `Some` if `gcd == 1`, `None` otherwise, right?

Thanks for the quick reply and for letting me ensure I have context!

src/uint/cmp.rs
tarcieri: Not in `const fn`

tarcieri (approved):
Will go ahead and land this and we can follow up on how best to expose it.

Personally I would like for it to be the one-and-only GCD implementation (well, the two flavors of bingcd, that is)

tarcieri:
Hmm, merging this broke the build due to a few deprecations, but hopefully we can get those fixed up easily

Issue #752, opened 2025-01-24
List of breaking changes for v0.7 release


tarcieri:
Now that v0.6.0 is out, we can collect and discuss potential breaking changes for a (not currently scheduled) v0.7 release:

- [ ] 2024 edition upgrade (requires Rust 1.85, to be released 2025-02-20)
- [ ] `rand_core` v0.9 (not yet released)
- [ ] `der` v0.8 (not yet released)

nbdd0121:
`rand_core` v0.9 is now [released](https://github.com/rust-random/rand/issues/1165).

fjarri:
That actually concerns me too. Feels like we should do a quick bump to 0.7 with `rand_core=0.9` before v0.6 is widely adopted.

tarcieri:
Yeah, alas, not great timing. I can yank the v0.6.0 release and retroactively upgrade it.

tarcieri:
Another thing I could do which is a little annoying is start a new development branch and upgrade there, and cut a v0.7.0 prerelease

fjarri:
> I think it is good to have 0.6 out with rand 0.8, at least for rsa I want to make a release with rand 0.8 and the bug fix soon.

If that's the plan, I'm going to make a `crypto-primes` release as well.

tarcieri:
Alright, with #1206 I think we're good to go on a v0.7.0 release.

There's still a couple outstanding PRs we can wrap up: #1139 #1181 but other than that I think we should wrap things up and get a final v0.7.0 released.

cc @andrewwhitehead

andrewwhitehead:
It might be nice to adjust some of the naming conventions as in https://github.com/RustCrypto/crypto-bigint/issues/537, and related to that maybe deprecating `BoxedUint::mul` (because it's ambiguous about the widening behavior).

tarcieri:
v0.7.0 release PR is up: #1218

tarcieri:
I will close this out as I believe we've done all of the renaming needed to match core conventions

PR #758, opened 2025-02-02
rand 0.9

serprex:


tarcieri:
We need to figure out what to do about versioning since this is a breaking change.

See also #752

baloo:
I believe this is one of the blocker to deploy other changes in the consumers of this crate.

Like `RSA` depends on `signature` and `crypto-bigint`. `signature` `2.3.0-pre.5` bumped `rand_core` to `0.9`, but `crypto-bigint` did not yet.

So we're stuck on `signature 2.3.0-pre.4` for now.

newpavlov:
Closing in favor of #770

PR #760, opened 2025-02-13
Add `?Sized` to all `RngCore` bounds

fjarri:
Doesn't matter to us, but allows users to pass `dyn RngCore` args.


tarcieri (approved):
Guess we can migrate these to explicit generic parameters (possibly again) for consistent with `crypto-common` once we update to `rand_core` v0.9

PR #764, opened 2025-02-14
v0.6.1

tarcieri:
### Added
- `?Sized` to all RngCore bounds ([#760])

### Changed
-  Make `as_limbs_mut` const ([#757])

[#757]: https://github.com/RustCrypto/crypto-bigint/pull/757
[#760]: https://github.com/RustCrypto/crypto-bigint/pull/760


fjarri:
Technically #760 may be breaking is someone implemented `RandomBits` for their type, but there's probably a very low likelihood of that

PR #770, opened 2025-02-19
provide `RandomMod::try_random_mod` and `Random::try_random` methods

baloo:
This method is to be used by downstream crates:
  - https://github.com/RustCrypto/RSA/pull/478
  
Depends on:
 - https://github.com/rust-random/rand/pull/1593

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ subtle = { version = "2.6", default-features = false }
 der = { version = "0.8.0-rc.1", optional = true, default-features = false }
 hybrid-array = { version = "0.2", optional = true }
 num-traits = { version = "0.2.19", default-features = false }
-rand_core = { version = "0.9", optional = true, default-features = false }
+rand_core = { version = "0.9.2", optional = true, default-features = false }
 rlp = { version = "0.6", optional = true, default-features = false }
 serdect = { version = "0.3", optional = true, default-features = false }
 zeroize = { version = "1", optional = true, default-features = false }
@@ -81,7 +81,3 @@ harness = false
 [[bench]]
 name = "int"
 harness = false
-
-[patch.crates-io]
-# https://github.com/rust-random/rand/pull/1593
-rand_core = { git = "https://github.com/rust-random/rand.git" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,3 +81,7 @@ harness = false
 [[bench]]
 name = "int"
 harness = false
+
+[patch.crates-io]
+# https://github.com/rust-random/rand/pull/1593
+rand_core = { git = "https://github.com/rust-random/rand.git" }
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -419,7 +419,24 @@ pub trait RandomMod: Sized + Zero {
     /// example, it implements `CryptoRng`), then this is guaranteed not to
     /// leak anything about the output value aside from it being less than
     /// `modulus`.
-    fn random_mod<R: RngCore + ?Sized>(rng: &mut R, modulus: &NonZero<Self>) -> Self;
+    fn random_mod<R: RngCore + ?Sized>(rng: &mut R, modulus: &NonZero<Self>) -> Self {
+        let Ok(out) = Self::try_random_mod(rng, modulus);
+        out
+    }
+
+    /// Generate a random number which is less than a given `modulus`.
+    ///
+    /// This uses rejection sampling.
+    ///
+    /// As a result, it runs in variable time that depends in part on
+    /// `modulus`. If the generator `rng` is cryptographically secure (for
+    /// example, it implements `CryptoRng`), then this is guaranteed not to
+    /// leak anything about the output value aside from it being less than
+    /// `modulus`.
+    fn try_random_mod<R: TryRngCore + ?Sized>(
+        rng: &mut R,
+        modulus: &NonZero<Self>,
+    ) -> Result<Self, R::Error>;
 }
 
 /// Compute `self + rhs mod p`.
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -299,7 +299,15 @@ pub trait Random: Sized {
     /// Generate a random value.
     ///
     /// If `rng` is a CSRNG, the generation is cryptographically secure as well.
-    fn random<R: RngCore + ?Sized>(rng: &mut R) -> Self;
+    fn random<R: RngCore + ?Sized>(rng: &mut R) -> Self {
+        let Ok(out) = Self::try_random(rng);
+        out
+    }
+
+    /// Generate a random value.
+    ///
+    /// If `rng` is a CSRNG, the generation is cryptographically secure as well.
+    fn try_random<R: TryRngCore + ?Sized>(rng: &mut R) -> Result<Self, R::Error>;
 }
 
 /// Possible errors of the methods in [`RandomBits`] trait.
--- a/src/uint/boxed/rand.rs
+++ b/src/uint/boxed/rand.rs
@@ -36,8 +36,17 @@ impl RandomBits for BoxedUint {
 impl RandomMod for BoxedUint {
     fn random_mod<R: RngCore + ?Sized>(rng: &mut R, modulus: &NonZero<Self>) -> Self {
         let mut n = BoxedUint::zero_with_precision(modulus.bits_precision());
-        random_mod_core(rng, &mut n, modulus, modulus.bits());
+        let Ok(()) = random_mod_core(rng, &mut n, modulus, modulus.bits());
         n
+    }
+
+    fn try_random_mod<R: TryRngCore + ?Sized>(
+        rng: &mut R,
+        modulus: &NonZero<Self>,
+    ) -> Result<Self, R::Error> {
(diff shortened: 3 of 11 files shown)


baloo:
I think this is fine as `random_mod` is rejection sampling and it's not constant time in the first place, but please confirm that.

Cargo.toml
```diff
@@ -23,7 +23,7 @@ subtle = { version = "2.6", default-features = false }
 der = { version = "0.8.0-rc.1", optional = true, default-features = false }
 hybrid-array = { version = "0.2", optional = true }
 num-traits = { version = "0.2.19", default-features = false }
-rand_core = { version = "0.9", optional = true, default-features = false }
+rand_core = { version = "0.9.1", optional = true, default-features = false }
```
newpavlov, line 26: I don't think you use anything added in `rand_core v0.9.1`, so you can use just `version = "0.9"` here.

baloo: yeah, this will need to eventually bump to `0.9.2`

src/uint/boxed/rand.rs
```diff
@@ -36,9 +36,19 @@ impl RandomBits for BoxedUint {
 impl RandomMod for BoxedUint {
     fn random_mod<R: RngCore + ?Sized>(rng: &mut R, modulus: &NonZero<Self>) -> Self {
         let mut n = BoxedUint::zero_with_precision(modulus.bits_precision());
-        random_mod_core(rng, &mut n, modulus, modulus.bits());
+        random_mod_core(rng, &mut n, modulus, modulus.bits())
+            .expect("RngCore used as TryRngCore failed");
```
baloo, line 40: This should be a `map_err(|_: Infallible| panic!())` to protect this assumption against a refactor.

newpavlov: You can use `let Ok(()) = random_mod_core(rng, &mut n, modulus, modulus.bits());`. This feature was [stabilized](https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html#omitting-empty-types-in-pattern-matching) in Rust 1.82, so we can use it.

Cargo.toml
newpavlov: But why do you need the relaxed `Sized` bounds in this PR?

src/uint/boxed/rand.rs
baloo: TIL

Cargo.toml
baloo:
```
error[E0277]: the trait bound `R: TryRngCore` is not satisfied
   --> src/uint/boxed/rand.rs:39:38
    |
39  |         let Ok(()) = random_mod_core(rng, &mut n, modulus, modulus.bits());
    |                      --------------- ^^^ the trait `TryRngCore` is not implemented for `R`
    |                      |
    |                      required by a bound introduced by this call
    |
    = note: required for `R` to implement `TryRngCore`
note: required by a bound in `random_mod_core`
   --> src/uint/rand.rs:102:37
    |
102 | pub(super) fn random_mod_core<T, R: TryRngCore + ?Sized>(
    |                                     ^^^^^^^^^^ required by this bound in `random_mod_core`
help: consider mutably borrowing here
    |
39  |         let Ok(()) = random_mod_core(&mut rng, &mut n, modulus, modulus.bits());
    |                                      ++++
```


Because the calls to `random_mod_core` rely on the blanket implementation: `impl TryRngCore for R where R: RngCore` which only work if you have a sized object in `0.9.1` (fixed in https://github.com/rust-random/rand/pull/1593)

newpavlov: You need to remove the unnecessary `?Sized` bounds. I pushed the commit with appropriate changes.

src/traits.rs
```diff
@@ -299,7 +299,7 @@ pub trait Random: Sized {
     /// Generate a random value.
     ///
     /// If `rng` is a CSRNG, the generation is cryptographically secure as well.
-    fn random<R: RngCore + ?Sized>(rng: &mut R) -> Self;
+    fn random<R: RngCore>(rng: &mut R) -> Self;
```
newpavlov, line 302: Please also add a `try_random` method to this trait. I think you also could add a blanket impl for `random` in terms of `try_random`. Same for `random_mod`/`try_random_mod`.

Cargo.toml
tarcieri: @newpavlov why did you remove these bounds? We added them everywhere to make it possible to pass trait object RNGs

newpavlov: The explicit `?Sized` bounds will become unnecessary after rand_core v0.9.2 lands. With it the bound will be part of `*Core` trait definitions.

fjarri: Are you referring to https://github.com/rust-random/rand/pull/1593? I don't think it will do that, it only allows for infallible to fallible conversion between dyn trait arguments.

newpavlov: @fjarri 
Hm... You are right, I misremembered that the PR does. Maybe we should add `?Sized` bound to the RNG traits as well?

baloo:
That last commit breaks rsa:
```
error[E0277]: the size for values of type `R` cannot be known at compilation time
   --> $HOME/.cargo/git/checkouts/crypto-primes-51ab25d802eaa634/a3c1206/src/hazmat/miller_rabin.rs:121:36
    |
108 |     pub fn test_random_base<R: CryptoRng + ?Sized>(&self, rng: &mut R) -> Primality {
    |                             - this type parameter needs to be `Sized`
...
121 |         let random = T::random_mod(rng, &range_nonzero)
    |                      ------------- ^^^ doesn't have a size known at compile-time
    |                      |
    |                      required by a bound introduced by this call
    |
note: required by an implicit `Sized` bound in `random_mod`
   --> $HOME/.cargo/git/checkouts/crypto-bigint-659be7e2563ea57f/f98d5a7/src/traits.rs:422:19
    |
422 |     fn random_mod<R: RngCore>(rng: &mut R, modulus: &NonZero<Self>) -> Self;
    |                   ^ required by the implicit `Sized` requirement on this type parameter in `RandomMod::random_mod`
help: consider removing the `?Sized` bound to make the type parameter `Sized`
    |
108 -     pub fn test_random_base<R: CryptoRng + ?Sized>(&self, rng: &mut R) -> Primality {
108 +     pub fn test_random_base<R: CryptoRng>(&self, rng: &mut R) -> Primality {
    |

error[E0277]: the size for values of type `R` cannot be known at compilation time
   --> $HOME/.cargo/git/checkouts/crypto-primes-51ab25d802eaa634/a3c1206/src/hazmat/sieve.rs:43:41
    |
36  | pub fn random_odd_integer<T: Integer + RandomBits, R: TryCryptoRng + ?Sized>(
    |                                                    - this type parameter needs to be `Sized`
...
43  |     let mut random = T::try_random_bits(rng, bit_length)?;
    |                      ------------------ ^^^ doesn't have a size known at compile-time
    |                      |
    |                      required by a bound introduced by this call
    |
note: required by an implicit `Sized` bound in `try_random_bits`
   --> $HOME/.cargo/git/checkouts/crypto-bigint-659be7e2563ea57f/f98d5a7/src/traits.rs:377:24
    |
377 |     fn try_random_bits<R: TryRngCore>(
    |                        ^ required by the implicit `Sized` requirement on this type parameter in `RandomBits::try_random_bits`
help: consider removing the `?Sized` bound to make the type parameter `Sized`
    |
36  - pub fn random_odd_integer<T: Integer + RandomBits, R: TryCryptoRng + ?Sized>(
36  + pub fn random_odd_integer<T: Integer + RandomBits, R: TryCryptoRng>(
    |
```

Cargo.toml
fjarri: Traits are already `?Sized` by default, but method's generic parameters (whether explicit or implicit) are implicitly `Sized`. Rust do be like that.

baloo:
I think `crypto-primes` relies fairly heavily on the `?Sized`

fjarri:
So I started this whole `?Sized` thing because it was mentioned in that `RSA` PR that switched to `crypto-bigint`. And also it seemed like we're not losing anything by relaxing this bound. 

I have a use for it in another project where I need to pass an RNG through a dyn-safe trait API. Before it was available I just had a wrapper that boxed a `dyn CryptoRngCore` and could be used as a `Sized` argument, which was kind of inconvenient and added a minor performance overhead.

Cargo.toml
newpavlov: You are right. I reverted my two commits.

newpavlov:
As discussed above, the `?Sized` bounds are indeed necessary, so I reverted my two commits.

Snowiiii:
When this is ready ?

baloo:
I think it's mostly ready, but we might want to wait for a release of rand_core.

newpavlov:
@baloo 
BTW we generally use "squash and merge" for all PRs, so having a clear commit history inside PRs is not very important. Force push also breaks links on the "notifications" page, resulting in a "couldn’t find those commits" page, and makes it a bit harder to track changes during PR development.

baloo:
This last push was fixing a merge conflict with #772 
I've switched to using Jujutsu a month or so ago, and it's less natural to keep piling commits with it that it would be with a straight git. Sorry for the notification noise.

baloo:
Thanks!

PR #772, opened 2025-02-20
Bump edition to 2024

baloo:
see https://github.com/RustCrypto/crypto-bigint/issues/752

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ microcontrollers).
 
 ## Minimum Supported Rust Version
 
-This crate requires **Rust 1.83** at a minimum.
+This crate requires **Rust 1.85** at a minimum.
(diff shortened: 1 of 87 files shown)


tarcieri:
@baloo can you also update the references in README.md?

README.md
```diff
@@ -39,7 +39,7 @@ microcontrollers).
 
 ## Minimum Supported Rust Version
 
-This crate requires **Rust 1.83** at a minimum.
+This crate requires **Rust 1.85** at a minimum.
```
newpavlov, line 42: You also need to update the badge. And I think we can remove this section completely. With the new MSRV-aware resolver we can bump MSRV without issues even in patch releases.

PR #776, opened 2025-02-22
Inline low level multiplication and reduction functions

fjarri:
Fixes a performance regression introduced in #667. Evidently, compiler relies a lot on knowing the slice sizes at compile time, so I'm inlining `schoolbook_multiplication()`, `schoolbook_squaring()`, and `montgomery_reduction_inner()`, so the compiler can optimize in case of `Uint`s.

Benchmarks:
- `wrapping ops/split_mul, U256xU256` - 26ns to 9ns
- `Const Montgomery arithmetic/multiplication, U256*U256` - 41ns to 21ns
- `Dynamic Montgomery arithmetic/multiplication, U256*U256` - 62ns to 44ns

The effect is less pronounced for longer integers, but sill amounts to 5-10% speedup for U4096.

On a higher level, this affects many `crypto-primes` benchmarks, e.g. doubles the speed of Lucas test for U128.

Possible addition: I think `panic!` in these functions can be replaced with `debug_assert!`, but I don't insist on it.

### DIFF
--- a/src/modular/reduction.rs
+++ b/src/modular/reduction.rs
@@ -6,6 +6,7 @@ use crate::{Limb, Odd, Uint};
 use {crate::BoxedUint, subtle::Choice};
 
 /// Algorithm 14.32 in Handbook of Applied Cryptography <https://cacr.uwaterloo.ca/hac/about/chap14.pdf>
+#[inline(always)]
 const fn montgomery_reduction_inner(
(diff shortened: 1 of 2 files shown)


fjarri:
Actually wait a second, the same effect is achieved by just inlining.

tarcieri (approved):

PR #777, opened 2025-02-23
Monty-related performance improvements

### LINKED ISSUE #782: Almost Montgomery Multiplication
`BoxedUint` exponentiation uses something called Almost Montgomery Multiplication, with a reference to https://eprint.iacr.org/2011/239.pdf. The idea is to use MM that is faster, but does not necessarily produce a fully reduced result, so the reduction can only be applied at the end (similar to what `k256` is doing). The comments claim that the result lies in the range `[0, 3M)` where `M` is the modulus. This was extended from `[0, 2M)` to get rid of some proptest failures in #455, with a reference to https://golang.org/issue/13907, which doesn't explain it either.

Now, the paper does not say anything about it. In fact, the only thing it claims is that the result is under `2^R` where `R` is the total `BoxedUint` capacity (that is, `2^bit_precision` in our terms). The authors don't investigate further since all the moduli they're interested in take the full capacity of the integer, and therefore the modulus needs to be subtracted at most once for full reduction.

Moreover, the algorithm in the paper does not really match the implementation. What they do is pretty much normal Montgomery reduction (with a separate multiplication step, like we do for `Uint`s), but they prove that if you start from non-reduced arguments (within `2^R`), you get the result within `2^R`. That's it. The algorithm we use performs interleaved multiplication + reduction. 

@dignifiedquire, since you were the original implementor of this algorithm in `num-bigint`, could you perhaps shed some light on this? How did you come up with it, or where did you take it from? (Looks like there's [similar code in Go](https://github.com/golang/go/blob/74ba2164a04bcc6600782d0d200103ecaa621ef3/src/math/big/nat.go#L187))

Based solely on what's said above, and with no further information, I think the best course of action is to perform a full reduction (we can optimize the current one given that we reduce a LIMBS size integer, not a 2*LIMBS one) in the end of `pow_montgomery_form()`.

### But wait, there's more.

I made a test stand with `Uint` composed of `u8`-sized limbs, to speed up testing and increase the probability of hitting edge cases, and ran a number of tests for AMM with different arguments. Hereinafter I denote `f(x) = floor(x / M)` where `x` is some number in Montgomery form, possibly non-reduced, and `M` is the modulus. That is, it is the number of subtractions of the modulus required to fully reduce `x`. 

The search space was still pretty large for exhaustive tests (at least without a significant optimization work which I didn't have time to do), so I ran random tests with 2-limb integers. The results are as follows.

1. In general, if `f(x) = m` and `f(y) = n`, then `f(AMM(x, y)) <= min(m, n) + 1`. That is the "reduction error" grows with every operation, but is determined by the argument with the lower error. 
2. To retrieve the number from Montgomery form we MM it by 1. In this case `f(AMM(x, 1)) = 0`, that is the result is always fully reduced regardless of `f(x)`. 
3. `f(AMM(x, x)) <= 1` regardless of `f(x)`. That is, squaring resets the error to  at most 1.

That is where the limit of `f = 2` in `pow_montomery_form()` comes from. We have an incoming `x` with `f(x) = 0`. We build an array of `powers` which are produced by multiplying the previous power by `x`, so for each power `f <= 1`. Then we take turns squaring the accumulator (bringing its `f` to 1) and multiplying by a power of `x` (bringing its `f` to at most 2). Then we either return it, or square again, which brings `f` back to 1.

I would very much like to have a formal mathematical foundation for this, but that first requires understanding what our current implementation is actually based on. 

If we accept the results of these numerical experiments, then the current approach is correct, we can use AMM with one reduction for normal multiplication and conversion to Montgomery form, and we can even get rid of reduction completely in `retrieve()`. Which would be very nice since AMM is significantly faster than mul + MM reduce. 

Thoughts?

fjarri:
Trying to figure out massive slowdowns `crypto-primes` experiences for boxed uints (up to 4x). Could be the reason of the slowdowns in `RSA` as well. 

Public changes:
- Added `Monty::div_by_2_assign()` (with a blanket impl).
- Added `BoxedUint::inv_mod2k_vartime()`. 
- Made `BoxedUint::inv_mod2k()` public. 
- Added `Monty::Multiplier` associated type and `Monty::copy_montgomery_from()` to assist with tight loops (specifically, Lucas test in `crypto-primes`). 
- Cleaned up AMM, added comments and references, and reduced the size of the internal buffer to N from 2N. Also made it `const fn`. Closes #782

**Note:** the multiplier for `Uint` is called `DynMontyMultiplier`. Not happy with the name, but we already have `MontyMultiplier` as a trait, and it clashes.

**Note:** the exact way MontyMultiplier is exposed and the naming I'm not sure about, also not sure how hazmat do we want to make them. Potentially AMM can be exposed too, but it would be good to wrap the results in some struct that will propagate the "reduction level". Not for this PR, I need to finalize the minimum viable solution.

Fixes:
- Fixed a bug in `BoxedUnsatInt::to_uint()` which created a 64-bit number instead of a 32-bit one on 32-bit targets

Internal:
- Added tests for `BoxedUint::inv_mod2k()` and `inv_mod2k_vartime()`. 
- Removed allocations inside the loop in `BoxedUint::inv_mod2k()`.
- Used  and `inv_mod2k_vartime()` in `BoxedMontyParams::new_vartime()` and `new()` - since it's only vartime in the `k`, which is fixed.
- `new_vartime()` can be made even faster (~15% for Uint, 25% for Boxed) if we make a variant of `inv_mod2k` that is vartime in both arguments. Currently added in the commit as `inv_mod2k_full_vartime()` (crate-private). **Can be removed if that's too much detail.**
- Removed an unnecessary allocation in Add/SubAssign of `BoxedMontgomeryForm`.

Performance notes:
- `BoxedUint::div_by_2()` uses `div_by_2_assign()` because it is faster and does not allocate. 
- `Uint::div_by_2()` uses the same approach, gets rid of one addition and one `shr1()`, so it is marginally faster (~10%).  
- As expected, because of `inv_mod2k_vartime()` usage `MontyParams::new/_vartime()` became massively faster (~10x for Uint, ~15x for Boxed, 4096 bits). 
- I tried using AMM in `Uint`, but it leads to performance degradation for smaller uints (U256). So for now we'll keep the status quo with `Uint` using multiply + reduce. Worth investigating later.

### DIFF
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -256,6 +256,12 @@ impl BoxedMontyForm {
             params: self.params.clone(),
         }
     }
+
+    /// Performs division by 2 inplace, that is finds `x` such that `x + x = self`
+    /// and writes it into `self`.
+    pub fn div_by_2_assign(&mut self) {
+        div_by_2::div_by_2_boxed_assign(&mut self.montgomery_form, &self.params.modulus)
+    }
 }
 
 impl Retrieve for BoxedMontyForm {
@@ -299,6 +305,10 @@ impl Monty for BoxedMontyForm {
 
     fn div_by_2(&self) -> Self {
         BoxedMontyForm::div_by_2(self)
+    }
+
+    fn div_by_2_assign(&mut self) {
+        BoxedMontyForm::div_by_2_assign(self)
     }
 
     fn lincomb_vartime(products: &[(&Self, &Self)]) -> Self {
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -8,10 +8,9 @@ mod neg;
 mod pow;
 mod sub;
 
-use super::{
-    ConstMontyParams, Retrieve, div_by_2,
-    reduction::{montgomery_reduction_boxed, montgomery_reduction_boxed_mut},
-};
+use super::{ConstMontyParams, Retrieve, div_by_2, reduction::montgomery_reduction_boxed};
+use mul::MontyMultiplier;
+
 use crate::{BoxedUint, Limb, Monty, Odd, Word};
 use alloc::sync::Arc;
 use subtle::Choice;
@@ -67,7 +66,10 @@ impl BoxedMontyParams {
 
         let mod_leading_zeros = modulus.as_ref().leading_zeros().min(Word::BITS - 1);
 
-        let r3 = montgomery_reduction_boxed(&mut r2.square(), &modulus, mod_neg_inv);
+        let r3 = {
+            let mut mm = MontyMultiplier::new(&modulus, mod_neg_inv);
+            mm.square(&r2)
+        };
 
         Self {
             modulus,
@@ -107,7 +109,10 @@ impl BoxedMontyParams {
 
         let mod_leading_zeros = modulus.as_ref().leading_zeros().min(Word::BITS - 1);
 
-        let r3 = montgomery_reduction_boxed(&mut r2.square(), &modulus, mod_neg_inv);
+        let r3 = {
+            let mut mm = MontyMultiplier::new(&modulus, mod_neg_inv);
+            mm.square(&r2)
+        };
 
         Self {
             modulus,
@@ -336,11 +341,8 @@ impl Zeroize for BoxedMontyForm {
 /// Convert the given integer into the Montgomery domain.
 #[inline]
 fn convert_to_montgomery(integer: &mut BoxedUint, params: &BoxedMontyParams) {
-    let mut product = integer.mul(&params.r2);
-    montgomery_reduction_boxed_mut(&mut product, &params.modulus, params.mod_neg_inv, integer);
-
-    #[cfg(feature = "zeroize")]
-    product.zeroize();
+    let mut mm = MontyMultiplier::from(params);
+    mm.mul_assign(integer, &params.r2);
 }
 
 #[cfg(test)]
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -9,7 +9,7 @@ mod pow;
 mod sub;
 
 use super::{ConstMontyParams, Retrieve, div_by_2};
-use mul::MontyMultiplier;
+use mul::BoxedMontyMultiplier;
 
 use crate::{BoxedUint, Limb, Monty, Odd, Word};
 use alloc::sync::Arc;
@@ -67,7 +67,7 @@ impl BoxedMontyParams {
         let mod_leading_zeros = modulus.as_ref().leading_zeros().min(Word::BITS - 1);
 
         let r3 = {
-            let mut mm = MontyMultiplier::new(&modulus, mod_neg_inv);
+            let mut mm = BoxedMontyMultiplier::new(&modulus, mod_neg_inv);
             mm.square(&r2)
         };
 
@@ -110,7 +110,7 @@ impl BoxedMontyParams {
         let mod_leading_zeros = modulus.as_ref().leading_zeros().min(Word::BITS - 1);
 
         let r3 = {
-            let mut mm = MontyMultiplier::new(&modulus, mod_neg_inv);
+            let mut mm = BoxedMontyMultiplier::new(&modulus, mod_neg_inv);
             mm.square(&r2)
         };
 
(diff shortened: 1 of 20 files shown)


fjarri:
@tarcieri I'm still working on this PR, but meanwhile, it seems that I'll need to expose the following in the public API:
- Multiplication/squaring of `Monty` with the provided buffer (since it seems to be impossible to do Montgomery multiplication fully in place). Currently it's `mul_into()` and `square_into()`.
- Copying one uint into another. Currently it's `fill_with()`.

The buffer can be the same size as the arguments if we use interleaved Montgomery multiplication (I added both CIOS and FIOS ones, seem to have the same performance, but CIOS is simpler). For some reason the interleaved form is a couple percent slower than multiplication + reduction, even though it's supposed to have less operations.

Another tight loop where a lot of allocations are happening is `pow_bounded_exp()`. I think I can reduce them there as well.

tarcieri:
I’m fine with exposing those methods in the public API.

Also getting rid of allocations would hopefully benefit `rsa` which is quite slow now, so that sounds great.

src/modular/boxed_monty_form.rs
```diff
@@ -82,23 +101,18 @@ impl BoxedMontyParams {
...
 
         let mod_neg_inv = Limb(Word::MIN.wrapping_sub(inv_mod_limb.limbs[0].0));
 
         let mod_leading_zeros = modulus.as_ref().leading_zeros().min(Word::BITS - 1);
 
-        let r3 = montgomery_reduction_boxed(&mut r2.square(), &modulus, mod_neg_inv);
+        let r3 = {
+            let mut mm = BoxedMontyMultiplier::new(&modulus, mod_neg_inv);
+            mm.square(&r2)
+        };
```
tarcieri, line 115: Any reason not to just write this as...

```suggestion
        let r3 = BoxedMontyMultiplier::new(&modulus, mod_neg_inv).square(&r2);
```

src/modular/monty_form.rs
```diff
@@ -90,12 +94,18 @@ impl<const LIMBS: usize> MontyParams<LIMBS> {
...
 
         let mod_neg_inv = Limb(Word::MIN.wrapping_sub(inv_mod.limbs[0].0));
 
-        let mod_leading_zeros = modulus.as_ref().leading_zeros_vartime().min(Word::BITS - 1);
+        let mod_leading_zeros = modulus.as_ref().leading_zeros_vartime();
+        let mod_leading_zeros = if mod_leading_zeros < Word::BITS - 1 {
+            mod_leading_zeros
+        } else {
+            Word::BITS - 1
+        };
```
tarcieri, line 108: What was wrong with the old version?

fjarri: `min()` is not `const fn`. Technically, I can revert making these constructors `const` since it is not required by this PR - I had an idea to merge them with `const_monty_form`, but it can wait.

src/modular/boxed_monty_form.rs
fjarri: Actually, now that I think about it, we can just use normal multiplication here, since the multiplier gets discarded immediately.

src/modular/monty_form.rs
tarcieri: Aah, it's `Ord::min`, I see.

I think it's fine to constify this.

src/uint/boxed/inv_mod.rs
```diff
@@ -13,14 +13,78 @@ impl BoxedUint {
     }
 
     /// Computes 1/`self` mod `2^k`.
+    /// This method is variable w.r.t. `self` and `k`.
     ///
     /// If the inverse does not exist (`k > 0` and `self` is even),
     /// returns `Choice::FALSE` as the second element of the tuple,
     /// otherwise returns `Choice::TRUE`.
-    pub(crate) fn inv_mod2k(&self, k: u32) -> (Self, Choice) {
+    pub(crate) fn inv_mod2k_full_vartime(&self, k: u32) -> (Self, Choice) {
```
tarcieri, line 21: Yeah, there are various permutations of what "vartime" means and I've not been sure how to handle when we want more than one. This seems fine, especially for `pub(crate)`

tarcieri (approved):

tarcieri:
Curious what the performance impact on `rsa` will be /cc @dignifiedquire

src/modular/boxed_monty_form.rs
fjarri: Actually number 2, we cannot use normal multiplication since the params are not created yet :) So I think the multiplier should stay, because the other option is to call `almost_montgomery_mul()` manually.

fjarri:
I am actually curious too, was planning to test it. `crypto-primes` with this PR reaches parity between `Uint` and `BoxedUint` for 1024 bits, but it needs some changes to make use of the newly exposed multiplier.

dignifiedquire:
do I need to update the usage, or does it make sense to test it directly?

fjarri:
> do I need to update the usage, or does it make sense to test it directly?

You might. Some calls are non-allocating automatically, but you may have to use the explicit multiplier object (that encapsulates the scratch buffer) if you do any multiplications in tight loops. Also `crypto-primes` needs some unpublished changes (to be made into a PR later today).

I'm testing it right now, and it seems that the changes in `crypto-primes` and `crypto-bigint` are not enough.

PR #780, opened 2025-02-24
Fix Clippy lints in tests

newpavlov:


src/int/sub.rs
```diff
@@ -75,138 +75,135 @@ impl<const LIMBS: usize> WrappingSub for Int<LIMBS> {
 }
 
 #[cfg(test)]
+#[allow(clippy::init_numbered_fields)]
```
newpavlov, line 78: I am not sure what exactly happens here, but replacing `Int { 0: I128::MIN.0.wrapping_add(&I128::ONE.0) }` with `Int(I128::MIN.0.wrapping_add(&I128::ONE.0))` does not work.

src/modular/boxed_monty_form/add.rs
```diff
@@ -71,6 +71,7 @@ impl AddAssign<BoxedMontyForm> for BoxedMontyForm {
 }
 
 #[cfg(test)]
+#[allow(clippy::unwrap_used)]
```
fjarri, line 74: May be better just do
```
#![cfg_attr(not(test), warn(clippy::unwrap_used))]
```
in `lib.rs`

src/int/sub.rs
tarcieri: cc @erik-3milabs

fjarri: I'm surprised `Int { 0:` even works given that the field is private. 

But this whole construct seems error-prone. Why not just `I128::MIN.wrapping_add(&I128::ONE)`?

fjarri: Or, if somehow creating an `Int` from `Uint` is desired, maybe it's worth creating a `from` method or a `From` impl (for the other direction there's already `as_uint()`)

erik-3milabs: I wrote these tests this way on purpose. Instead of using `add` and `sub` to test one another (thus, only having relative tests), I anchored them to `Uint` instead. I'm using `Uint` as a "source of truth" here.

I agree that `I128::MIN.wrapping_add(&I128::ONE)` would be much more readable, but it would remove that source of truth. Perhaps the neatest solution would be to use the solution you're proposing here, while also writing a new test anchoring the behavior to `Uint`.

E.g.

```rust
#[test]
fn sub_relative_to_uint() {
    assert_eq!(I128::MIN.wrapping_sub(&I128::ONE).as_uint(), Uint::ZERO); 
    assert_eq!(I128::MINUS_ONE.wrapping_sub(&I128::ONE).as_uint(), Uint::MAX.wrapping_sub(&Uint::ONE));
    assert_eq!(I128::ZERO.wrapping_sub(&I128::ONE).as_uint(), Uint::MAX); 
    assert_eq!(I128::ONE.wrapping_sub(&I128::ONE).as_uint(), Uint::ZERO); 
    assert_eq!(I128::MAX.wrapping_sub(&I128::ONE).as_uint(), Uint::MAX.shr(1).wrapping_sub(&Uint::ONE));
    
    assert_eq!(I128::MIN.wrapping_sub(&I128::MINUS_ONE).as_uint(), Uint::ONE.shl(Uint::BITS-1).wrapping_add(Uint::ONE)); 
    assert_eq!(I128::MINUS_ONE.wrapping_sub(&I128::MINUS_ONE).as_uint(), Uint::ZERO);
    assert_eq!(I128::ZERO.wrapping_sub(&I128::MINUS_ONE).as_uint(), Uint::ONE); 
    assert_eq!(I128::ONE.wrapping_sub(&I128::MINUS_ONE).as_uint(), Uint::ONE.shl(1)); 
    assert_eq!(I128::MAX.wrapping_sub(&I128::MINUS_ONE).as_uint(), Uint::ONE.shl(Uint::BITS-1));
}
```

fjarri: Or you can look at it this way: you're testing the contract `neg(I28::MAX) == I128::MIN + 1` without regard to how it might be implemented internally.

newpavlov: I will leave `allow(clippy::init_numbered_fields)` for now then. It can be fixed in a later PR by someone more familiar with the code.

PR #781, opened 2025-02-25
Make `random_bits_core` platform independent

erik-3milabs:
In its current state, `random_bits_core` behaves differently on 32-bit and 64-bit targets in two ways:
1. The platforms order the randomly sampled bytes in a different order, generating different outputs.
2. For bit_length $\mod 64 \in [1, 32]$, the 64-bit platform samples four random bytes more than the 32-bit platform.
    a.  this means the state of the RNG after calling this function is platform-dependent. This makes it impossible to use this function as an rng-seed-to-integer map.

This patch introduces:
- benchmarks to gauge the performance of the `random_bits_core` function,
- a patch to the `random_bits_core` function, and
- a test ensuring the `random_bits_core` output and rng state after calling it are platform-independent.

On my local machine, the function's performance remains constant with this patch applied.

### DIFF
--- a/src/uint/rand.rs
+++ b/src/uint/rand.rs
@@ -176,9 +176,9 @@ mod tests {
         "0F6760A48F9B887CB2FB0D3281E2A6E67746A55FBAD8C037B585F767A79A3B6C"
     ]);
 
-    /// Construct 4-sequential `rng`, i.e.,
+    /// Construct a 4-sequential `rng`, i.e., an `rng` such that
     /// `rng.fill_bytes(&mut buffer[..x]); rng.fill_bytes(&mut buffer[x..])` will construct the
-    /// same `buffer`, regardless the choice of `x` in `0..buffer.len()`.
+    /// same `buffer`, for `x` any in `0..buffer.len()` that is `0 mod 4`.
     fn get_four_sequential_rng() -> ChaCha8Rng {
         ChaCha8Rng::seed_from_u64(0)
     }
--- a/src/uint/rand.rs
+++ b/src/uint/rand.rs
@@ -20,6 +20,12 @@ impl<const LIMBS: usize> Random for Uint<LIMBS> {
 /// Fill the given limbs slice with random bits.
 ///
 /// NOTE: Assumes that the limbs in the given slice are zeroed!
+///
+/// When combined with a platform-independent "sequential" `rng`, this function acts
+/// platform-independent and deterministic. We consider an RNG "sequential" when
+/// `rng.fill_bytes(&mut buffer[..x]); rng.fill_bytes(&mut buffer[x..])` will construct the same
+/// `buffer`, regardless the choice of `x` in `0..buffer.len()`.
+/// Note that the `TryRngCore` trait does _not_ require this behaviour from `rng`.
 pub(crate) fn random_bits_core<R: TryRngCore + ?Sized>(
     rng: &mut R,
     zeroed_limbs: &mut [Limb],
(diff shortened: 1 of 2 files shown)


erik-3milabs:
I don't know what is up with the formatter, but it is doing some funky stuff! Sorting `Uint, U256, U1024` as `U256, U1024, Uint`. Since when does `rustfmt` sort on length first?

tarcieri:
The 2024 edition has new formatting rules I’m not particularly familiar with yet

erik-3milabs:
> The 2024 edition has new formatting rules I’m not particularly familiar with yet

ahhhh, I see.

src/uint/rand.rs
```diff
@@ -39,12 +39,26 @@ pub(crate) fn random_bits_core<R: TryRngCore + ?Sized>(
     for i in 0..nonzero_limbs - 1 {
         rng.try_fill_bytes(&mut buffer)
             .map_err(RandomBitsError::RandCore)?;
-        zeroed_limbs[i] = Limb(Word::from_be_bytes(buffer));
+        zeroed_limbs[i] = Limb(Word::from_le_bytes(buffer));
     }
 
-    rng.try_fill_bytes(&mut buffer)
+    // This algorithm should sample the same number of random bytes, regardless of the pointer width
```
fjarri, line 45: What guarantees does `rand_core` actually provide? Can we be sure that, when we ask for 4 random bytes, it doesn't still sample 8 and discard 4 of them? Like it's mentioned [here](https://docs.rs/rand_core/latest/rand_core/trait.RngCore.html#tymethod.next_u32).

Would it be safer to instead round up and sample 64-bit chunks both on 32- and 64-bit platforms?

erik-3milabs: > What guarantees does rand_core actually provide? Can we be sure that, when we ask for 4 random bytes, it doesn't still sample 8 and discard 4 of them?

If an RNG does something strange like sampling $x$ bytes when we ask for $y$, it sounds to me like either
- that RNG does not satisfy the contract set by its API, or
- a user calling this function with a "non-sequential"[^1] RNG incorrectly expects the `random_bits_core` function to produce a platform independent output.

Given that neither problem is with `random_bits_core`, I'd say that this crate is not responsible for dealing with these issues. Our sole responsibility is to make sure that if somebody passes in a sequential RNG, the output is deterministic.

> Would it be safer to instead round up and sample 64-bit chunks both on 32- and 64-bit platforms?

I don't know whether "safer" is relevant here. Even if it is, I'd be inclined to vote against such a solution. My concern with this approach is that, in my understanding, generating random bytes can be a relatively expensive procedure. The solution you're proposing 
1. puts the extra burden on 32-bit platforms (instead of 64-bit platforms, as I do here) as it needs to test whether to sample extra bytes, and then sample them in half of these cases. I usually assume 32-bit platforms are resource-restricted, so it feels misplaced to put the burden on them.
2. has 32-bit platforms sometimes sample four bytes it does not need, rather than stopping 64-bit platforms from sampling random bytes that would be tossed anyway (which is what the current implementation proposes).

[^1]: Here, we use the term `sequential` to refer to RNGs for which `rng.sample_bytes(x) || rng.sample_bytes(y-x)` is constant for any pair of integers `1 < x < y` , with `||` denoting concatenation.

fjarri: > that RNG does not satisfy the contract set by its API, or

The `RngCore::next_u32` docs I linked literally recommend doing it. 

> Our sole responsibility is to make sure that if somebody passes in a sequential RNG, the output is deterministic.

I'm ok with this contract. I wish `rand_core` had a `SequentialRng` as a trait, unlike `CryptoRng`, it's specific and easily testable.

fjarri: Looks good to me.

erik-3milabs: @tarcieri do you agree this is ready to be merged?

tarcieri: Let me do another pass

tarcieri (approved):

PR #785, opened 2025-03-02
Fixup assertion in AddAssign and SubAssign of BoxedUint

### LINKED ISSUE #784: Inconsistent behavior of AddAssign
(crypto-bigint master (0feb3c58))

```rust
use crypto_bigint::BoxedUint;

fn main() {
    let mut h = BoxedUint::one().widen(1024);
    assert_eq!(h.bits_precision(), 1024);

    h = h + BoxedUint::one();
    println!("h {h:?}");

    let mut h = BoxedUint::one().widen(1024);
    assert_eq!(h.bits_precision(), 1024);

    // self.bits_precision() <= (rhs.as_ref().len() as u32 * Limb::BITS)
    // thrown here
    h += BoxedUint::one();
}
```

In this example, the first `h = h + BoxedUint::one();` works fine, but the use of `AddAssign` triggers an assertion failure because it expects the `rhs` to have a larger bit precision than `self`.

I'm not actually sure I understand what's going on. Could it be the assertion is backwards?

baloo:
Fixes #784

### DIFF
--- a/src/uint/boxed/add.rs
+++ b/src/uint/boxed/add.rs
@@ -16,7 +16,7 @@ impl BoxedUint {
     /// Panics if `rhs` has a larger precision than `self`.
     #[inline]
     pub fn adc_assign(&mut self, rhs: impl AsRef<[Limb]>, mut carry: Limb) -> Limb {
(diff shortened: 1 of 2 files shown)


baloo:
This should be backported to 0.6

tarcieri (approved):

PR #587, opened 2024-04-02
Add `serde` for `BoxedUInt`

pinkforest:
Missing impl `Serialize` and `Deserialize`

PR #789, opened 2025-03-09
Implement serde for BoxedUint

wawwior:
Adapted from #587

### DIFF
--- a/src/uint/boxed/encoding/serde.rs
+++ b/src/uint/boxed/encoding/serde.rs
@@ -1,0 +1,72 @@
+use serdect::serde::{Deserialize, Deserializer, Serialize, Serializer, de::Error};
+
+use crate::BoxedUint;
+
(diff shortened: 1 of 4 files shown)


src/uint/boxed/encoding/serde.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use crate::BoxedUint;
+
+impl<'de> Deserialize<'de> for BoxedUint {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let slice = serdect::slice::deserialize_hex_or_bin_vec(deserializer)?;
+        Self::from_le_slice(&slice, slice.len() as u32 * 8).map_err(Error::custom)
```
wawwior, line 11: I should probably add an integer overflow check here

tarcieri (approved):

PR #791, opened 2025-03-12
RemMixed implementation for even splits

dvdplm:
When the `RemMixed` trait was added in #746 I missed that we need an impl for even splits as well. This rectifies that mistake.


tarcieri:
@dvdplm sorry this one went by the wayside. Can you rebase just so I know this still works?

dvdplm:
@tarcieri Done!

tarcieri (approved):

PR #792, opened 2025-03-12
Make Integer be Binary, LowerHex and UpperHex

dvdplm:
As per the title, but also add support for the "alternate" formatting flag for Binary, LowerHex and UpperHex. And tests.

Question: is it at all useful to test that `Integer` has the bounds? Probably not eh?

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -349,6 +349,36 @@ mod tests {
         assert_eq!(hex, n.to_string());
     }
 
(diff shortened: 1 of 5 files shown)


tarcieri (approved):

tarcieri:
> Question: is it at all useful to test that Integer has the bounds? Probably not eh?

I'd side with "probably not"

PR #795, opened 2025-03-14
impl `ConditionallyNegatable` for `BoxedUint`

dvdplm:
Add an equivalent to `subtle`'s `ConditionallyNegate` trait without the `Copy` requirement.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -77,6 +77,27 @@ impl<T: ConditionallySelectable> ConstantTimeSelect for T {
     }
 }
 
+/// A type which can be conditionally negated in constant time.
+///
+/// Similar to `subtle`'s [`ConditionallyNegate`] trait, but without the `Copy` bound which allows
+/// it to to be impl'd for heap allocated types.
+pub trait ConstantTimeNegatable {
+    /// Negate `self` according to `choice`.
+    fn ct_negate(&mut self, choice: Choice);
+}
+
+impl<T> ConstantTimeNegatable for T
+where
+    T: ConstantTimeSelect,
+    for<'a> &'a T: Neg<Output = T>,
+{
+    #[inline]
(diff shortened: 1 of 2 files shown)


tarcieri:
I'm not seeing a `Copy` bound on `subtle::ConditionallyNegatable`?

https://docs.rs/subtle/latest/subtle/trait.ConditionallyNegatable.html

dvdplm:
The Copy bound comes from T having to be ConditionallySelectable. On 14 Mar 2025, at 16:14, Tony Arcieri ***@***.***> wrote:﻿
  I'm not seeing a Copy bound on subtle::ConditionallyNegatable?
https://docs.rs/subtle/latest/subtle/trait.ConditionallyNegatable.html—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>



tarcieri left a comment (RustCrypto/crypto-bigint#795)
I'm not seeing a Copy bound on subtle::ConditionallyNegatable?
https://docs.rs/subtle/latest/subtle/trait.ConditionallyNegatable.html

tarcieri:
@dvdplm that's to leverage the blanket impl though. You should still be able to impl `ConditionallyNegatable` directly without issue

dvdplm:
Ah, I see what you mean now! Better now?

src/traits.rs
```diff
@@ -926,7 +926,7 @@ pub trait Monty:
 /// Allows one to perform inplace multiplication without allocations
 /// (important for the `BoxedUint` case).
 ///
-/// NOTE: You will be operating with Montgomery represntations directly,
+/// NOTE: You will be operating with Montgomery representations directly,
```
tarcieri, line 929: Bleh, we should set up `typos-action`

tarcieri (approved):

PR #799, opened 2025-03-31
`div` with differently sized denominators

erik-3milabs:
With this PR in place, denominators are no longer required to be a `Uint`/`Int` of the same number of limbs as the numerator, enabling more efficient division.


src/uint/div.rs
```diff
@@ -124,11 +127,17 @@ impl<const LIMBS: usize> Uint<LIMBS> {
         // Copy out the remainder
         y[0] = Limb::select(x[0], Limb(rem2), limb_div);
         i = 1;
-        while i < LIMBS {
+
+        let min = if LIMBS < RHS_LIMBS { LIMBS } else { RHS_LIMBS };
```
dvdplm, line 131: From a constant-time perspective, is this ok? I think it is, but perhaps worth spelling out explicitly in a comment? Like "This does not invalidate the constant time contract, because the compiler will eliminate this branch"?

dvdplm (commented):
It would be good to have this working when used through traits as well (i.e when I have an `Integer` rather than a concrete `U1024`). Is that something you can fit into this PR you think?

benches/uint.rs
```diff
@@ -219,7 +219,19 @@ fn bench_division(c: &mut Criterion) {
...
 
-    group.bench_function("div/rem, U256/U128, full size", |b| {
+    group.bench_function("div/rem, U256/U128", |b| {
+        b.iter_batched(
+            || {
+                let x = U256::random(&mut rng);
+                let y = U128::random(&mut rng);
+                (x, NonZero::new(y).unwrap())
+            },
+            |(x, y)| black_box(x.div_rem(&y)),
```
fjarri, line 229: Just a side node, you don't need to wrap the output in `black_box`, Criterion already does that for all `iter*` methods.

fjarri:
This fixes a part of #811

src/uint/div.rs
tarcieri: A comment might be good but IMO it's not too worrisome since it's just branching on constants

tarcieri:
@erik-3milabs mind rebasing this?

src/uint/div.rs
erik-3milabs: Done.

erik-3milabs:
> It would be good to have this working when used through traits as well (i.e when I have an `Integer` rather than a concrete `U1024`). Is that something you can fit into this PR you think?

Could you elaborate on this? I.e., support, say, `Int.div(Uint)` through the traits?

dvdplm:
> Could you elaborate on this? I.e., support, say, `Int.div(Uint)` through the traits?

What I had in mind here was a situation where you're working with generic numbers, i.e. `where T: Integer`. Example:

```
pub struct Thing<S, T> {
  a: S,
  b: T,
}

impl Thing<S, T> where T: Integer, S: Integer {
  pub fn my_thing(&self) {
    // do maths to change the world here
  }
}
```

My `S` and `T` are going to be `Uint`s (or `Int`s, once `Int` impls `Integer`?) but only features accessible through traits are available. For example `Uint::mul_mod_special` is not available through `Integer`, but `mul_mod` is.

Makes sense?

tarcieri (approved):

tarcieri:
> Is that something you can fit into this PR you think?

Let's save it for a followup

Issue #756, opened 2025-01-30
Abstraction over `[Limb]`


tarcieri:
`BoxedUint` contains large amount of code copy-pasted from `Uint`, some of which has been rewritten to perform in-place operations to reduce heap allocations.

Before Rust 1.83 it just wasn't possible to share code between `Uint` and `BoxedUint` because `Uint` has been carefully written to maximally support `const fn` which precludes `&mut [Limb]`, whereas with `BoxedUint` we always want to prefer in-place operations to reduce allocations and these require `&mut [Limb]`. The only way to "abstract" between the two was to use macros, which wasn't great.

As of Rust 1.83 we can now have `&mut` in `const fn` which makes it possible to share a common implementation.

One option might be to define a type like this:

```rust
pub struct UintRef([Limb]);
```

(it doesn't necessarily need to be `pub` but might potentially be useful as part of the public API)

Then various operations can be implemented on `&mut self` references to this type. However, we'd have to be careful that such functions are inlined to eliminate bounds checks on the size of the slice in the case of `Uint` where it's known statically.

Another option is simply to define the implementation in terms of free functions that operate on `&mut [Limb]` which is what we have now in a few places where macro-based abstractions were used previously. However, there are various algorithms that would benefit being implemented in terms of an abstraction between `Uint`/`BoxedUint`, and in such cases it might be nice to have a real type for that rather than having to import a bunch of free functions.

PR #803, opened 2025-04-21
implement `PartialOrd<Uint>` for `BoxedUint`

baloo:
TODO: 
 - [ ] Provide an implementation for `BoxedInt`

### DIFF
--- a/src/uint/boxed/cmp.rs
+++ b/src/uint/boxed/cmp.rs
@@ -5,7 +5,7 @@
 pub(super) use core::cmp::{Ordering, max};
 
 use super::BoxedUint;
-use crate::{ConstChoice, Limb};
+use crate::{ConstChoice, Limb, Uint};
 use subtle::{
     Choice, ConditionallySelectable, ConstantTimeEq, ConstantTimeGreater, ConstantTimeLess,
 };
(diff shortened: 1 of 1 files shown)


baloo:
This came up in https://github.com/RustCrypto/signatures/pull/906 but I'm not sure that's the right approach.

The alternative would be to duplicate the implementation of the `subtle::ConstantTimeGreater` and `ConstantTimeLess`, but I'm not sure I like this approach better than creating a short-lived `BoxedUint`.

Opinion?

tarcieri:
@baloo I think the real goal of this is to make it possible to compare a `Uint` constant to a `BoxedUint` while avoiding boxing the former, which this doesn't accomplish.

It should be possible to have a generic core implementation of this functionality that operates over `[Limb]` (or a higher level type, i.e. #756) which can be reused for comparing `Uint`, `BoxedUint`, or a mixture thereof.

tarcieri:
That said, if this came with TODOs and an open issue to optimize it, adding it now does open the door to future optimizations and enable code to be written with `Uint` constants, so I think it's OK for a start.

tarcieri:
Err, sorry, my last comment meant this is fine as a start and we can optimize later

baloo:
Oh that's wasn't about your comment, but I'd rather start over.
After looking at the code for a little while, I have actually no idea how to achieve an abstraction over `[Limb]`. That would not carry whether the outer type was actually a Uint or an Int, so I'm a bit lost, this is why I closed that PR.

tarcieri (approved):

PR #806, opened 2025-04-23
Add `UintRef`

### LINKED ISSUE #756: Abstraction over `[Limb]`

tarcieri:
Adds a reference newtype for a `Limb` slice which can be used as an abstraction for implementing shared functionality for both `Uint` and `BoxedUint`.

The initial implementation supports a limited amount of functionality but supports usage as both `&UintRef` and `&mut UintRef`.

This includes a common implementation of `fmt` impls shared between `Uint` and `BoxedUint`, replacing a copy-paste implementation.

Closes #756

### DIFF
--- a/src/uint.rs
+++ b/src/uint.rs
@@ -13,6 +13,8 @@ use zeroize::DefaultIsZeroes;
 #[cfg(feature = "extra-sizes")]
 pub use extra_sizes::*;
 
+pub use ref_type::UintRef;
+
 use crate::{
     Bounded, ConstCtOption, ConstZero, Constants, Encoding, FixedInteger, Int, Integer, Limb,
     NonZero, Odd, PrecomputeInverter, PrecomputeInverterWithAdjuster, Word,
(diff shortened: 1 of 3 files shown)


tarcieri:
Note: we'll have to be careful trying to use this for common implementations of mathematical functionality shared between `Uint` and `BoxedUint`, since using a slice type rather than a const generic array type might introduce bounds checks that don't exist when the compiler knows the number of limbs at the type level.

This will require judicious benchmarking and perhaps `#[inline]` or `#[inline(always)]` annotations.

baloo (approved):

PR #808, opened 2025-04-24
Keep `BoxedMontyParams` in an `Arc`

### LINKED ISSUE #804: Move `Arc` inside `BoxedMontyParams`?
Something I noticed while working with Montgomery form of boxed integers:

We currently store the parameters in an `Arc` in `BoxedMontgomeryForm` to reduce the amount of allocations while cloning those. But we still have to clone the parameters themselves while creating `BoxedMontgomeryForm`, since its `new()` takes `BoxedMontyParams` by value. Wouldn't it be better to have `BoxedMontyParams` a wrapper around an `Arc` of actual parameters, so that its cloning is cheap as well?

fjarri:
Fixes #804

Deref impl leaks the inner type. Should I make the conversion explicit instead?

### DIFF
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -13,7 +13,6 @@ use mul::BoxedMontyMultiplier;
 
 use crate::{BoxedUint, Limb, Monty, Odd, Word};
 use alloc::sync::Arc;
-use core::ops::Deref;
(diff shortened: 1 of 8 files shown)


tarcieri:
Yeah, maybe instead of `Deref` just add explicit delegator methods to the inner type?

PR #809, opened 2025-04-25
More convenience methods for boxed integer resizing

fjarri:
This PR attempts to simplify some scenarios that were encountered during the work on https://github.com/RustCrypto/RSA/pull/506.

- Instead of having `BoxedUint::shorten()`/`widen()` methods, introduce the `Resize` trait with more convenience methods. Implemented for `BoxedUint`, `NonZero<BoxedUint>`, `Odd<BoxedUint>`, and their references.
- `BoxedUint::shorten()` and `widen()` are marked as deprecated.

Note that `try_resize` fails if `self.bits() > at_least_bits_precision`, unlike the more relaxed `shorten()` behavior.

### DIFF
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -11,7 +11,7 @@ mod sub;
 use super::{ConstMontyParams, Retrieve, div_by_2};
 use mul::BoxedMontyMultiplier;
 
-use crate::{BoxedUint, Limb, Monty, Odd, Word};
+use crate::{BoxedUint, Limb, Monty, Odd, Resize, Word};
 use alloc::sync::Arc;
 use subtle::Choice;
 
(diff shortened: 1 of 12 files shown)


fjarri:
@tarcieri any comments on that?

The other inconvenience in https://github.com/RustCrypto/RSA/pull/506 is that the constant-time division requires both arguments to be the same size, but the variable time division has no such restriction. I'm looking into that, will be in another PR.

tarcieri:
I think "resize" is fine over "resized".

> The trait is currently made to be implemented both for BoxedUint and &BoxedUint, to reduce the amount of allocations. Is it worth the increased code complexity?

Why not just have the trait borrow `&self` in its methods?

> Do we need the shorten() and widen() shortcuts, or resize() is enough?

Maybe mark them as deprecated and we can try to migrate to `resize()`

> try_resize() returns an Option since it is only constant-time wrt self. Is that acceptable?

Can start with that, sure

> Do we make try_resize() fail if at_least_bits_precision is less than self.bits(), but the number of resulting limbs is unchanged (that is, no truncation required)?

This seems fine at first glance although I'd be curious if there might be some issues that pop up on 32-bit with this that don't on 64-bit

fjarri:
> Why not just have the trait borrow &self in its methods?

To save on allocations when the new length happens to be the same as the old one, or in some cases when it's smaller (depends on the allocator).

fjarri:
> This seems fine at first glance although I'd be curious if there might be some issues that pop up on 32-bit with this that don't on 64-bit

That would actually make the behavior more uniform across platforms. Imagine you have a 100-bit integer and you call `shorten(90)`. Currently this would truncate on 32-bit platforms (to 96 bits), but keep the integer the same on 64-bit platforms.

fjarri:
An alternative `resize()` behavior would be to unconditionally clear all the high bits above `at_least_bits_precision`. This way it will be marginally faster, and `try_resize()` is not needed.

tarcieri (approved):

PR #816, opened 2025-05-01
Inversion method naming

fjarri:
Part of #693, namely https://github.com/RustCrypto/crypto-bigint/issues/693#issuecomment-2580788859

Renamed all `inv` methods and traits to `invert()`.

Public API:
- `inv() -> invert()`, 
- `inv_mod()` -> `invert_mod()`, 
- `InvMod` -> `InvertMod`
- `inv_odd_mod()` -> `invert_odd_mod()`
- `inv_mod2k()` -> `invert_mod2k()`

Internal:
- `inv_mod2_62()` -> `invert_mod2_62()`
- Rename `inv`/`inv_mod` modules to `invert`/`invert_mod`


tarcieri (approved):

tarcieri:
Wish I would've gotten this done in the last release so we could easily deprecate the old names and then get rid of the deprecations in the following release. Oh well

tarcieri:
> Note that the inv modules were not renamed - although they can be.

Might as well

tarcieri (approved):
Looks great now (approving again)

PR #817, opened 2025-05-01
Adopt `core` naming conventions for bigint methods

### LINKED ISSUE #693: Adopt `core` naming conventions for bigint methods

fjarri:
Fixes #693

Public API:
- Rename `WideningMul` -> `ConcatenatingMul`, `widening_mul()` -> `concatenating_mul()`
- Rename `mul_wide()`/`split_mul()` -> `widening_mul()`
- Rename `adc()` -> `carrying_add()`
- Rename `sbb()` -> `borrowing_sub()`
- Rename `mac()` -> `carrying_mul_add()` and change the order of arguments to match the `core` signature

Internal:
- Remove `primitives::mulhilo()` (same as `widening_mul()`)
- Change argument order in `addhilo()` to match the rest of the primitives
- Mark functions in `primitives` as `pub(crate)` (just in case)

Note that `widening_mul()` methods were not marked as deprecated since they had their behavior changed instead.

### DIFF
--- a/benches/int.rs
+++ b/benches/int.rs
@@ -60,54 +60,54 @@ fn bench_mul(c: &mut Criterion) {
     });
 }
 
-fn bench_widening_mul(c: &mut Criterion) {
+fn bench_concatenating_mul(c: &mut Criterion) {
     let mut rng = ChaChaRng::from_os_rng();
     let mut group = c.benchmark_group("widening ops");
 
-    group.bench_function("widening_mul, I128xI128", |b| {
+    group.bench_function("concatenating_mul, I128xI128", |b| {
         b.iter_batched(
             || (I128::random(&mut rng), I128::random(&mut rng)),
-            |(x, y)| black_box(x.widening_mul(&y)),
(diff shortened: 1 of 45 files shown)


tarcieri:
For these I would probably prefer leaving the old names with a `#[deprecated]` attribute to help people upgrade.

(We could arguably do that for #816 as well)

tarcieri (approved):

tarcieri:
I should note that the `widening_mul` function being effectively replaced by a new function might be a little confusing, but hopefully type system errors help people figure that out

fjarri:
The `concatenating_mul` name is still open for discussion until the 7.0 release, but at least it will be easy to find.

PR #820, opened 2025-05-13
adjust deprecation notice on mac

baloo:
This adjust the deprecation notice on `Limb::mac` to avoid confusion when upgrading crypto-bigint dependency.


baloo:
example that: https://github.com/RustCrypto/elliptic-curves/pull/1187#issuecomment-2870852508

tarcieri (approved):

PR #824, opened 2025-05-22
impl `to_be_bytes_trimmed` for `BoxedUint`

baloo:
This helpers strips the leading zeroes from the serialization.


tarcieri:
It would be good to add some tests, particularly around boundary conditions (e.g. empty, 1-bit, fits in 1-byte, needs 1-bit more than 1-byte).

This is a variable-time method and per crate policy it should probably have `vartime` in its name, though `to_be_bytes_trimmed_vartime` is quite long.

It would also be good to add a corresponding `to_le_bytes_trimmed` (or thereabouts) method which strips trailing zeros.

tarcieri (approved):

PR #831, opened 2025-05-30
v0.7.0-pre.4

tarcieri:
Release that includes the trimmed serialization API for the `rsa` crate


tarcieri:
@fjarri would it be possible to get a corresponding prerelease of `crypto-primes`? That would allow us to cut a crate (pre)release of `rsa`

tarcieri:
@fjarri thank you!

PR #833, opened 2025-06-01
Add `BoxedUint::{from_be_slice_vartime, from_le_slice_vartime}`

tarcieri:
Adds methods for decoding `BoxedUint` which infer the precision from the input size.

This is useful for decoding public parameters, especially when the size of other parameters is inferred from some public parameter that needs to be decoded first, e.g. RSA modulus.

### DIFF
--- a/src/uint/boxed/encoding.rs
+++ b/src/uint/boxed/encoding.rs
@@ -39,6 +39,20 @@ impl BoxedUint {
         Ok(ret)
     }
 
+    /// Create a new [`BoxedUint`] from the provided big endian bytes, automatically selecting its
+    /// precision based on the size of the input.
+    ///
+    /// This method is variable-time with respect to all subsequent operations since it chooses the
+    /// limb count based on the input size, and is therefore only suitable for public inputs.
+    ///
+    /// When working with secret values, use [`BoxedUint::from_be_slice`].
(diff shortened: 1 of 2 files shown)


tarcieri:
Not sure this is the greatest name or implementation. Notably the existing `BoxedUint::from_*_slice` methods are vartime to an extent.

These methods are unsuitable for use on secret values, however, so `*_vartime` felt appropriate.

src/uint/boxed/encoding.rs
```diff
@@ -39,6 +39,20 @@ impl BoxedUint {
...
 
+    /// Create a new [`BoxedUint`] from the provided big endian bytes, automatically selecting its
+    /// precision based on the size of the input.
+    ///
+    /// This method is variable-time with respect to all subsequent operations since it chooses the
+    /// limb count based on the input size, and is therefore only suitable for public inputs.
+    ///
+    /// When working with secret values, use [`BoxedUint::from_be_slice`].
+    pub fn from_be_slice_vartime(bytes: &[u8]) -> Self {
+        let bits_precision = bytes.len().saturating_mul(8) as u32;
```
baloo:
```suggestion
        let bits_precision = (bytes.len() as u32).saturating_mul(8);
```
(and same below)

Otherwise you would lose most of the benefits of saturating? (as usize would be an u64 on some platforms)

PR #837, opened 2025-06-11
`Int`-`Uint` conversion

erik-3milabs:
This PR introduces `Int::to_uint` and `Uint::to_int`, allowing better conversion between these types.

### DIFF
--- a/src/const_choice.rs
+++ b/src/const_choice.rs
@@ -395,7 +395,7 @@ impl<const LIMBS: usize> ConstCtOption<Uint<LIMBS>> {
     /// Returns the contained value, interpreting the underlying [`Uint`] value as an [`Int`].
     #[inline]
     pub const fn as_int(&self) -> ConstCtOption<Int<LIMBS>> {
-        ConstCtOption::new(self.value.as_int(), self.is_some)
+        ConstCtOption::new(*self.value.as_int(), self.is_some)
     }
 }
 
--- a/src/int.rs
+++ b/src/int.rs
@@ -154,9 +154,22 @@ impl<const LIMBS: usize> Int<LIMBS> {
         ConstCtOption::new(Odd(self), self.0.is_odd())
     }
 
-    /// Interpret the data in this type as a [`Uint`] instead.
+    /// Interpret the data in this object as a [Uint] instead.
+    ///
+    /// Note: this is a casting operation. See
+    /// - [Self::to_uint] for the checked equivalent, and
+    /// - [Self::abs] to obtain the absolute value of `self`.
     pub const fn as_uint(&self) -> &Uint<LIMBS> {
         &self.0
+    }
+
+    /// Get a [Uint] equivalent of this value; returns `None` if `self` is negative.
+    ///
+    /// Note: this is a checked conversion operation. See
+    /// - [Self::as_uint] for the unchecked equivalent, and
+    /// - [Self::abs] to obtain the absolute value of `self`.
+    pub const fn to_uint(self) -> ConstCtOption<Uint<LIMBS>> {
+        ConstCtOption::new(self.0, self.is_negative().not())
     }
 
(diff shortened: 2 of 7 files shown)


tarcieri:
Hmm, I'm used to `as_*` and `to_*` with similar names being symmetric operations which provide borrowed and owned conversions

erik-3milabs:
Wdyt of
- sticking with `as_(u)int` for the casting operation (i.e., similar to `u32 as i32`), and
- using `try_into_(u)int` for the conversion operation, which may fail, hence the `try_*`

erik-3milabs:
Hmmm, but `try_into_(u)int` is such a long name :(

And `to_(u)int` does match the `to_odd` and `to_nz` naming schemes :thinking:

tarcieri:
> using try_into_(u)int for the conversion operation, which may fail, hence the try_*

That sounds fine

> Hmmm, but try_into_(u)int is such a long name :(

It isn't that long

> And to_(u)int does match the to_odd and to_nz naming schemes 🤔

The problem is the asymmetry between the `as_*` and `to_*` conversions.

There is no `as_odd` or `as_nz`.

If you'd prefer API symmetry, the other fallible conversions could be renamed `try_into_*` (I do find it odd for `to_*` conversions to be fallible in the first place)

Edit: FWIW here are the API guidelines, however they don't mention fallibility specifically: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

erik-3milabs:
> the API guidelines

This was a very interesting read; thanks!


The following is the cleanest&fastest I can think of:

```rust
impl Int {
    pub fn as_uint(&self) -> &Uint;
    
    pub fn try_into_uint(self) -> ConstCtOption<Uint>;
}
```


```rust
impl Uint {
    pub fn as_int(&self) -> Int;
    
    pub fn try_into_int(self) -> ConstCtOption<Int>;
}
```

This breaks the naming convention in one tiny way, however: `Uint::as_int` is `borrowed/owned -> owned` instead of `borrowed -> borrowed` as the convention recommends.

ycscaly:
> > the API guidelines
> 
> This was a very interesting read; thanks!
> 
> The following is the cleanest&fastest I can think of:
> 
> ```rust
> impl Int {
>     pub fn as_uint(&self) -> &Uint;
>     
>     pub fn try_into_uint(self) -> ConstCtOption<Uint>;
> }
> ```
> 
>  ```rust
> impl Uint {
>     pub fn as_int(&self) -> Int;
>     
>     pub fn try_into_int(self) -> ConstCtOption<Int>;
> }
> ```
> 
> This breaks the naming convention in one tiny way, however: `Uint::as_int` is `borrowed/owned -> owned` instead of `borrowed -> borrowed` as the convention recommends.

In this case I think it is best to just drop `Uint::as_int`. It isn’t a borrowing and you have try_into_int as well. Why do we need it?

erik-3milabs:
> In this case I think it is best to just drop Uint::as_int. It isn’t a borrowing and you have try_into_int as well. Why do we need it?

The unchecked `as_int` conversion is significantly faster than `try_into_uint().unwrap()`

erik-3milabs:
We could perhaps make `&Uint::as_int() -> &Int` work similar to getting a `&NonZero` from a `&Odd`:

```rust
/// All odd integers are definitionally non-zero, so we can also obtain a reference to [`NonZero`].
pub const fn as_nz_ref(&self) -> &NonZero<T> {
    #[allow(trivial_casts, unsafe_code)]
    unsafe {
        &*(&self.0 as *const T as *const NonZero<T>)
    }
}
```

tarcieri:
@erik-3milabs that would be fine if you marked `Int` as `#[repr(transparent)]`

erik-3milabs:
@tarcieri let me know what you think!

PR #842, opened 2025-06-15
Fix deprecations

tarcieri:
These were introduced in #755 which was just merged

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,3 +80,6 @@ harness = false
 [[bench]]
 name = "int"
 harness = false
+
+[profile.dev]
+opt-level = 2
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -80,3 +80,6 @@ harness = false
 [[bench]]
 name = "int"
 harness = false
+
+[profile.dev]
+opt-level = 2
```
tarcieri, line 85: @erik-3milabs noticed the bingcd tests are a bit slow, this helped some but I'm curious if it could be sped up otherwise?

erik-3milabs: The primary reason for the slow tests seems to be that it is comparing the output of `bingcd` against that of the existing Bernstein-Yang `gcd` algorithm (which we're trying to replace because it is slow). 

The solution would be to refactor the test-suite so that it no longer leans on B-Y.

tarcieri: @erik-3milabs I would suggest refactoring those tests to use `proptest`.

You could also use `num-bigint` as a reference to compare to instead of B-Y.

erik-3milabs: Good ideas! I'll make sure to apply them :D

PR #848, opened 2025-06-19
Mention `Uint::carrying_mul_add` interface change in `CHANGELOG.md`

erik-3milabs:
This cost me more than an hour of my life to debug. Let's add this to the change log right now, so we don't accidentally forget it in the 0.7.0 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 _The below list is very much incomplete._
 
 ### Changed
-- Replace `Limb::mac` with `::carrying_add_mul` ([#817]).
-  - Note: `::carrying_add_mul` requires a different parameter order than `::mac`.
+- Replace `Limb::mac` with `::carrying_mul_add` ([#817]).
+  - Note: `::carrying_mul_add` requires a different parameter order than `::mac`.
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.7.0-pre.5 (2025-06-16)
+_The below list is very much incomplete._
+
+### Changed
+- Replace `Limb::mac` with `::carrying_add_mul` ([#817]).
+  - Note: `::carrying_add_mul` requires a different parameter order than `::mac`.
```
tarcieri: These should all be `carrying_mul_add`, right?

```suggestion
- Replace `Limb::mac` with `::carrying_mul_add` ([#817]).
  - Note: `::carrying_mul_add` requires a different parameter order than `::mac`.
```

tarcieri:
We also noted the argument order change in the deprecation message. I guess that wasn't helpful. Can we make it more visible?

https://github.com/RustCrypto/crypto-bigint/blob/c1978721bac5ea65ddac85f62980a87526058993/src/limb/mul.rs#L15

tarcieri (approved):

PR #846, opened 2025-06-17
Update `bingcd`

erik-3milabs:
This PR
- [x] refactors the `bingcd` code to better locations in the crate,
- [x] fixes a bug in `bingcd` (as shown in 110735a), and
- [x] support `Int::bingcd`
- [x] speeds up the `bingcd` test suite (as requested in #755)
- [x] Implement `bingcd_vartime`

PR #850, opened 2025-06-19
Add `#[ignore]` to bingcd tests

tarcieri:
In their current form these tests take quite awhile to execute.

Adding `#[ignore]` omits them from typical test runs so the test suite still executes quickly.

The main `test` job in CI has been configured to run them with `--ignored`.

### DIFF
--- a/src/uint/bingcd.rs
+++ b/src/uint/bingcd.rs
@@ -63,6 +63,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore] // TODO(tarcieri): improve performance
     fn test_bingcd() {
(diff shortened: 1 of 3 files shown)


tarcieri:
cc @erik-3milabs

src/uint/bingcd.rs
```diff
@@ -63,6 +63,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore] // TODO(tarcieri): improve performance
```
tarcieri, line 66: @erik-3milabs I can list you on these if you prefer (or we can open a tracking issue and put the issue number)

erik-3milabs: I'm already working on this in #846 :)

PR #854, opened 2025-06-24
Classic Binary Extended GCD

erik-3milabs:
This PR introduces the binary extended GCD algorithm, and all the boiler plate code necessary to use the xgcd algorithm.

Supersedes #761.

### DIFF
--- a/src/modular/bingcd/extension.rs
+++ b/src/modular/bingcd/extension.rs
@@ -1,4 +1,4 @@
-use crate::{ConstChoice, ConstCtOption, Int, Limb, Uint};
+use crate::{ConstChoice, ConstCtOption, Limb, Uint};
 
 pub(crate) struct ExtendedUint<const LIMBS: usize, const EXTENSION_LIMBS: usize>(
     Uint<LIMBS>,
@@ -6,6 +6,15 @@ pub(crate) struct ExtendedUint<const LIMBS: usize, const EXTENSION_LIMBS: usize>
 );
 
 impl<const LIMBS: usize, const EXTRA: usize> ExtendedUint<LIMBS, EXTRA> {
+    /// Construct an [ExtendedUint] from the product of a [Uint<LIMBS>] and an [Uint<EXTRA>].
+    ///
+    /// Assumes the top bit of the product is not set.
+    #[inline]
+    pub const fn from_product(lhs: Uint<LIMBS>, rhs: Uint<EXTRA>) -> Self {
+        let (lo, hi) = lhs.widening_mul(&rhs);
+        ExtendedUint(lo, hi)
+    }
+
     /// Interpret `self` as an [ExtendedInt]
     #[inline]
     pub const fn as_extended_int(&self) -> ExtendedInt<LIMBS, EXTRA> {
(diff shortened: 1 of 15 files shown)


erik-3milabs:
@tarcieri this PR is ready for review.

src/modular/bingcd/extension.rs
```diff
@@ -1,11 +1,20 @@
-use crate::{ConstChoice, ConstCtOption, Int, Limb, Uint};
+use crate::{ConstChoice, ConstCtOption, Limb, Uint};
 
 pub(crate) struct ExtendedUint<const LIMBS: usize, const EXTENSION_LIMBS: usize>(
     Uint<LIMBS>,
     Uint<EXTENSION_LIMBS>,
 );
 
 impl<const LIMBS: usize, const EXTRA: usize> ExtendedUint<LIMBS, EXTRA> {
+    /// Construct an [ExtendedUint] from the product of a [Uint<LIMBS>] and an [Uint<EXTRA>].
```
tarcieri, line 9: Using this syntax will linkify the resulting rustdoc:

```suggestion
    /// Construct an [`ExtendedUint`] from the product of a [`Uint<LIMBS>`] and an [`Uint<EXTRA>`].
```

tarcieri: Note: there seem to be several instances of this below, including ones I didn't catch before

src/modular/bingcd/xgcd.rs
```diff
@@ -99,62 +246,366 @@ impl<const LIMBS: usize> Odd<Uint<LIMBS>> {
...
+    /// Helper function to effectively test xgcd.
+    fn test_xgcd<const LIMBS: usize, const DOUBLE: usize>(
+        lhs: Uint<LIMBS>,
+        rhs: Uint<LIMBS>,
+        output: PatternXgcdOutput<LIMBS>,
+    ) where
+        Uint<LIMBS>: ConcatMixed<Uint<LIMBS>, MixedOutput = Uint<DOUBLE>>,
+    {
+        // Test the gcd
+        assert_eq!(lhs.bingcd(&rhs), output.gcd, "{} {}", lhs, rhs);
```
tarcieri:
```suggestion
        assert_eq!(lhs.bingcd(&rhs), output.gcd, "{lhs} {rhs}");
```

src/modular/bingcd/extension.rs
tarcieri: Going to approve this PR but it would be good to get any of these which are visible in public rustdoc fixed up

erik-3milabs: > Using this syntax will linkify the resulting rustdoc:

You're absolutely correct that it is good thing to add those...

... which I'll have to do that in the next PR, given that you already merged this one :see_no_evil:

erik-3milabs: (Also, on my end `[Uint<LIMBS>]` links to `Uint` and gets displayed as `Uint`, while
```
[`Uint<LIMBS>`]
```
still links to `Uint` but gets formatted as `Uint<LIMBS>`.

So, the backticks just add the `LIMBS`, but not the link.)

erik-3milabs: Also, there is a difference in the formatting. Adding backticks also formats the word as code, instead of just text.

dolevmu (commented):
Nice!

PR #855, opened 2025-06-24
`Xgcd` trait

erik-3milabs:
This PR 
- transitions `Gcd for Uint/Int` from `safegcd` to `bingcd`
- introduces the `Xgcd` trait, and
- implements the `Xgcd for Uint/Int`.


erik-3milabs:
@tarcieri this PR is ready for review.

src/uint/gcd.rs
```diff
@@ -1,149 +1,538 @@
...
+    /// threshold. When using [Uint]s with `LIMBS` close to the threshold, it may be useful to
+    /// manually test whether the classic or optimized algorithm is faster for your machine.
+    #[inline(always)]
+    pub const fn bingcd_vartime(&self, rhs: &Uint<LIMBS>) -> Self {
+        if LIMBS < 4 {
+            self.classic_bingcd_vartime(rhs)
+        } else {
+            self.optimized_bingcd_vartime(rhs)
+        }
+    }
```
tarcieri, line 185: Since we're no longer disambiguating between safegcd and bingcd, perhaps these should just be called `gcd` and `gcd_vartime`?

erik-3milabs: I'm inclined to agree that we _can_ do this. I don't know whether it is the best solution, however.

Renaming `bingcd` to `gcd` means that there would be two `gcd` functions available for `Uint`: 
- one through the `impl Uint`, and
- one through the `GCD` trait. 

By keeping things the way they are now, we can implement the `GCD` trait for every possible combination of `Uint` and `Int`, making life easy for the caller, while also allowing those who wish to have access to the `const` functions to do their thing. (<- I see now that this statement makes the underlying assumption that there is a benefit to having access to the `const` function, while I don't know for sure whether there is such a benefit)

tarcieri: Having two functions with the same name, where one is an inherent `const fn` and the other a trait method, is a pretty common pattern in this library, but usually the trait method calls the `const fn`, and otherwise the two are identical.

It looks like perhaps that isn't the case in this PR, and if not, why not?

tarcieri: I'll note the `const` access is quite important, particularly if we use XGCD to implement inversions, as we use inversions to compute various constants at compile time in the https://github.com/RustCrypto/elliptic-curves crates.

We can go with this for now though I may investigate preserving the existing names in a followup PR so long as the `const fn` and trait methods are identical

PR #857, opened 2025-06-24
`NonZeroInt::new_unwrap`

erik-3milabs:
plus some minor clean-up

### DIFF
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -137,7 +137,7 @@ impl NonZero<Limb> {
     }
 }
 
-impl<const LIMBS: usize> NonZero<Uint<LIMBS>> {
+impl<const LIMBS: usize> NonZeroUint<LIMBS> {
     /// Creates a new non-zero integer in a const context.
     /// Panics if the value is zero.
     ///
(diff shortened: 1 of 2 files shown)


src/non_zero.rs
```diff
@@ -111,7 +111,7 @@ impl NonZero<Limb> {
         }
     }
 
-    /// Create a [`NonZero<Limb>`] from a [`NonZeroU8`] (const-friendly)
+    /// Create a [`NonZero<Limb>`]from a [`NonZeroU8`] (const-friendly)
```
tarcieri: Looks like you're removing a space that should probably be there

tarcieri (commented):
Looks good aside from the removed space

src/non_zero.rs
erik-3milabs:
```suggestion
    /// Create a [`NonZero<Limb>`] from a [`NonZeroU8`] (const-friendly)
```

erik-3milabs: fixed!

tarcieri (approved):

PR #859, opened 2025-07-03
Fix clippy for Rust v1.88

erik-3milabs:


erik-3milabs:
@tarcieri please review and merge if you're on board: this is blocking #846 and subsequent PRs.

tarcieri (approved):

tarcieri:
Hmm, usually we pin `clippy` to a specific Rust version so these sorts of breakages don't happen but I guess it's set to `stable` in this repo

PR #872, opened 2025-07-18
DRY out Monty params between const and non-const forms

tarcieri:
Creates a `modular::params` module containing a set of `const fn` functions for computing Montgomery params, and uses these functions for both the `const_monty_params!` macro as well as `MontyParams`.

Ideally we could just use `MontyParams` for this (as it has a `const fn` constructor) and eventually this can probably be factored there (i.e. like it was before this PR). However, doing that properly would require `adt_const_params` (rust-lang/rust#95174), which is why we have `const_monty_params!` in the first place, so for now this is the best we can do.

### DIFF
--- a/src/modular.rs
+++ b/src/modular.rs
@@ -16,6 +16,8 @@
 //! The [`MontyForm`] and [`MontyParams`] types implement support for modular arithmetic where
 //! the modulus can vary at runtime.
 
+pub mod params;
+
 mod const_monty_form;
 mod lincomb;
 mod monty_form;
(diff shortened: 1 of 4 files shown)


tarcieri:
One possible alternative to this that could eliminate the need for a macro and DRY things out even further is to change the `ConstMontyParams` trait to have an associated `const PARAMS: MontyParams`, using `MontyParams::new` to construct them

tarcieri:
Closing in favor of #873

PR #873, opened 2025-07-18
Unify `ConstMontyParams` and `MontyParams`

tarcieri:
Changes `ConstMontyParams` from having associated constants that duplicate `MontyParams` to having an associated `MontyParams` constant.

`MontyParams` has `const fn` constructors, which the `const_monty_params!` (and legacy `impl_modulus!`) macros have been changed to use.

This gets us a bit closer to the ideal solution which would leverage the `adt_const_params` feature (rust-lang/rust#95174), as that would allow `ConstMontyForm` to be generic around a `MontyParams` constant.


src/modular/monty_form.rs
```diff
@@ -22,18 +22,18 @@ use subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
...
     /// `R^3 mod modulus`, used to compute the multiplicative inverse
-    r3: Uint<LIMBS>,
+    pub(super) r3: Uint<LIMBS>,
     /// The lowest limbs of -(MODULUS^-1) mod R
     /// We only need the LSB because during reduction this value is multiplied modulo 2**Limb::BITS.
-    mod_neg_inv: Limb,
+    pub(super) mod_neg_inv: Limb,
     /// Leading zeros in the modulus, used to choose optimized algorithms
-    mod_leading_zeros: u32,
+    pub(super) mod_leading_zeros: u32,
```
tarcieri, line 36: Alternatively we could add `pub(crate)` reader methods, similar to the ones that already exist on `BoxedMontyParams`

tarcieri: I think that is probably the way to go, but maybe as a followup

PR #882, opened 2025-07-19
Rename `div_by_2` to `shr1`

tarcieri:
For a few reasons:
- It's more consistent with the existing uses of `shr1` in the codebase
- It's implemented in terms of `shr1` already
- With a little work it can be adapted into a full-fledged `shr`

This preserves the `div_by_2*` names with deprecations

### DIFF
--- a/benches/monty.rs
+++ b/benches/monty.rs
@@ -180,13 +180,13 @@ fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
         )
     });
 
-    group.bench_function("div_by_2, U256", |b| {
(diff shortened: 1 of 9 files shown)


tarcieri:
I know this is six-of-one, half dozen of another, but we should pick a name to use consistently.

I might also take a stab at doing the conversion to a full fledged `shr` instead of merging this.

tarcieri:
On second thought, I'm going to merge this because it's a fairly trivial change, and a backwards compatible one due to the use of `deprecate`.

That said, I think the same logic should apply to to arbitrary right shifts, and a PR to implement that will probably be easier to review if it's extending `shr1` to `shr(n)`

PR #886, opened 2025-07-24
Update safegcd GCD and modular inversion

### LINKED ISSUE #634: Implement `safegcd-bounds`
This is a corresponding tracking issue for this TODO: https://github.com/RustCrypto/crypto-bigint/blob/cdc8487/src/modular/bernstein_yang.rs#L341

The bounds we currently implement for Bernstein-Yang are the ones described in the paper, which proves that the algorithm will always converge within the prescribed bounds. However, the bounds are overly conservative and not optimal:

https://github.com/sipa/safegcd-bounds

There is both an improved bounds calculation we can use, as well as an improved divsteps algorithm (`hddivsteps`) which itself has improved bounds over the original divsteps algorithm.

andrewwhitehead:
This updates the safegcd implementation to remove the extra trait bounds on the precomputed inverter and improve the performance. It implements the safegcd-bounds half-delta step and bounds (https://github.com/sipa/safegcd-bounds).

For the smallest sizes of Uint (U64 and maybe U128) it is currently slightly slower, but the speed improves with larger sizes. For U256 is it approximately 50% faster than the old implementation or bingcd. `invert_odd_mod` is over 90% faster for U256.

Fixes #634


src/modular/safegcd.rs
```diff
@@ -46,536 +49,584 @@ use subtle::CtOption;
...
+            adjuster: *adjuster,
+            inverse: invert_mod_u64(modulus.0.as_words()),
         }
     }
 
     /// Returns either the adjusted modular multiplicative inverse for the argument or `None`
     /// depending on invertibility of the argument, i.e. its coprimality with the modulus.
     #[deprecated(since = "0.7.0", note = "please use `invert` instead")]
-    pub const fn inv(&self, value: &Uint<SAT_LIMBS>) -> ConstCtOption<Uint<SAT_LIMBS>> {
+    pub fn inv(&self, value: &Uint<LIMBS>) -> ConstCtOption<Uint<LIMBS>> {
```
tarcieri: Not sure why you removed `const` here?

src/modular/safegcd.rs
```diff
@@ -46,536 +49,584 @@ use subtle::CtOption;
...
+    let (x, y) = (x.wrapping_mul(y.wrapping_add(1)), y.wrapping_mul(y));
+    let (x, y) = (x.wrapping_mul(y.wrapping_add(1)), y.wrapping_mul(y));
+    x.wrapping_mul(y.wrapping_add(1))
+}
 
-        ret
-    }
+/// A `Uint` which carries a separate sign in order to maintain the same range.
+#[derive(Clone, Copy)]
+struct Sint<const LIMBS: usize> {
```
tarcieri: Not wild about that name, but I guess it disambiguates from `Int`

src/modular/safegcd.rs
```diff
@@ -46,536 +49,584 @@ use subtle::CtOption;
 /// - P. Wuille, "The safegcd implementation in libsecp256k1 explained",
 ///   <https://github.com/bitcoin-core/secp256k1/blob/master/doc/safegcd_implementation.md>
 #[derive(Clone, Debug)]
-pub struct SafeGcdInverter<const SAT_LIMBS: usize, const UNSAT_LIMBS: usize> {
+pub struct SafeGcdInverter<const LIMBS: usize> {
```
tarcieri, line 54: 🎉

src/modular/safegcd.rs
andrewwhitehead: Oops, fixed.

andrewwhitehead:
It looks like I need to do a bit more testing on the modular inversion, there could be an issue with the normalization. I haven't seen a failure on my machine yet.

andrewwhitehead:
Fixed the issue on 32 bit platforms.

I renamed `invert_mod2_62` and adjusted it to return a u64 so that it can potentially be used in `MontyParams`. If that stored the inverted modulus as a U64 I think it could be reused in inversion and Montgomery reduction.

src/modular/safegcd.rs
tarcieri: How about `SignedInt`?

andrewwhitehead:
I did notice one edge case in `invert_odd_mod`, currently `(0).invert_odd_mod(1)` returns 0, but the old implementation returns 1. Neither is wrong exactly (`0 * 1 = 1 mod 1` and `0 * 0 = 1 mod 1`) but maybe it should just fail.

tarcieri:
@andrewwhitehead which type are we talking about here? I would've expected that case to return a `None`-ish `CtOption`-like type

andrewwhitehead:
That's with Uint, there's a unit test for it currently. I'll see if there's an easy way to fix it up.

erik-3milabs:
Congrats on this excellent PR! :partying_face: `safegcd` is a significant upgrade to `bingcd`. Hats off! I'm glad to see we managed to speed up `gcd` by ~50x in a timespan of ~9mo. Solid stuff!

In terms of tweaking, we might be able to win a couple percent here and there:
On my end, `safegcd` is faster than `bingcd` for as little as 3 limbs, so not much to be gained there.
However, `bingcd_vartime` outperforms `safegcd` by ~30% for up to 7 limbs and by ~15% up to 16 limbs. 
Moreover, `safegcd` outperforms `safegcd_vartime` for 32+ limbs (but perhaps this only applies to multiples of two), so using the constant time algorithm might actually be faster than the vartime algorithm.

tarcieri:
@erik-3milabs I was meaning to ask if you thought we still needed `bingcd` after this, so glad to hear your thoughts. It would be nice, if only for compile times, if we only had one implementation of GCD/inversions.

Kind of surprised to hear that about the vartime implementation however, as I would always expect it would do fewer iterations than the constant-time version.

Edit: opened #905 to discuss the future of bingcd

PR #887, opened 2025-07-25
Add `Reduce` trait for fixed modulus reduction.

carloskiki:
This trait comes from [`elliptic_curve::Reduce`](https://docs.rs/elliptic-curve/latest/elliptic_curve/ops/trait.Reduce.html). I thought this would fit better here as `cyrpto-bigint` already hosts a lot of [`modular`](https://docs.rs/crypto-bigint/latest/crypto_bigint/modular/index.html) arithmetic functionality.

Note that it slightly differs from `elliptic_curve::Reduce` because the input type is not bound by `Integer`, and there is no associated `Bytes` (and no `reduce_bytes` function). That is because `Reduce` can be implemented both on the input type and on `Bytes` as separate implementations, if the input type is not bounded by `Integer`.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -699,6 +699,19 @@ pub trait RemMixed<Reductor>: Sized {
     fn rem_mixed(&self, reductor: &NonZero<Reductor>) -> Reductor;
 }
 
+/// Modular reduction on a fixed modulus.
+///
+/// Used when there is a fixed modulus for which it is natural to reduce the value to.
+///
+/// For modular reduction with a variable modulus, use [`Rem`].
+pub trait Reduce<Modulus> {
+    /// Output type.
+    type Output;
+
+    /// Reduces `self` modulo `Modulus`.
+    fn reduce(&self) -> Self::Output;
+}
+
 /// Division in variable time.
 pub trait DivVartime: Sized {
     /// Computes `self / rhs` in variable time.
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -699,17 +699,15 @@ pub trait RemMixed<Reductor>: Sized {
     fn rem_mixed(&self, reductor: &NonZero<Reductor>) -> Reductor;
 }
 
-/// Modular reduction on a fixed modulus.
+/// Modular reduction from a larger value `T`.
 ///
-/// Used when there is a fixed modulus for which it is natural to reduce the value to.
+/// This can be seen as fixed modular reduction, where the modulus is fixed at compile time
+/// by `Self`.
 ///
 /// For modular reduction with a variable modulus, use [`Rem`].
-pub trait Reduce<Modulus> {
-    /// Output type.
-    type Output;
-
+pub trait Reduce<T>: Sized {
     /// Reduces `self` modulo `Modulus`.
-    fn reduce(&self) -> Self::Output;
+    fn reduce(value: T) -> Self;
 }
 
 /// Division in variable time.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 0.7.0-pre.5 (2025-06-16)
 _The below list is very much incomplete._
 
+### Added
+- `Reduce` trait for modular reduction with a constant modulus.
+
 ### Changed
 - Replace `Limb::mac` with `::carrying_mul_add` ([#817]).
   - Note: `::carrying_mul_add` requires a different parameter order than `::mac`.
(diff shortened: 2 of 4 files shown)


carloskiki:
Additional note: This trait is semantically identical to [`From`](https://doc.rust-lang.org/stable/std/convert/trait.From.html), but `From` requires that:
> The conversion is lossless: semantically, it should not lose or discard information.

Hence why we need a separate trait.

tarcieri:
I think it's a little odd just to move this trait here for the sake of moving it here (i.e. because reduction "feels" more fundamental than elliptic curves). It would make more sense if `crypto-bigint` consumed it in some way, e.g. providing generic reduction impls on something like `MontyForm`/`ConstMontyForm`.

It would also be good to see PRs to the `elliptic-curve` crate, `ecdsa`, and https://github.com/RustCrypto/elliptic-curves which consume it, so we can see how it works in practice.

> there is no associated Bytes (and no reduce_bytes function). That is because Reduce can be implemented both on the input type and on Bytes as separate implementations, if the input type is not bounded by Integer.

That doubles the number of impls required, but I guess halves the number of methods, so a reasonable tradeoff.

src/traits.rs
```diff
@@ -699,6 +699,17 @@ pub trait RemMixed<Reductor>: Sized {
...
 
+/// Modular reduction from a larger value `T`.
+///
+/// This can be seen as fixed modular reduction, where the modulus is fixed at compile time
+/// by `Self`.
+///
+/// For modular reduction with a variable modulus, use [`Rem`].
+pub trait Reduce<T>: Sized {
+    /// Reduces `self` modulo `Modulus`.
+    fn reduce(value: T) -> Self;
```
tarcieri, line 710: I'd suggest a borrow here. It's a transformation of the input, not something that takes ownership. It's also potentially a secret, so enforcing a borrow leaves the original value zeroizable.

See also: https://rust-lang.github.io/api-guidelines/flexibility.html?highlight=clone#caller-decides-where-to-copy-and-place-data-c-caller-control

```suggestion
    fn reduce(value: &T) -> Self;
```

carloskiki: I'm sort of on the fence. We can always `impl` for a reference which achieves the same thing as taking a `&` as input, but also allows us to take things by value if we want. I agree however that `reduce` will mostly take references to its input.

carloskiki: Actually, here are some examples in `RustCrypto` that take things by value in reduce operations:

- https://docs.rs/p256/0.14.0-pre.9/p256/struct.Scalar.html#method.reduce
- https://docs.rs/p384/0.14.0-pre.9/p384/struct.Scalar.html#method.reduce

If we keep the input by value we can do both these implementations and implementations on referenced byte slices or arrays.

erik-3milabs: It seems to me like both P256 and P384 would be OK with being passed a reference instead.

Their current sources [[P256](https://docs.rs/p256/0.14.0-pre.9/src/p256/arithmetic/scalar.rs.html#626-630)] 
```rust
impl Reduce<U256> for Scalar {
    fn reduce(w: U256) -> Self {
        let (r, underflow) = w.borrowing_sub(&NistP256::ORDER, Limb::ZERO);
        let underflow = Choice::from((underflow.0 >> (Limb::BITS - 1)) as u8);
        Self(U256::conditional_select(&w, &r, !underflow))
    }
}
```
and [[P384](https://docs.rs/p384/0.14.0-pre.9/src/p384/arithmetic/scalar.rs.html#233-237)]
```rust
impl Reduce<U384> for Scalar {
    fn reduce(w: U384) -> Self {
        let (r, underflow) = w.borrowing_sub(&NistP384::ORDER, Limb::ZERO);
        let underflow = Choice::from((underflow.0 >> (Limb::BITS - 1)) as u8);
        Self::from_uint_unchecked(U384::conditional_select(&w, &r, !underflow))
    }
}
```

carloskiki: Indeed. The claim is that having the function signature by value allows both `impl Reduce<U384>` and `impl Reduce<&U384>`, whereas the proposed change invalidates `impl Reduce<U384>` without adding anything.

I'm not against changing it, I just want to make sure this is the correct representation.

tarcieri: Since it immediately reborrows, providing an owned value doesn't do anything except let you omit the borrow. I don't see any value in it other than that.

It has the drawback that you can't zeroize the input value, which is perhaps a notable omission if anything. And it doubles the number of impls.

I think we should just change it to borrow everywhere. There's nothing the function can do with the owned value it can't do with the borrowed value.

tarcieri (approved):
I'll merge this, but I do intend on changing it to borrow its argument.

carloskiki:
Sorry! this went under my radar for some reason.

PR #891, opened 2025-07-25
Update GCD and modular inversion for BoxedUint

andrewwhitehead:
This brings them in line with the implementation for Uint/Int. On my machine `gcd` is about 98% faster for the benchmarked sizes and `invert_odd_mod` about 95%. `BoxedMontyForm::invert()` also benefits.


src/modular/safegcd/boxed.rs
```diff
@@ -33,572 +33,412 @@ impl BoxedSafeGcdInverter {
     /// Modulus must be odd. Returns `None` if it is not.
     pub fn new(modulus: &Odd<BoxedUint>, adjuster: &BoxedUint) -> Self {
         Self {
-            modulus: BoxedUnsatInt::from(&modulus.0),
-            adjuster: BoxedUnsatInt::from(&adjuster.resize(modulus.bits_precision())),
-            inverse: invert_mod_u64(modulus.0.as_words()) as i64,
+            modulus: modulus.clone(),
```
tarcieri: If you're just cloning the modulus then we should probably just get rid of the borrow and have the caller clone if need be

andrewwhitehead: Updated now

PR #892, opened 2025-07-28
Make Int % Uint return a Uint

### LINKED ISSUE #847: Bug in `Int::div_rem_uint`
The following code fails an internal `expect`:

```rust
let num = I256::from_be_hex("eed8f6c7a5b1a65031ebc9b7a93492e89f282d49e75d25607a5693b3d8ae2e87");
let denom = U128::from_be_hex("de157f812521c55eccf9a903b31e4a34").to_nz().unwrap();
let (q, r) = num.div_rem_uint(&denom);
```

The author of the `div_rem_uint` function (=me) incorrectly assumed that the remainder of an `int`/`uint` will always fit in an `int`.

Edit:  Actually, I was correct about this, until we modified `Uint::div_rem` to accept numerators and denominators with different LIMBS-sizes.

andrewwhitehead:
Fixes #847

The result of taking a signed integer modulo an unsigned one should be unsigned. Python agrees at least:

```python3
>>> hex(-0x112709385A4E59AFCE14364856CB6D1760D7D2B618A2DA9F85A96C4C2751D179 % 0xde157f812521c55eccf9a903b31e4a34)
'0x44bfe3176459710d0d0848f99b37e47b'
>>> hex(0x112709385A4E59AFCE14364856CB6D1760D7D2B618A2DA9F85A96C4C2751D179 % 0xde157f812521c55eccf9a903b31e4a34)
'0x99559c69c0c85451bff1600a17e665b9'
```

This also removes the `RemAssign` implementations, `Int %= Uint`. If somebody needs to do this, I think they can do a checked conversion to an `Int` for the modulus first.

### DIFF
--- a/src/int/div_uint.rs
+++ b/src/int/div_uint.rs
@@ -1,14 +1,14 @@
 //! Operations related to dividing an [`Int`] by a [`Uint`].
-use core::ops::{Div, DivAssign, Rem, RemAssign};
+use core::ops::{Div, DivAssign, Rem};
 
 use crate::{ConstChoice, Int, NonZero, Uint, Wrapping};
 
 /// Checked division operations.
 impl<const LIMBS: usize> Int<LIMBS> {
-    #[inline]
     /// Base div_rem operation on dividing an [`Int`] by a [`Uint`].
     /// Computes the quotient and remainder of `self / rhs`.
     /// Furthermore, returns the sign of `self`.
+    #[inline]
     const fn div_rem_base_uint<const RHS_LIMBS: usize>(
         &self,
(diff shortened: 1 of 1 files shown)


fjarri:
I think in Rust's terms that makes it `rem_euclid` rather than `rem`.

andrewwhitehead:
> I think in Rust's terms that makes it `rem_euclid` rather than `rem`.

Yes, it's also possible to rename these methods to `rem_euclid` and add a separate `RemEuclid` trait. A `checked_rem` method that returned a `CtOption<Int<LIMBS>>` would also be possible (checking if the modulus fits in an Int).

PR #893, opened 2025-07-28
Update modular inverse calculation in MontyParams, BoxedMontyParams

andrewwhitehead:
This reuses `invert_mod_u64` to calculate the modular inverse, which improves performance when creating new MontyParams and BoxedMontyParams by around 20%. `Uint::mul_mod` is similarly improved.

### DIFF
--- a/src/modular/monty_form.rs
+++ b/src/modular/monty_form.rs
@@ -161,7 +161,7 @@ impl<const LIMBS: usize> zeroize::Zeroize for MontyParams<LIMBS> {
         self.one.zeroize();
         self.r2.zeroize();
         self.r3.zeroize();
-        self.mod_neg_inv.zeroize();
+        self.mod_inv.zeroize();
         self.mod_leading_zeros.zeroize();
     }
 }


tarcieri (approved):

src/modular/boxed_monty_form.rs
```diff
@@ -33,9 +36,9 @@ struct BoxedMontyParamsInner {
     r2: BoxedUint,
     /// R^3, used to compute the multiplicative inverse
     r3: BoxedUint,
-    /// The lowest limbs of -(MODULUS^-1) mod R
-    /// We only need the LSB because during reduction this value is multiplied modulo 2**Limb::BITS.
-    mod_neg_inv: Limb,
+    /// The lowest limbs of MODULUS^-1 mod 2**64
+    /// This value is used in Montgomery reduction and modular inversion
+    mod_inv: U64,
```
fjarri, line 41: Are there any performance regressions on 32-bit platforms because of this?

andrewwhitehead: I doubt it, the `Limb` is still used in Montgomery reduction. It just adds a `u32::wrapping_neg`.

PR #894, opened 2025-07-28
Remove `PrecomputedInverter` and `Inverter` traits

tarcieri:
`PrecomputedInverter` was largely a hack around the previous implementation of safegcd, namely that it used a 62-bit unsaturated implementation, which needed to be bounded by a different const generic parameter `UNSAT_LIMBS`. Namely the `Invert` impl was bounded on the presence of `PrecomputedInverter`, and the latter was only available for certain limb sizes.

Now that safegcd no longer uses `UNSAT_LIMBS`, the traits and all of the incidental complexity therein is unnecessary and can all be removed.

cc @andrewwhitehead


tarcieri:
I think we can go farther than this and make the inverter types private now, but ripping out these traits seems like a start

andrewwhitehead:
I don't think the inverters really need to exist, the places they are called can just call the internal methods directly.

src/modular/boxed_monty_form/invert.rs
```diff
@@ -22,22 +22,8 @@ impl BoxedMontyForm {
     }
 }
 
-impl Invert for BoxedMontyForm {
```
andrewwhitehead: I think this should still implement Invert?

tarcieri: Whoops, good catch

tarcieri:
@andrewwhitehead by that I even mean `SafeGcdInverter` and `BoxedSafeGcdInverter`

PR #896, opened 2025-07-28
Add Uint::rem_wide

andrewwhitehead:
This adds a constant time `rem_wide` method (based on `rem_wide_vartime` and `div_rem`). It is used to speed up `MontyParams::new` about 30% and `Uint::mul_mod` about 75% while removing some annoying trait bounds.


andrewwhitehead:
Adding some tests against the new method and I think this needs a minor fix.

andrewwhitehead:
For the `r2` calculation it would be much cleaner to have `square_mod` and `square_mod_vartime` methods. I'll likely follow up with that. It does imply a `SquareMod` trait which appears difficult to efficiently implement for `BoxedUint` at the moment.

src/uint/div.rs
```diff
@@ -1158,19 +1268,22 @@ mod tests {
...
+        let rem = U256::rem_wide(lo_hi, &modulus);
         // Lower half is zero
         assert_eq!(rem.to_be_bytes()[0..16], U128::ZERO.to_be_bytes());
         // Upper half
         let expected = U128::from_be_hex("203F80FE03F80FE03F80FE03F80FE041");
         assert_eq!(rem.to_be_bytes()[16..], expected.to_be_bytes());
+
+        let remv = U256::rem_wide_vartime(lo_hi, &modulus);
+        assert_eq!(rem, remv);
     }
```
tarcieri, line 1287: Wonder if we could capture these in `uint.proptest-regressions`

PR #899, opened 2025-07-30
Test reducing `p - 1`

daxpedda:
I'm surprised this doesn't work. Am I missing something?

### DIFF
--- a/src/modular.rs
+++ b/src/modular.rs
@@ -173,6 +173,18 @@ mod tests {
     }
 
     #[test]
+    fn test_reducing_close() {
+        assert_eq!(
+            montgomery_reduction::<{ Modulus2::LIMBS }>(
+                &(Modulus2::PARAMS.modulus.0 - Uint::ONE, Uint::ZERO),
+                &Modulus2::PARAMS.modulus,
+                Modulus2::PARAMS.mod_neg_inv()
+            ),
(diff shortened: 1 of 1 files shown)


daxpedda:
Interestingly enough this seems to work just fine:
```rust
assert_eq!(
    (Modulus2::PARAMS.modulus.0 - Uint::ONE)
        .rem(&Modulus2::PARAMS.modulus.to_nz().unwrap()),
    Modulus2::PARAMS.modulus.0 - Uint::ONE
);
```
So something seems to be wrong with the montgomery reduction algorithm, where it fails when you are very close to the modulus.

daxpedda:
Yeah, false alarm, my bad:
```rust
let x = Modulus2::PARAMS.modulus.0 - Uint::ONE;
let x_mod = ConstMontyForm::<Modulus2, { Modulus2::LIMBS }>::new(&x);

// Confirm that when creating a Modular and retrieving the value, that it equals the original
assert_eq!(x, x_mod.retrieve());
```
This works just fine, I misinterpreted how `ConstMontyForm::from_montgomery()` is supposed to work.

tarcieri:
`montgomery_reduction` can probably be removed from the public API now that it is no longer used by the `const_monty_params!` macro

PR #902, opened 2025-07-30
Update `BoxedUint::div_rem` to accept mixed sized inputs

andrewwhitehead:
Allowing mixed size inputs means that `rem` can be performed on 'wide' inputs without resizing the modulus, improving performance. The updated method can also be used to remove restrictions on `BoxedUint::mul_mod` but that's another update.

### DIFF
--- a/src/uint/boxed/div.rs
+++ b/src/uint/boxed/div.rs
@@ -3,6 +3,7 @@
 use crate::{
     BoxedUint, CheckedDiv, ConstChoice, ConstantTimeSelect, DivRemLimb, DivVartime, Limb, NonZero,
     Reciprocal, RemLimb, RemMixed, Wrapping,
+    const_choice::u32_min,
     uint::{
         boxed,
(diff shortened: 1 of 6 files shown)


src/uint/boxed/div.rs
```diff
@@ -135,71 +136,77 @@ impl BoxedUint {
...
         // Further explanation at https://janmr.com/blog/2014/04/basic-multiple-precision-long-division/
 
-        let size = self.limbs.len();
-        assert_eq!(
-            size,
-            rhs.limbs.len(),
-            "the precision of the divisor must match the dividend"
-        );
+        let xsize = self.nlimbs();
+        let ysize = rhs.nlimbs();
```
tarcieri, line 140: Hmm, not super wild about these names as they don't match the inputs

tarcieri: Oh I see, I guess they eventually do later in the function

PR #904, opened 2025-08-02
`ConstMontyForm::lincomb_vartime()` is constant-time

daxpedda:
I'm unsure if I interpreted [this line](https://docs.rs/crypto-bigint/0.7.0-pre.7/crypto_bigint/modular/struct.ConstMontyForm.html#method.lincomb_vartime) correctly:
> This method is variable time only with the value of the modulus.

But it seems to me this is actually constant-time if the modulus is the same.

I'm not sure if this shouldn't actually apply to `MontyForm` and `BoxedMontyForm` as well. Because while they set their modulus during runtime, it still doesn't change between multiple calls to the same instantiation (AFAIU).

### DIFF
--- a/src/modular/const_monty_form/lincomb.rs
+++ b/src/modular/const_monty_form/lincomb.rs
@@ -7,10 +7,7 @@ use crate::modular::lincomb::lincomb_const_monty_form;
 
 impl<MOD: ConstMontyParams<LIMBS>, const LIMBS: usize> ConstMontyForm<MOD, LIMBS> {
     /// Calculate the sum of products of pairs `(a, b)` in `products`.
-    ///
-    /// This method is variable time only with the value of the modulus.
-    /// For a modulus with leading zeros, this method is more efficient than a naive sum of products.
-    pub const fn lincomb_vartime(products: &[(Self, Self)]) -> Self {
+    pub const fn lincomb(products: &[(Self, Self)]) -> Self {
         Self {
             montgomery_form: lincomb_const_monty_form(
                 products,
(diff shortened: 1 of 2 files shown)


tarcieri:
cc @andrewwhitehead

andrewwhitehead:
The description could use some elaboration, but yes it is constant time for a fixed modulus. It is named `_vartime` as the modulus may be secret in some cases.

tarcieri:
@andrewwhitehead this is the const form though, where the modulus is a constant and also a source code literal, which I certainly hope means it isn't a secret!

andrewwhitehead:
Yes, I suppose it makes sense to change that one.

daxpedda:
[The documentation also says](https://docs.rs/crypto-bigint/0.7.0-pre.7/crypto_bigint/modular/struct.ConstMontyForm.html#method.lincomb_vartime):
> For a modulus with leading zeros, this method is more efficient than a naive sum of products.

However I just confirmed with benchmarks that there is also a significant performance increase with a modulus without leading zeros. So I'm not sure what this is supposed to mean, or maybe its just outdated.

I propose we remove this line from the documentation as well?

src/modular/const_monty_form/lincomb.rs
```diff
@@ -8,9 +8,8 @@ use crate::modular::lincomb::lincomb_const_monty_form;
 impl<MOD: ConstMontyParams<LIMBS>, const LIMBS: usize> ConstMontyForm<MOD, LIMBS> {
     /// Calculate the sum of products of pairs `(a, b)` in `products`.
     ///
-    /// This method is variable time only with the value of the modulus.
     /// For a modulus with leading zeros, this method is more efficient than a naive sum of products.
```
tarcieri:
```suggestion
```

src/modular/const_monty_form/lincomb.rs
```diff
@@ -8,9 +8,8 @@ use crate::modular::lincomb::lincomb_const_monty_form;
 impl<MOD: ConstMontyParams<LIMBS>, const LIMBS: usize> ConstMontyForm<MOD, LIMBS> {
     /// Calculate the sum of products of pairs `(a, b)` in `products`.
     ///
```
tarcieri:
```suggestion
```

tarcieri (commented):
Let's remove that comment

src/modular/const_monty_form/lincomb.rs
daxpedda: Done.

PR #919, opened 2025-08-10
Faster constant time bit shifts

andrewwhitehead:
This update adds internal `bounded_wrapping_sh(r/l)` and `wrapping_sh(r/l)_by_limbs` methods and optimizes `shr` and `shl` for `Uint` by applying sub-limb and super-limb shifts in different ways. Performance is generally improved (a lot), with division methods, `sqrt`, `bingcd`, and `MontyParams::new` also being particularly affected.

There is a variant of the bounded-wrapping shr method in `bingcd` which can be updated to use the new methods later.

### DIFF
--- a/src/uint/shl.rs
+++ b/src/uint/shl.rs
@@ -1,6 +1,6 @@
 //! [`Uint`] bitwise left shift operations.
 
-use crate::{ConstChoice, ConstCtOption, Limb, ShlVartime, Uint, Word, WrappingShl};
+use crate::{ConstChoice, ConstCtOption, Limb, ShlVartime, Uint, WrappingShl};
 use core::ops::{Shl, ShlAssign};
 use subtle::CtOption;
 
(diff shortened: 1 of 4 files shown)


src/uint/shl.rs
```diff
@@ -141,24 +162,41 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+            Limb::select(Limb::ZERO, carry, nz),
+        )
+    }
+
+    /// Computes `self << shift` where `0 <= shift < Limb::BITS`,
+    /// returning the result and the carry.
+    ///
+    /// Note: this operation should not be used in situations where `shift == 0`; it looks like
+    /// something in the execution pipeline can sometimes sniff this case out and optimize it away,
+    /// possibly leading to variable time behaviour.
```
tarcieri, line 179: Feels a bit like https://rustsec.org/advisories/RUSTSEC-2024-0344.html

Perhaps `NonZero<u32>` then?

andrewwhitehead: I've also added bit shift operations to `UintRef` (to be used by `BoxedUint`) and in that case `shl_limb` is implemented in terms of `conditional_shl_limb` which works around the issue a bit more cleanly, I'll look at switching this one over.

Also that wasn't my comment originally, it was copied from `shr_limb_nonzero`.

PR #923, opened 2025-08-12
Clean up gcd API

andrewwhitehead:
This mostly avoids exposing the separate `safegcd`/`bingcd` implementations. Both use the same pre-processing for inputs so the active variant only needs to be selected in `Odd<Uint>`. The `safegcd(_vartime)` and `bingcd(_vartime)` methods are only exposed for benchmarking at the moment.

Because a `const gcd` method on each integer type would have a higher priority than the trait method, it would produce type errors when trying to compute the GCD between different types (like `Uint` and `Int`). I've gone with the name `gcd_uint` for these methods.

### DIFF
--- a/src/uint/gcd.rs
+++ b/src/uint/gcd.rs
@@ -352,7 +352,7 @@ impl_gcd_uint_rhs!(
     NonZeroUint<LIMBS>,
     [NonZeroUint<LIMBS>, OddUint<LIMBS>]
 );
-impl_gcd_uint_lhs!(OddUint<LIMBS>, [Uint<LIMBS>, NonZeroUint<LIMBS>]);
+impl_gcd_uint_lhs!(OddUint<LIMBS>, [Uint<LIMBS>, NonZeroUint<LIMBS>, OddUint<LIMBS>]);
 
 impl<const LIMBS: usize> Xgcd for Uint<LIMBS> {
     type Output = UintXgcdOutput<LIMBS>;


tarcieri:
Curious what @erik-3milabs thinks about these changes

erik-3milabs:
> This mostly avoids exposing the separate safegcd/bingcd implementations. Both use the same pre-processing for inputs so the active variant only needs to be selected in Odd<Uint>.

This sounds like an excellent change: this removes so much ugly boiler plate!

>Because a const gcd method on each integer type would have a higher priority than the trait method, it would produce type errors when trying to compute the GCD between different types (like Uint and Int). I've gone with the name gcd_uint for these methods.

I would propose we use `gcd`/`gcd_(u)int` as follows:
```
Uint.gcd(Uint);
Uint.gcd_int(Int);
Int.gcd(Int);
Int.gcd_uint(Uint);
```
i.e., have `gcd` with a trailing `_(u)int` to indicate the expected right hand side type if this is not the same as the lhs type.

Edit: this will clash with the trait, though :thinking:

src/int/gcd.rs
```diff
@@ -248,121 +212,107 @@ impl<const LIMBS: usize, T: Copy> XgcdOutput<LIMBS, T> {
...
-    }
-}
-
-impl<const LIMBS: usize> Gcd<Uint<LIMBS>> for Int<LIMBS> {
-    type Output = Uint<LIMBS>;
-
-    fn gcd(&self, rhs: &Uint<LIMBS>) -> Self::Output {
-        self.abs().safegcd(rhs)
-    }
+macro_rules! impl_int_gcd_abs_lhs {
```
erik-3milabs, line 215: I find the name of this macro rather confusing. What does the `abs_lhs` mean here?

src/int/gcd.rs
```diff
@@ -248,121 +212,107 @@ impl<const LIMBS: usize, T: Copy> XgcdOutput<LIMBS, T> {
...
+    };
 }
 
-impl<const LIMBS: usize> Gcd<NonZeroUint<LIMBS>> for NonZeroInt<LIMBS> {
-    type Output = NonZeroUint<LIMBS>;
-
-    fn gcd(&self, rhs: &NonZeroUint<LIMBS>) -> Self::Output {
-        self.abs().safegcd(rhs)
-    }
+macro_rules! impl_int_gcd_abs_rhs {
```
erik-3milabs, line 238: ... and what does `abs_rhs` mean here?

src/int/gcd.rs
andrewwhitehead: Yeah, the naming is difficult. I made the inherent methods accept a Uint argument for consistency, so whenever the combination is two Int's it just needs to call the method with the `abs` of one side, chosen to avoid extra reductions to `Odd<Uint>`.

src/uint/gcd.rs
```diff
@@ -207,10 +197,18 @@ impl<const LIMBS: usize> OddUint<LIMBS> {
     /// Compute the greatest common divisor of `self` and `rhs`.
     ///
     /// Executes in variable time w.r.t. all input parameters.
+    #[doc(hidden)]
     #[inline]
     pub const fn safegcd_vartime(&self, rhs: &Uint<LIMBS>) -> Self {
         safegcd::gcd_odd::<LIMBS, true>(self, rhs)
     }
```
tarcieri, line 204: Why `#[doc(hidden)]` here? (and above)

tarcieri (approved):
Going to land this to get another release out

src/uint/gcd.rs
andrewwhitehead: Because they're only exposed for benchmarks at the moment, not meant to be used elsewhere. I suppose deprecated would also work but they just add noise to the documentation.

PR #927, opened 2025-08-15
Add Jacobi symbol calculation

andrewwhitehead:
This implementation is based on `bingcd`. It does not appear to affect the GCD performance, being optimized away when it is not needed. This would be useful in `primefield` for adding `is_square`/`is_square_vartime` methods which should be faster than `sqrt`.

### DIFF
--- a/src/int/mod_symbol.rs
+++ b/src/int/mod_symbol.rs
@@ -1,0 +1,101 @@
+//! Support for computing modular symbols.
+
+use crate::{ConstChoice, Int, Odd, Uint};
+
+impl<const LIMBS: usize> Int<LIMBS> {
+    /// Compute the Jacobi symbol `(self|rhs)`.
+    ///
+    /// For prime `rhs`, this corresponds to the Legendre symbol and
+    /// indicates whether `self` is quadratic residue modulo `rhs`.
+    pub const fn jacobi_symbol(&self, rhs: &Odd<Uint<LIMBS>>) -> i8 {
+        let (abs, sign) = self.abs_sign();
+        let jacobi = abs.jacobi_symbol(rhs) as i64;
+        // (-self|rhs) = -(self|rhs) iff rhs = 3 mod 4
+        let swap = sign.and(ConstChoice::from_word_eq(rhs.as_ref().limbs[0].0 & 3, 3));
+        swap.select_i64(jacobi, -jacobi) as i8
+    }
+
+    /// Compute the Jacobi symbol `(self|rhs)`.
+    ///
+    /// For prime `rhs`, this corresponds to the Legendre symbol and
+    /// indicates whether `self` is quadratic residue modulo `rhs`.
+    ///
+    /// This method executes in variable-time for the value of `self`.
+    pub const fn jacobi_symbol_vartime(&self, rhs: &Odd<Uint<LIMBS>>) -> i8 {
+        let (abs, sign) = self.abs_sign();
+        let jacobi = abs.jacobi_symbol(rhs);
+        if sign.is_true_vartime() && rhs.as_ref().limbs[0].0 & 3 == 3 {
+            -jacobi
+        } else {
+            jacobi
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{I256, U256};
+
+    #[test]
+    fn jacobi_quad_residue() {
+        // Two semiprimes with no common factors, and
+        // f is quadratic residue modulo g
+        let f = I256::from(59i32 * 67);
+        let g = U256::from(61u32 * 71).to_odd().unwrap();
+        let res = f.jacobi_symbol(&g);
+        let res_vartime = f.jacobi_symbol_vartime(&g);
+        assert_eq!(res, 1);
+        assert_eq!(res, res_vartime);
+    }
+
+    #[test]
+    fn jacobi_non_quad_residue() {
+        // f and g have no common factors, but
+        // f is not quadratic residue modulo g
+        let f = I256::from(59i32 * 67 + 2);
+        let g = U256::from(61u32 * 71).to_odd().unwrap();
+        let res = f.jacobi_symbol(&g);
+        let res_vartime = f.jacobi_symbol_vartime(&g);
+        assert_eq!(res, -1);
+        assert_eq!(res, res_vartime);
+    }
+
+    #[test]
+    fn jacobi_non_coprime() {
+        let f = I256::from(4391633i32);
+        let g = U256::from(2022161u32).to_odd().unwrap();
+        let res = f.jacobi_symbol(&g);
+        let res_vartime = f.jacobi_symbol_vartime(&g);
+        assert_eq!(res, 0);
+        assert_eq!(res, res_vartime);
+    }
+
+    #[test]
+    fn jacobi_zero() {
+        assert_eq!(I256::ZERO.jacobi_symbol(&U256::ONE.to_odd().unwrap()), 1);
+    }
+
+    #[test]
+    fn jacobi_neg_one() {
+        let f = I256::ONE;
+        assert_eq!(f.jacobi_symbol(&U256::ONE.to_odd().unwrap()), 1);
+        assert_eq!(f.jacobi_symbol(&U256::from(3u8).to_odd().unwrap()), 1);
+    }
+
+    #[test]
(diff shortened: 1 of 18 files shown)


tarcieri:
@erik-3milabs can you take a look at the `bingcd` changes?

src/modular/bingcd/gcd.rs
```diff
@@ -13,57 +51,57 @@ impl<const LIMBS: usize> Odd<Uint<LIMBS>> {
...
+    ///
+    /// Note: this algorithm is efficient for [Uint]s with relatively few `LIMBS`.
+    ///
+    /// Ref: Pornin, Optimized Binary GCD for Modular Inversion, Algorithm 1.
+    /// <https://eprint.iacr.org/2020/972.pdf>
+    ///
+    /// This method returns a pair consisting of the GCD and the sign of the Jacobi symbol,
+    /// 0 for positive and 1 for negative.
+    #[inline(always)]
+    pub const fn classic_bingcd_(&self, rhs: &Uint<LIMBS>) -> (Self, Word) {
```
fjarri: What's with the underscores in the end of method names?

src/int/mod_symbol.rs
```diff
@@ -0,0 +1,101 @@
+//! Support for computing modular symbols.
+
+use crate::{ConstChoice, Int, Odd, Uint};
+
+impl<const LIMBS: usize> Int<LIMBS> {
+    /// Compute the Jacobi symbol `(self|rhs)`.
+    ///
+    /// For prime `rhs`, this corresponds to the Legendre symbol and
+    /// indicates whether `self` is quadratic residue modulo `rhs`.
+    pub const fn jacobi_symbol(&self, rhs: &Odd<Uint<LIMBS>>) -> i8 {
```
fjarri: `rhs: Odd<Int>` seems more appropriate here

src/modular/bingcd/gcd.rs
```diff
@@ -13,57 +51,57 @@ impl<const LIMBS: usize> Odd<Uint<LIMBS>> {
     /// Ref: Pornin, Optimized Binary GCD for Modular Inversion, Algorithm 1.
     /// <https://eprint.iacr.org/2020/972.pdf>
     #[inline]
-    pub(crate) const fn classic_bingcd(&self, rhs: &Uint<LIMBS>) -> Self {
+    pub const fn classic_bingcd(&self, rhs: &Uint<LIMBS>) -> Self {
```
fjarri: Why are these and other similar methods made public? Because of the performance difference for different uint sizes?

src/modular/bingcd/gcd.rs
```diff
@@ -93,63 +132,76 @@ impl<const LIMBS: usize> Odd<Uint<LIMBS>> {
...
+            rhs,
+            U64::BITS - 1,
+        )
+        .0
     }
 
     /// Variable time equivalent of [`Self::optimized_bingcd_`].
-    #[inline]
-    const fn optimized_bingcd_vartime_<
+    #[inline(always)]
```
fjarri, line 197: Not sure if inlining such a big method is a good idea. Does that affect performance if `inline` is removed?

src/modular/const_monty_form/mod_symbol.rs
```diff
@@ -0,0 +1,61 @@
+//! Modular symbol calculation for integers in Montgomery form with a constant modulus.
+
+use super::{ConstMontyForm, ConstMontyParams};
+
+impl<MOD: ConstMontyParams<LIMBS>, const LIMBS: usize> ConstMontyForm<MOD, LIMBS> {
+    /// Compute the Jacobi symbol `(self|modulus)`.
+    ///
+    /// For a prime modulus, this corresponds to the Legendre symbol and indicates
+    /// whether `self` is quadratic residue.
+    pub const fn jacobi_symbol(&self) -> i8 {
```
fjarri: Perhaps it's worth it to introduce an enum for the return value, since there's only three possible values?

src/modular/bingcd/gcd.rs
```diff
@@ -93,63 +132,76 @@ impl<const LIMBS: usize> Odd<Uint<LIMBS>> {
...
-    const fn optimized_bingcd_<const K: u32, const LIMBS_K: usize, const LIMBS_2K: usize>(
+    ///
+    /// This method returns a pair consisting of the GCD and the sign of the Jacobi symbol,
+    /// 0 for positive and 1 for negative.
+    #[inline(always)]
+    pub const fn optimized_bingcd_<const K: u32, const LIMBS_K: usize, const LIMBS_2K: usize>(
         &self,
         rhs: &Uint<LIMBS>,
-    ) -> Self {
+        batch_max: u32,
```
fjarri, line 150: Since this method is public: how is the user supposed to choose `batch_max`? The docs don't really mention it.

src/modular/const_monty_form/mod_symbol.rs
tarcieri: [`Ordering`](https://doc.rust-lang.org/std/cmp/enum.Ordering.html) is good precedent for how to construct it

src/int/mod_symbol.rs
andrewwhitehead: Jacobi symbol is only defined for positive, odd lower arguments (unlike say the Kronecker symbol), although it supports negative upper arguments.

src/modular/bingcd/gcd.rs
andrewwhitehead: That was the convention in this file for 'private' methods. They should probably be free functions but that would complicate the changes a lot.

src/int/mod_symbol.rs
fjarri: Ah yes, you are right. I was thinking of its usage in `crypto-primes`, but I forgot that the negative argument there is indeed the upper one.

src/modular/bingcd/gcd.rs
andrewwhitehead: Originally yes, and to enable benchmarking of the different methods. I've adjusted them all to be `pub(crate)` now.

src/modular/bingcd/gcd.rs
andrewwhitehead: I'm being especially aggressive with this one because it's called in two places: one that uses the Jacobi symbol and one that doesn't, and I want the optimizer to always remove the symbol calculation when it's not needed (which it does so far).

src/modular/bingcd/gcd.rs
andrewwhitehead: Fixed: it's not meant to be public.

src/modular/const_monty_form/mod_symbol.rs
andrewwhitehead: That's doable, but it seemed a bit complicated if other modular symbols are added later, such as the Kronecker symbol. Then you'd have two types representing the same values. Also, this is compatible with the `num-modular` crate which is helpful for testing.

tarcieri: @andrewwhitehead if you follow the `Ordering` pattern, it's `repr(i8)`, so you can always cast it to `i8` anywhere, even a `const fn` context.

It also lets you add predicate methods, e.g. `pub const fn is_quadratic_residue(self) -> ConstChoice`. I've also added upstream impls of all the `subtle` traits to `Ordering` and you can use a similar pattern.

andrewwhitehead: That also raises the question of what to call the values, since 1 only necessarily indicates that the value is quadratic residue if the lower argument is prime. NonCoprime, MaybeQuadraticResidue, NotQuadraticResidue?

fjarri: In `crypto-primes` I just called them `One`, `Zero` and `MinusOne` :)

src/modular/monty_form/mod_symbol.rs
```diff
@@ -0,0 +1,63 @@
+//! Modular symbol calculation for integers in Montgomery form with a constant modulus.
```
tarcieri, line 1: Nit: isn't the `mod_` part implied by this being in `modular`?

andrewwhitehead: I suppose so, but it's consistent with `uint/mod_symbol` and `int/mod_symbol`

PR #930, opened 2025-08-22
Faster wrapping multiplication

andrewwhitehead:
This implements basic wrapping multiplication for `Uint` and `BoxedUint`, deferring to Karatsuba multiplication when it would be faster (estimated by benchmarks). The number of limbs needs to get fairly high before it needs to switch over though, and `Uint` currently only implements Karatsuba for a few limb counts that are powers of two (I also have a `const` implementation that might help with that).

A few relevant benchmarks:

```
wrapping ops/wrapping_mul, U256xU256
                        time:   [4.3439 ns 4.4469 ns 4.5640 ns]
                        change: [−52.321% −49.534% −46.632%] (p = 0.00 < 0.05)
                        Performance has improved.
wrapping ops/wrapping_mul, U4096xU4096
                        time:   [1.9225 µs 1.9347 µs 1.9493 µs]
                        change: [−14.332% −13.199% −11.674%] (p = 0.00 < 0.05)
                        Performance has improved.
wrapping ops/wrapping_mul, U8192xU4096
                        time:   [5.7493 µs 5.7893 µs 5.8438 µs]
                        change: [−31.980% −29.997% −28.668%] (p = 0.00 < 0.05)
                        Performance has improved.
wrapping ops/boxed_wrapping_mul
                        time:   [2.2299 µs 2.2406 µs 2.2524 µs]
                        change: [−27.376% −27.000% −26.631%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #932, opened 2025-08-25
Faster inversion modulo 2^k

andrewwhitehead:
`invert_mod2k` and `invert_mod2k_vartime` are optimized for `Uint` and `BoxedUint` (often by 95% or more), which brings the performance of `invert_mod` much closer to `invert_odd_mod`.

This depends on some of the updates in #930, and adds more functionality to `UintRef` for operating on limb slices in const methods. Support for operating on `Odd<UintRef>` or `NonZero<UintRef>` is also added.

### DIFF
--- a/src/uint/mul/schoolbook.rs
+++ b/src/uint/mul/schoolbook.rs
@@ -180,11 +180,12 @@ pub const fn square_wide(limbs: &[Limb], lo: &mut [Limb], hi: &mut [Limb]) {
     }
 }
 
-/// Schoolbook squaring which only calculates the lower limbs of the product.
+/// Schoolbook squaring which may calculate a limited number of limbs of the product.
 #[inline(always)]
+#[track_caller]
 pub const fn wrapping_square(limbs: &[Limb], out: &mut [Limb]) {
     assert!(
-        limbs.len() == out.len(),
+        limbs.len() * 2 >= out.len(),
         "schoolbook wrapping squaring length mismatch"
     );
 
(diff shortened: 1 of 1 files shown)


kathleenmotley49-maker (approved):

kathleenmotley49-maker (approved):

src/uint/ref_type.rs
```diff
@@ -92,12 +93,37 @@ impl UintRef {
         self.0.iter_mut()
     }
 
+    /// Access the number of limbs.
+    #[inline]
+    pub const fn len(&self) -> usize {
```
tarcieri: This is fairly consistently `fn nlimbs` elsewhere, e.g. `BoxedUint::nlimbs`, or the `nlimbs` macro for `Uint`

kathleenmotley49-maker (approved):

tarcieri:
@andrewwhitehead can you rebase?

src/non_zero.rs
```diff
@@ -51,20 +51,22 @@ impl<T> NonZero<T> {
...
-
     /// Returns the inner value.
     pub fn get(self) -> T {
         self.0
     }
 }
 
+impl<T: ?Sized> NonZero<T> {
+    /// Provides access to the contents of `NonZero` in a `const` context.
+    pub const fn as_ref(&self) -> &T {
```
tarcieri, line 62: I've been wondering if we should use a different method name for these which doesn't clash with the `AsRef` trait, but this is at least consistent with `Odd`

andrewwhitehead: Something like `inner()` could work.

tarcieri: `as_inner` perhaps to denote the borrow

kathleenmotley49-maker (approved):

PR #933, opened 2025-08-30
Faster retrieval from Montygomery form

andrewwhitehead:
This optimizes the `retrieve` methods of Montygomery forms by implementing a specialized Montgomery multiplication. This corresponds to the `almost_montgomery_mul_by_one` method (which is removed). In my tests performance is about 7% better for `BoxedMontyForm` at 4096 bits, and about 30-50% better for most `MontyForm` sizes.

### DIFF
--- a/src/modular/boxed_monty_form/mul.rs
+++ b/src/modular/boxed_monty_form/mul.rs
@@ -148,26 +148,6 @@ impl<'a> BoxedMontyMultiplier<'a> {
         debug_assert!(&*a < self.modulus);
     }
 
-    /// Perform a Montgomery multiplication, assigning a fully reduced result to `a`.
-    pub(super) fn mul_by_one(&mut self, a: &BoxedUint) -> BoxedUint {
-        debug_assert_eq!(a.bits_precision(), self.modulus.bits_precision());
-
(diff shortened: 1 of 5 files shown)


src/modular/boxed_monty_form/mul.rs
```diff
@@ -330,43 +310,6 @@ pub(crate) const fn almost_montgomery_mul(
...
-/// Same as `almost_montgomery_mul()` with `y == 1`.
-///
-/// Used for retrieving from Montgomery form.
-pub(crate) const fn almost_montgomery_mul_by_one(z: &mut [Limb], x: &[Limb], m: &[Limb], k: Limb) {
-    let n = z.len();
-
-    // This preconditions check allows compiler to remove bound checks later in the code.
-    if !(x.len() == n && m.len() == n) {
-        panic!("Failed preconditions in `almost_montgomery_mul_by_one`");
-    }
```
tarcieri, line 342: Is the compiler still able to elide these bounds checks? (or are they even still relevant)

Edit: aah, I guess that's the `assert!` on L82 of `reduction.rs` below

tarcieri (approved):

PR #920, opened 2025-08-12
Move bit operations to `UintRef`

andrewwhitehead:
This aligns `Uint` and `BoxedUint` bit operations by moving the logic to `UintRef` and removing the free functions.

PR #935, opened 2025-09-02
Revert "Move bit operations to `UintRef` (#920)"

tarcieri:
This reverts commit 4c891c200c309646b54f7e36572422d4dc91e129.

The `dsa` crate's tests in RustCrypto/signatures#1047 were failing and I bisected the failures to this commit.

I have confirmed the `dsa` crate tests pass with this reverted.

cc @andrewwhitehead


baloo:
It could very well be my fault for misusing the API back when I bumped dsa to use crypto-bigint.

tarcieri:
Anything that causes the tests to break in another project is suspect, and an indication we need more extensive testing in this crate

andrewwhitehead:
I believe I've narrowed down the issue to `shl1_assign` in `UintRef`: the last line should be `ConstChoice::from_word_lsb(carry.0)` instead of `ConstChoice::from_word_lsb(carry.0 >> Limb::HI_BIT)`. With this change the tests in `dsa` pass.

These particular tests do look a little fragile as they rely on `crypto-primes` coming up with the same random number(s) from the deterministic RNG, but the method for selecting prime candidates could easily change.

tarcieri:
@andrewwhitehead can you open a PR with the fix? (and ideally, a regression test) Thanks!

fjarri:
> These particular tests do look a little fragile as they rely on crypto-primes coming up with the same random number(s) from the deterministic RNG

There are some rudimentary vector tests as of https://github.com/entropyxyz/crypto-primes/pull/80, but more can be added. Certainly, it's a guarantee that we want to maintain.

PR #942, opened 2025-09-10
Add `Signed` and `Unsigned` marker traits

tarcieri:
These are notably bounded on `Integer` and thus usable in place of `Integer` anywhere a signed or unsigned integer specifically is expected

I note there's TODOs to impl `Integer` and `FixedInteger` for `Int`. I went ahead and added another TODO to impl `Signed`.

### DIFF
--- a/src/int.rs
+++ b/src/int.rs
@@ -240,6 +240,8 @@ impl<const LIMBS: usize> Default for Int<LIMBS> {
 
 // TODO: impl Integer
 
+// TODO: impl Signed
+
 impl<const LIMBS: usize> ConstZero for Int<LIMBS> {
     const ZERO: Self = Self::ZERO;
 }
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -198,10 +198,13 @@ pub trait Integer:
     }
 }
 
-/// Marker trait for signed integers.
-pub trait Signed: Integer {}
+/// Signed integers.
+pub trait Signed: Integer {
+    /// Corresponding unsigned integer type.
+    type Unsigned: Unsigned;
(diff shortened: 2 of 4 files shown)


tarcieri:
cc @erik-3milabs @andrewwhitehead  @dvdplm

andrewwhitehead:
I feel like `Signed` should implement `sign` and `abs` (and maybe `abs_sign`), and have an associated Unsigned type? Although that could come later.

tarcieri:
An associated type on `Signed` for its corresponding `Unsigned` type sounds good.

As far as factoring methods onto those traits, I think one thing that could really help inform that is actually getting `Integer` and `Signed` impl'd for `Int`. So I'd prefer to start with something as simple as possible and then work on getting actual trait impls in place to inform the trait design.

dvdplm:
> I think one thing that could really help inform that is actually getting `Integer` and `Signed` impl'd for `Int`

I think this is the way. With that in place, the path forward will be clear.

tarcieri:
Let me go ahead and merge this then, and followup with a PR that impls `Integer` and `Signed` for `Int` so we can discuss what needs to change/be factored where in that PR (and others can propose alternatives)

erik-3milabs:
> > I think one thing that could really help inform that is actually getting `Integer` and `Signed` impl'd for `Int`
> 
> I think this is the way. With that in place, the path forward will be clear.

This indeed sounds like the way to go. FYI: what stopped me from implementing `Integer` for `Int` last year, was that `Integer` requires `Encoding` which seemed to be rather tricky for signed integers.

PR #943, opened 2025-09-10
Impl `(Fixed)Integer` and `Signed` for `Int`

tarcieri:
Followup to #942

This factors the parts of `Integer` that `Int` currently doesn't implement onto `Unsigned`, along with things that definitely belong on `Unsigned`.

In particular, there's a TODO for `WrappingMul`, where it appears `Int` lacks an implementation.

The two types diverge on `Div`-by-`NonZero`, where `Int` currently returns a `CtOption`. From what I can tell this is an artifact of the implementation (the potential failing case is the quotient when computed as a `Uint` exceeding the size of an `Int` from what I can tell). As it were, the `DivAssign` impl uses `expect` in such a case. Ideally, the two could be unified and have common, `CtOption`-free bounds on `Integer` instead.

There are several places `From` impls and methods of `Integer` expect it's infallibly possible to create an `Integer` from a `Limb`, or a `u32`, or a `u64`. But this isn't possible with `Int<1>` on 32-bit platforms, where a `u32` or anything bigger (which includes `Limb`) could overflow the maximum positive value.

@fjarri @dvdplm heads up this involves some breaking changes to the traits, namely you'll need to use `Unsigned` instead of `Integer` for accessing `<T as Unsigned>::Monty`, and also bound on `Unsigned` instead of `Integer` in such places, and also for a handful of other supertraits and methods you're probably using. I'll leave some line notes.

Hopefully we can coordinate on both the design in a way that works for `crypto-primes` and also ensure we can do some coordinated releases to crates.io that include the updates for the breaking changes.

cc @andrewwhitehead @erik-3milabs


src/traits.rs
```diff
@@ -198,14 +170,49 @@ pub trait Integer:
...
-pub trait Unsigned: Integer {}
+/// Unsigned [`Integer`]s.
+pub trait Unsigned:
+    AddMod<Output = Self>
+    + BitOps
+    + Div<NonZero<Self>, Output = Self>
+    + for<'a> Div<&'a NonZero<Self>, Output = Self>
+    + DivRemLimb
+    + From<u8>
+    + From<u16>
```
tarcieri, line 210: These two could actually go on `Integer`, but for the time being I just moved them all

src/traits.rs
```diff
@@ -198,14 +170,49 @@ pub trait Integer:
     }
 }
 
-/// Signed integers.
-pub trait Signed: Integer {
+/// Signed [`Integer`]s.
+pub trait Signed:
```
tarcieri, line 175: Are there additional basic trivial things we can add here?

- [x] `From<i8> + From<i16> + From<i32>` (I now notice `Unsigned` has `From<u64>` though that might panic on 32-bit for `Uint<1>`?)
- [x] `Signed::abs` and `Signed::abs_sign`
- [x] `Signed::is_negative` and `Signed::is_positive`

src/traits.rs
```diff
@@ -198,14 +170,49 @@ pub trait Integer:
...
+    /// optimized for the performance of modular operations at the price of a conversion overhead.
+    type Monty: Monty<Integer = Self>;
+
+    /// The value `1` with the same precision as `other`.
+    fn one_like(other: &Self) -> Self {
+        Self::from_limb_like(Limb::ONE, other)
+    }
+
+    /// Returns an integer with the first limb set to `limb`, and the same precision as `other`.
+    fn from_limb_like(limb: Limb, other: &Self) -> Self;
```
tarcieri: I had to move this because the default impl of `one_like` calls `from_limb_like`, and it's not possible to  impl `from_limb_like` on a `Signed` type.

Alternatively `one_like` can stay on `Integer` and we can get rid of the default impl.

src/traits.rs
```diff
@@ -198,14 +170,49 @@ pub trait Integer:
     }
 }
 
-/// Signed integers.
-pub trait Signed: Integer {
+/// Signed [`Integer`]s.
+pub trait Signed:
+    Div<NonZero<Self>, Output = CtOption<Self>>
+    + for<'a> Div<&'a NonZero<Self>, Output = CtOption<Self>>
```
tarcieri, line 177: @erik-3milabs am I right that `CtOption` could potentially be removed here, or is there a failing case that's a hard requirement?

It seems like it's there to handle the case that `Uint` overflows `Int`?

I noticed `DivAssign` does `expect`.

src/traits.rs
```diff
@@ -198,14 +170,49 @@ pub trait Integer:
...
+    + From<u8>
+    + From<u16>
+    + From<u32>
+    + From<u64>
+    + From<Limb>
+    + Integer
+    + MulMod<Output = Self>
+    + NegMod<Output = Self>
+    + RemLimb
+    + SquareRoot
```
tarcieri, line 218: Are there any cryptosystems that use 𝑖? 😅

src/traits.rs
```diff
@@ -198,14 +170,49 @@ pub trait Integer:
...
+    + for<'a> Div<&'a NonZero<Self>, Output = Self>
+    + DivRemLimb
+    + From<u8>
+    + From<u16>
+    + From<u32>
+    + From<u64>
+    + From<Limb>
+    + Integer
+    + MulMod<Output = Self>
+    + NegMod<Output = Self>
```
tarcieri, line 216: I moved all of the `*Mod` traits here for now since `Int` doesn't impl them

src/traits.rs
```diff
@@ -198,14 +170,49 @@ pub trait Integer:
...
 
-/// Unsigned integers.
-pub trait Unsigned: Integer {}
+/// Unsigned [`Integer`]s.
+pub trait Unsigned:
+    AddMod<Output = Self>
+    + BitOps
+    + Div<NonZero<Self>, Output = Self>
+    + for<'a> Div<&'a NonZero<Self>, Output = Self>
+    + DivRemLimb
```
tarcieri, line 208: `Int` doesn't impl `(Div)RemLimb`

src/traits.rs
```diff
@@ -198,14 +170,49 @@ pub trait Integer:
...
+    + From<u64>
+    + From<Limb>
+    + Integer
+    + MulMod<Output = Self>
+    + NegMod<Output = Self>
+    + RemLimb
+    + SquareRoot
+    + SquareMod<Output = Self>
+    + SubMod<Output = Self>
+    + WrappingMul // here instead of `Integer` because `Int` lacks a `wrapping_mul` impl
```
tarcieri: Perhaps we can get this one addressed prior to a release /cc @erik-3milabs

src/traits.rs
```diff
@@ -198,16 +172,65 @@ pub trait Integer:
...
+    /// The magnitude of this [`Signed`].
+    fn abs(&self) -> Self::Unsigned {
+        self.abs_sign().0
+    }
+
+    /// Whether this [`Signed`] is negative, as a [`ConstChoice`].
+    fn is_negative(&self) -> ConstChoice;
+
+    /// Whether this [`Signed`] is positive, as a [`ConstChoice`].
+    fn is_positive(&self) -> ConstChoice;
```
tarcieri: Used `ConstChoice` for all of these though that may not be the best option. I know elsewhere we have some hacks like using `ConstChoice` for one which is provided as a `const fn`, and another which is a trait method which returns `Choice`

(would that we didn't have to deal with such a split, sigh)

src/traits.rs
erik-3milabs: Dividing `Int::MIN` by `Int::MINUS_ONE` overflows; this `CtOption` should be `None` only in this specific case.

(There are other cases when the numerator and denominator have different LIMBS, as pointed out in issue #847)

src/traits.rs
erik-3milabs: I didn't get around to implementing it, since I couldn't think of a proper way to interpret what a `wrapping_mul` operation would look like for signed integers.

src/traits.rs
erik-3milabs: (please be very clear what the definitions of `is_negative` and `is_positive` will be, since, from a mathematical standpoint, it is unclear whether `0` belongs to either category. I.e., nonnegative and nonpositive).

src/traits.rs
tarcieri: Aah, and I see Rust core handles this case with a panic for the built-in integer types. Unless we do that (not seriously suggesting we do) I guess it's unavoidable.

src/traits.rs
tarcieri: Shouldn't it just be casting the bit pattern to unsigned, performing an unsigned wrapping mul, and casting the resulting bit pattern back?

erik-3milabs: > Shouldn't it just be casting the bit pattern to unsigned, performing an unsigned wrapping mul, and casting the resulting bit pattern back?

If I understand you correctly, you're proposing
```rust
def wrapping_mul(self, rhs: &Self) -> Self {
    self.as_uint().wrapping_mul(rhs.as_uint()).as_int()
}
```

That could work.


I always felt that the at least the sign of the wrapped result should be the same as the sign of the `widened` result (which is not the case here) but I cannot argue why that should be the case.

tarcieri: I think the semantics should follow from the wrapping behavior of core integers, where e.g. `i8::MAX.wrapping_mul(2)` is `-2` (which I think holds for all of the `i*` types, perhaps useful for a test?)

erik-3milabs: Actually, this only works when `rhs: &Self`; when it has a different bit length, you run into trouble.

E.g.: `-13i8 * -4i4`:
- it would be cast to `243u8 * 12u4` 
- which evaluates to `100u8`
- which would be cast to `100i8`.

However, the actual result should be `56i8`.

We can achieve this by instead implementing

```rust
pub const fn wrapping_mul<const RHS_LIMBS: usize>(&self, rhs: &Int<RHS_LIMBS>) -> Self {
    let (abs_lhs, lhs_sgn) = self.abs_sign();
    let (abs_rhs, rhs_sgn) = rhs.abs_sign();
    let (lo, _) = abs_lhs.widening_mul(&abs_rhs);
    *lo.wrapping_neg_if(lhs_sgn.xor(rhs_sgn)).as_int()
}
```

tarcieri: @erik-3milabs the way that bound is notated, i.e. as `WrappingMul`, it requires `Rhs = Self` (implicitly, as that's the default), though I wouldn't be opposed to adding a more flexible `const fn` version

src/traits.rs
tarcieri: In [59a869f](https://github.com/RustCrypto/crypto-bigint/pull/943/commits/59a869f0fc1c02a4abe371f639bcc9b341199c31) I updated the docs to note a non-zero requirement on both.

I also changed them all to use `Choice` for consistency with the other APIs on `Integer` and `Unsigned`.

src/traits.rs
tarcieri: @erik-3milabs added your suggested `wrapping_mul` impl in [a4bda86](https://github.com/RustCrypto/crypto-bigint/pull/943/commits/a4bda86d165ce1c6249fe30ff3b061f15b6b8b46) and updated the bounds so `Integer` now bounds on `WrappingMul`, thanks!

src/traits.rs
tarcieri: This makes me wonder if we should define a `One` trait (and `ConstOne`) similar to our `Zero` trait and `num_traits::One`/`ConstOne`

src/traits.rs
tarcieri: There aren't any `From` impls for unsigned types on `Int` at all now, so let's go with this for now and potentially revisit these bounds when/if such impls are added

erik-3milabs:
@tarcieri perhaps #946 can be merged into this branch instead?

src/traits.rs
tarcieri: I think we're good for now but feel free to leave a comment still

src/traits.rs
tarcieri: Moved `one_like` back to `Integer`. We can potentially follow up with a `One` trait.

tarcieri:
@erik-3milabs heh, too late! 😅 

Rebased with #946 merged though. This is "done" from my perspective.

tarcieri:
I opened https://github.com/entropyxyz/crypto-primes/pull/91 to update `crypto-primes`, so this is just waiting on review from @fjarri @dvdplm and we can move forward with it and another set of crate releases to make the breaking change

tarcieri:
In the meantime, I might experiment with extracting a `One` trait

tarcieri:
Rebased on #947, calling this done

tarcieri:
@fjarri if this one looks good to you I can merge it and get a `crypto-bigint` v0.7.0-rc.5 release out that includes it, but if I do it'd be great if you can merge https://github.com/entropyxyz/crypto-primes/pull/91 shortly thereafter and get another `crypto-primes` prerelease out as it is very much a breaking change

fjarri:
Sure, I have free time. I'll be on look out.

tarcieri:
Alright, let's do it!

tarcieri:
@fjarri can you cut a new `crypto-primes` crate release to crates.io which is compatible with these changes? thanks!

fjarri:
Done

PR #946, opened 2025-09-11
Implement `Int::wrapping_mul`

erik-3milabs:
As requested [here](https://github.com/RustCrypto/crypto-bigint/pull/943#discussion_r2337736569).


tarcieri:
Looks good except for rustfmt

erik-3milabs:
@tarcieri fixed `rustfmt`

PR #947, opened 2025-09-11
Add `One` trait (like `Zero`)

tarcieri:
Adds a trait for the multiplicative identity element of `Self`, similar to `num_traits::One` but with support for constant-time comparisons and predicate methods like `is_one`.

It follows the same basic shape as the existing `Zero` trait.

The `Integer` trait is now bounded on it, and this removes the previous `one` and `one_like` methods from `Integer`, which are now available via the new trait.

`One` is impl'd for all the same types as `Zero`, and notably it can also be impl'd for `NonZero<T>` and `Odd<T>` where `Zero` definitionally cannot.

cc @andrewwhitehead @dvdplm @erik-3milabs

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -265,20 +265,6 @@ pub trait One: ConstantTimeEq + Sized {
     }
 }
 
-impl<T: ConstZero + ConstantTimeEq> Zero for T {
-    #[inline(always)]
-    fn zero() -> T {
-        Self::ZERO
-    }
-}
-
-impl<T: ConstOne + ConstantTimeEq> One for T {
-    #[inline(always)]
-    fn one() -> T {
-        Self::ONE
-    }
-}
-
 /// Trait for associating constant values with a type.
 pub trait Constants: ConstZero + ConstOne {
     /// Maximum value this integer can express.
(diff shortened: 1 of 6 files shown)


src/traits.rs
```diff
@@ -269,22 +233,88 @@ pub trait Zero: ConstantTimeEq + Sized {
...
-/// Trait for associating constant values with a type.
-pub trait Constants: ConstZero {
-    /// The value `1`.
-    const ONE: Self;
+impl<T: ConstOne + ConstantTimeEq> One for T {
+    #[inline(always)]
+    fn one() -> T {
+        Self::ONE
+    }
+}
```
tarcieri: I'm not sure these blanket impls are actually a good idea, though I did this for consistency with the blanket impl of `Zero` for `ConstZero`.

Notably this precludes adding `ConstOne` impls on `NonZero<T>` and `Odd<T>` where `T: ConstOne`, because the blanket impl would conflict with the explicit impl of `One`, which we still need for e.g. `BoxedUint`

tarcieri: Removed the blanket impls and it's now possible to impl `ConstOne` for `NonZero` and `Odd`! 🎉 

TODO: `Wrapping`, but that's a bit less trivial.

tarcieri:
I'm decently pleased with how this turned out so I'll merge it and get #943 rebased on top of it

PR #950, opened 2025-09-12
Extract `jacobi` module; add `Neg` impl

tarcieri:
Factors `JacobiSymbol` out of the `traits` module and into its own `jacobi` module.

Also adds a `Neg` impl, adapted from the one in `crypto-primes`, which seems to be the one thing missing for `crypto-primes` to switch to it for its Jacobi symbol type.

cc @andrewwhitehead @fjarri

### DIFF
--- a/src/jacobi.rs
+++ b/src/jacobi.rs
@@ -1,0 +1,87 @@
+use crate::ConstChoice;
+use core::ops::Neg;
+use subtle::{Choice, ConstantTimeEq};
+
+/// Possible return values for Jacobi symbol calculations.
+#[derive(Debug, Copy, Clone)]
+#[repr(i8)]
+pub enum JacobiSymbol {
+    /// The two arguments are not coprime, they have a common divisor apart from 1.
+    Zero = 0,
+
+    /// The two arguments are coprime. If the lower argument is prime, then the upper argument
+    /// is quadratic residue modulo the lower argument. Otherwise, the upper argument is known to
+    /// be quadratic nonresidue for an even number of prime factors of the lower argument.
+    One = 1,
+
+    /// The two terms are coprime, and the upper argument is a quadratic nonresidue modulo the
+    /// lower argument.
+    MinusOne = -1,
+}
+
+impl JacobiSymbol {
+    /// Determine if the symbol is zero.
+    pub const fn is_zero(&self) -> ConstChoice {
+        ConstChoice::from_i64_eq(*self as i8 as i64, 0)
+    }
+
+    /// Determine if the symbol is one.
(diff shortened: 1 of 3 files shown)


tarcieri:
PR to switch `crypto-primes` to using the upstream `JacobiSymbol` enum: https://github.com/entropyxyz/crypto-primes/pull/92

tarcieri:
Sidebar: it looks like some of the functionality from `crypto-primes` could potentially be upstreamed as well

fjarri:
What functionality specifically?

tarcieri:
@fjarri that was mostly to say I was wondering if it might make sense to upstream anything else from https://github.com/entropyxyz/crypto-primes/blob/9e69b9e/src/hazmat/jacobi.rs besides the `Neg` impl

fjarri:
It is sort of a specialized vartime Jacobi symbol implementation for small `a` values. I need to check how fast the `crypto-bigint` variant is, maybe it can be removed completely.

tarcieri:
@fjarri never mind then. Does this otherwise look good to you? (that is to say, from the perspective of adopting it in `crypto-primes`)

src/jacobi.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Negate the symbol.
+    pub const fn neg(self) -> Self {
+        match self {
+            Self::Zero => Self::Zero,
+            Self::One => Self::MinusOne,
+            Self::MinusOne => Self::One,
+        }
+    }
+
+    pub(crate) const fn from_i8(value: i8) -> Self {
```
fjarri, line 47: Not sure what the common convention is, but shouldn't this return a `Result`?

fjarri: Or I guess it would be `try_from_i8()` then

tarcieri: Sidebar: I think not returning a `Result` is fine since it's an internal function which is used as the final step of selecting the Jacobi symbol in a constant time-ish manner, though I'm curious if any of the techniques used here might apply for ensuring a branchless conversion to a particular enum variant: https://sailor.li/ints-to-enums

fjarri:
Btw, just realized that in order to use the upstream Jacobi symbol in `crypto-primes`, I need it to be a method in `Unsigned` (or `Integer`)

tarcieri:
@fjarri in this PR I noticed it was defined in the `traits` module but there wasn't actually a trait for it.

@andrewwhitehead I assume the idea was to add a trait for it?

If we do that it seems like it can be impl'd for `Int` and `Uint` and added to `Integer` (and also impl'd for `*MontyForm` and added to the `Monty` bounds, perhaps?)

PR #952, opened 2025-09-13
Revert "Faster wrapping multiplication"

tarcieri:
This reverts commit 9fd27fa004f25e0ac914887424a42ef760b6e06d (#930)

This is causing panics and other failures in the RSA crate:

    thread 'key::tests::key_generation_multi_16_1024' panicked at
    crypto-bigint-0.7.0-rc.5/src/uint/mul/schoolbook.rs:94:51:
    index out of bounds: the len is 1 but the index is 1

cc @andrewwhitehead


andrewwhitehead:
Hmm, sorry about that. More test coverage needed.

PR #954, opened 2025-09-13
Re-apply "Faster wrapping multiplication"

andrewwhitehead:
This reapplies the changes in #930 with additional tests for `wrapping_mul` and `wrapping_square`, and a small fix for an error occurring in the `rsa` crate.

### DIFF
--- a/src/uint/mul.rs
+++ b/src/uint/mul.rs
@@ -119,7 +119,9 @@ impl<const LIMBS: usize> Uint<LIMBS> {
 
     /// Perform wrapping square, discarding overflow.
     pub const fn wrapping_square(&self) -> Uint<LIMBS> {
-        self.square_wide().0
+        let mut lo = Uint::ZERO;
+        schoolbook::wrapping_square(&self.limbs, &mut lo.limbs);
+        lo
     }
 
     /// Perform saturating squaring, returning `MAX` on overflow.
--- a/src/uint/mul.rs
+++ b/src/uint/mul.rs
@@ -12,124 +12,7 @@ use crate::{
 use self::karatsuba::UintKaratsubaMul;
 
 pub(crate) mod karatsuba;
-
-/// Schoolbook multiplication a.k.a. long multiplication, i.e. the traditional method taught in
-/// schools.
-///
-/// The most efficient method for small numbers.
-#[inline(always)]
-const fn schoolbook_multiplication(lhs: &[Limb], rhs: &[Limb], lo: &mut [Limb], hi: &mut [Limb]) {
-    if lhs.len() != lo.len() || rhs.len() != hi.len() {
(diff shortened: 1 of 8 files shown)


src/uint/mul.rs
```diff
@@ -184,8 +67,15 @@ impl<const LIMBS: usize> Uint<LIMBS> {
     }
 
     /// Perform wrapping multiplication, discarding overflow.
-    pub const fn wrapping_mul<const H: usize>(&self, rhs: &Uint<H>) -> Self {
-        self.widening_mul(rhs).0
+    pub const fn wrapping_mul<const RHS_LIMBS: usize>(&self, rhs: &Uint<RHS_LIMBS>) -> Self {
+        // A single special case that is faster for now
+        if LIMBS == RHS_LIMBS && LIMBS == 128 {
+            return self.widening_mul(rhs).0;
+        }
```
tarcieri, line 74: Seems strange?

src/uint/mul/schoolbook.rs
```diff
@@ -87,7 +87,7 @@ pub const fn wrapping_mul(lhs: &[Limb], rhs: &[Limb], out: &mut [Limb]) {
 
         while k < lhs.len() {
             let j = k - i;
-            if k == rhs.len() {
+            if j == rhs.len() {
```
tarcieri, line 90: Was this the only issue? I saw some other test failures besides the out-of-bounds panic

andrewwhitehead: Yep, seems to fix all the tests. Besides the panic it could cause incorrect results, some of the new tests capture that:

```
thread 'uint::mul::tests::wrapping_mul_mixed' panicked at src/uint/mul.rs:357:9:
assertion `left == right` failed
  left: Uint(0x0000000000000000751EA4E5BF0EB289)
 right: Uint(0x7527C959D947A4F1751EA4E5BF0EB289)
```

src/uint/mul.rs
andrewwhitehead: This would change with improvements to the Karatsuba multiplication, but right now it's only for specific Uint sizes.

baloo (approved):
Ran CI on RSA with this change: https://github.com/RustCrypto/RSA/pull/581

tarcieri:
I generally wish we had better testing coverage testing which is something I intend to pursue after a release. I feel like we've had a few deficiencies which weren't caught by manual code review or tests. It's generally a bit scary to have defects that aren't caught until we actually integration test against a full cryptographic algorithm implementation. And the real worries are in all those untested branches.

I keep poking at things like [hax](https://hax.cryspen.com/) which would let us possibly have small snippets of code which are proven to be equivalent to some formally verified reference implementation (for all inputs, using e.g. symbolic execution).

Issue #810, opened 2025-04-26
`BoxedUint::checked_mul()` inefficiency


fjarri:
`BoxedUint::checked_mul()` and `Mul` impls allocate twice (they do a widening mul then shorten and check for overflow), which also means it performs extra work calculating high limbs that will be discarded anyway. Can this be avoided?

andrewwhitehead:
Is there a use case to test this with?

fjarri:
Hm, what do you mean by a use case? It happens in every multiplication.

andrewwhitehead:
I mean a crate that is using `checked_mul` or `saturated_mul`, I think they're only used in tests here.

fjarri:
In my case, it's used in https://github.com/entropyxyz/synedrion . The cryptographic scheme there involves a lot of operations with integers that are supposed to be bounded and never overflow (specifically, the part that deals with Paillier encryption and related ZK proofs). 

Granted, I am currently using `Uint`, not `BoxedUint`, but I think it is very likely that someone else might use heap integers for that, or for something similar.

PR #962, opened 2025-09-24
Update Karatsuba multiplication

andrewwhitehead:
This switches Karatsuba multiplication to a more unified approach across Uint and BoxedUint:

- Wrapping Karatsuba multiplication and squaring is implemented, for any length of output (up to the whole product)
- BoxedUint does not make larger allocations than necessary (no external scratch space), and essentially all BoxedUint multiplications are improved
- Most Uint multiplications over 16 limbs are improved, especially for non power-of-2 sizes and mixed-size multiplications: U8192xU4096 is 50% faster for a wide product and 47% for wrapping for example
- Macros are removed and all methods are made `const`
- Performance is slower for integers with thousands of limbs, but another approach like FFT multiplication should really be used there

### DIFF
--- a/src/uint/mul/karatsuba.rs
+++ b/src/uint/mul/karatsuba.rs
@@ -22,7 +22,7 @@
 //!
 //!   x^2 = (x0 + x1•b)^2 = x0^2 + 2x0x1•b + (x1•b)^2
 //!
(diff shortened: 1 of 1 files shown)


andrewwhitehead:
I think maybe tracking the carry from the wrapping multiplication can help optimize checked multiplication (#810)?

PR #965, opened 2025-09-26
Unify division methods

andrewwhitehead:
This moves the core division code to `UintRef` and reuses it for `BoxedUint` and `Uint`. There should not be a significant impact on performance.

Along the way `BoxedUint::sqrt` is optimized to reduce allocations.


src/uint/div.rs
```diff
@@ -1,548 +1,169 @@
...
     pub const fn rem_limb(&self, rhs: NonZero<Limb>) -> Limb {
-        rem_limb_with_reciprocal(self, &Reciprocal::new(rhs))
+        self.as_uint_ref().rem_limb(rhs)
     }
 
-    /// Computes `self` / `rhs`, returns the quotient (q) and the remainder (r)
+    /// Computes `self` / `rhs`, returning the quotient and the remainder.
     ///
     /// This function is constant-time with respect to both `self` and `rhs`.
     pub const fn div_rem<const RHS_LIMBS: usize>(
```
andrewwhitehead, line 41: This could simply delegate to the `div_rem` method of `UintRef`, but then performance suffers. There's little to no difference when the size-aware bitshift methods of `Uint` are used instead.

tarcieri: Curious why. Did you try adding `#[inline]`?

src/uint/ref_type/bits.rs
```diff
@@ -3,6 +3,11 @@ use crate::{ConstChoice, Limb, traits::BitOps};
 use subtle::Choice;
 
 impl UintRef {
+    /// Get the precision of this [`BoxedUint`] in bits.
```
tarcieri: Copypasta? (references `BoxedUint`, surprised it didn't trip the doctest CI?)

tarcieri (approved):
One nit, but generally like the direction of this PR

src/uint/div.rs
andrewwhitehead: Yes, I'm not sure why I couldn't get matching performance but I've seen this a few times now with the bit shifting methods. It seems to be a step too far for `Uint::<LIMBS>::as_uint_ref().nlimbs()` to be interpreted as a constant, whereas `LIMBS` is not a problem.

tarcieri: Maybe try `#[inline(always)]`?

andrewwhitehead: It looks like the performance matches if `bounded_wrapping_shl_assign` is also inlined. I'll do a bit of testing to see if there's any negative impact from that.

andrewwhitehead: This is updated now. I added `rem_wide` support, which is actually more flexible than the old version as it does not need the low limbs to be the same width as the high limbs. This would allow `mul_mod(_vartime)` to support multiplication by a different number of limbs, although I haven't enabled that yet. Performance should be a little better in essentially all cases.

PR #983, opened 2025-10-14
Optimize checked multiplication and squaring

### LINKED ISSUE #810: `BoxedUint::checked_mul()` inefficiency

andrewwhitehead:
This optimizes checked (and saturated) multiplication and squaring to perform a wrapping multiplication with a carry, and check for overflow by a more efficient method. For boxed integers this means the allocation is no larger than needed. Performance is 30+% improved at the 4096 bit size and more at larger sizes.

A few missing methods are added for consistency: inherent `checked_mul` methods, as well as `saturating_mul` and `saturating_square` for `BoxedUint`.

`Int` could also use some improvements here but I think that's a separate PR.

Resolves #810

### DIFF
--- a/src/uint/mul.rs
+++ b/src/uint/mul.rs
@@ -252,9 +252,7 @@ pub(crate) const fn wrapping_mul_overflow(
 
 #[cfg(test)]
 mod tests {
-    use crate::{ConstChoice, U64, U128, U192, U256, Zero};
-    #[cfg(feature = "rand_core")]
-    use rand_core::{RngCore, SeedableRng};
+    use crate::{ConstChoice, U64, U128, U192, U256, Uint, Zero};
 
     #[test]
     fn widening_mul_zero_and_one() {
@@ -399,6 +397,7 @@ mod tests {
     #[test]
     fn mul_cmp() {
         use crate::{Random, U4096};
+        use rand_core::SeedableRng;
         let mut rng = rand_chacha::ChaCha8Rng::seed_from_u64(1);
 
         for _ in 0..50 {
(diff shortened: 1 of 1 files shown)


src/uint/div.rs
```diff
@@ -715,9 +710,10 @@ mod tests {
                 NonZero::new(U256::random(&mut rng).overflowing_shr_vartime(128).unwrap()).unwrap();
             let n = num.checked_mul(den.as_ref());
             if n.is_some().into() {
-                let (q, _) = n.unwrap().div_rem(&den);
+                let n = n.unwrap();
```
tarcieri, line 713: Minor nit, but seems like you could get rid of this `unwrap` with:

```rust
if let Some(n) = n.into_option() {
```

src/uint/mul.rs
```diff
@@ -46,15 +46,24 @@ impl<const LIMBS: usize> Uint<LIMBS> {
     }
 
     /// Perform wrapping multiplication, discarding overflow.
-    // #[inline(always)]
```
tarcieri, line 49: Hah, guess that got missed?

src/uint/mul.rs
```diff
@@ -46,15 +46,24 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
 
     /// Perform saturating multiplication, returning `MAX` on overflow.
     pub const fn saturating_mul<const RHS_LIMBS: usize>(&self, rhs: &Uint<RHS_LIMBS>) -> Self {
-        let (res, overflow) = self.widening_mul(rhs);
-        Self::select(&res, &Self::MAX, overflow.is_nonzero())
+        self.checked_mul(rhs).unwrap_or(Uint::MAX)
+    }
+
+    /// Perform wrapping multiplication, checking that the result fits in the original [`Uint`] size.
+    pub const fn checked_mul<const RHS_LIMBS: usize>(
```
andrewwhitehead, line 59: FYI @tarcieri This method can clobber the trait method, returning a `ConstCtOption` instead of a `CtOption`. It's consistent with the `checked_square` method though, and provides `const` support. I don't think the method is widely used but in some cases this would produce an unneeded import warning for `CheckedMul`, in others it could produce an error if a `CtOption` was expected.

tarcieri: Yeah the discrepancy between `ConstCtOption` and `CtOption` continues to be frustrating. I think there are a few other places that happens.

We could potentially use different method names to avoid the ambiguity.

PR #988, opened 2025-10-16
Make AMM-based modular `pow` function generic

tarcieri:
The AMM modpow implementation was previously specialized to `BoxedMontyForm`. This extracts the implementation into the toplevel `pow` module and makes it generic.

Note it doesn't yet add support for using it with `MontyForm`, but that should be a straightforward next step after this.

Making this work involved defining a `MontyMultiplier` extension trait which I've called `AmmMultiplier`, however it's marked `pub(crate)` and is otherwise entirely internal.

Constructing `BoxedMontyForm` from an AMM now has an internal helper function called `BoxedMontyForm::from_amm`, both to permit reuse and because the methods needed to implement it with optimum performance aren't defined on the `Integer`/`Unsigned` traits.

This has no impact on benchmarks:

```
Boxed Montgomery arithmetic/modpow, BoxedUint^BoxedUint
                        time:   [30.203 ms 30.263 ms 30.325 ms]
                        change: [−0.1536% +0.2861% +0.6654%] (p = 0.17 > 0.05)
```

cc @andrewwhitehead

### DIFF
--- a/src/modular/boxed_monty_form/mul.rs
+++ b/src/modular/boxed_monty_form/mul.rs
@@ -7,10 +7,11 @@
 
 use super::{BoxedMontyForm, BoxedMontyParams};
 use crate::{
-    BoxedUint, ConstChoice, Limb, MontyMultiplier, Square, SquareAssign,
+    AmmMultiplier, BoxedUint, ConstChoice, Limb, MontyMultiplier, Square, SquareAssign,
     modular::mul::montgomery_multiply_inner,
 };
 use core::ops::{Mul, MulAssign};
+use subtle::ConstantTimeLess;
 
 #[cfg(feature = "zeroize")]
 use zeroize::Zeroize;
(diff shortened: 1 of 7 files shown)


src/modular/boxed_monty_form/mul.rs
```diff
@@ -48,6 +49,32 @@ impl BoxedMontyForm {
...
     }
+
+    /// Instantiate [`BoxedMontyForm`] from the result of an "Almost Montgomery Multiplication".
+    pub(crate) fn from_amm(mut z: BoxedUint, params: BoxedMontyParams) -> Self {
+        // Ensure the output is properly reduced.
+        //
+        // Using the properties of `almost_montgomery_mul()` (see its documentation):
+        // - We have an incoming `x` which is fully reduced (`floor(x / modulus) = 0`).
+        // - We build an array of `powers` which are produced by multiplying the previous power by
+        //   `x`, so for each power `floor(power / modulus) <= 1`.
```
tarcieri, line 60: This is perhaps a bit specific to its usage in `pow`

Issue #991, opened 2025-10-22
Exposing internal functions


ycscaly:
Some algorithms we write use internal functionality that is currently not exposed, across multiple modules (e.g. `ConstChoice`, `Uint` etc). 
These are usually `pub(crate)` that we need as `pub`. 

Currently, we have to maintain a fork for several reasons, and this is one of them.

I was wondering how to best tackle this; should we create interfaces, or (my preference as its the easiest way) simply make them `pub`?

fjarri:
`Uint` is public, isn't it? `ConstChoice` can probably be made public. I wish I could add it to `subtle`, but it seems to be unmaintained.

tarcieri:
I assume @ycscaly means there are certain methods defined on `Uint` and `ConstChoice` (both of which are already in the public API)  which are currently `pub(crate)`.

@ycscaly I'd say list the specific method names here and we can go through them on a case-by-case basis.

ycscaly:
I opened a PR that exposes everything I need, let me know what are your thoughts

ElijahVlasov:
It's currently impossible to do const trait implementations in stable rust. So it's not possible to use arithmetic operations over Uint/Ints in constant context. The arithmetic ops are defined in a const manner though in the private modules (e.g. [here](https://github.com/RustCrypto/crypto-bigint/blob/master/src/uint/add.rs)). Would be great to expose them.

tarcieri:
@ElijahVlasov this entire library has been very much written with `const fn` support in mind. Those `pub const fn` methods are not “in private modules”: that’s not how the Rust visibility system works. If they’re defined on a type in the public API they’re accessible.

We have chosen to factor related operations and tests in individual modules to help keep the code more maintainable.

PR #992, opened 2025-10-22
Expose functions

ycscaly:
Addresses #991


tarcieri:
@ycscaly it might be helpful to at least split this into separate PRs for `ConstChoice` vs `Uint`

ycscaly:
> @ycscaly it might be helpful to at least split this into separate PRs for `ConstChoice` vs `Uint`

No problem, but is this a valid approach, to simply `pub` the functions I need?

tarcieri:
Yes, we can discuss what should or should not be `pub` on the PR

ycscaly:
This PR is replaced by #994, #995

PR #994, opened 2025-10-23
Expose const_choice.rs functions

ycscaly:
Addresses https://github.com/RustCrypto/crypto-bigint/issues/991

### DIFF
--- a/src/const_choice.rs
+++ b/src/const_choice.rs
@@ -44,7 +44,7 @@ impl ConstChoice {
     /// Returns the truthy value if `value == Word::MAX`, and the falsy value if `value == 0`.
     /// Panics for other values.
     #[inline]
-    pub(crate) const fn from_word_mask(value: Word) -> Self {
+    pub const fn from_word_mask(value: Word) -> Self {
         debug_assert!(value == Self::FALSE.0 || value == Self::TRUE.0);
         Self(value)
     }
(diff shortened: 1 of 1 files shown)


src/const_choice.rs
```diff
@@ -327,7 +333,7 @@ impl<T> ConstCtOption<T> {
     ///
     /// **Note:** if the second element is `None`, the first value may take any value.
     #[inline]
-    pub(crate) const fn components_ref(&self) -> (&T, ConstChoice) {
+    pub const fn components_ref(&self) -> (&T, ConstChoice) {
```
tarcieri, line 336: This is the only one I'm a bit wary of exposing.  Notably it exposes the inner `T` which may be in an invalid state.

src/const_choice.rs
```diff
@@ -429,6 +435,20 @@ impl<const LIMBS: usize> ConstCtOption<(Uint<LIMBS>, Uint<LIMBS>)> {
...
+    /// # Panics
+    ///
+    /// Panics if the value is none with a custom panic message provided by
+    /// `msg`.
+    #[inline]
+    pub const fn expect(self, msg: &str) -> (Uint<LIMBS>, ConstChoice) {
+        assert!(self.is_some.is_true_vartime(), "{}", msg);
+        self.value
+    }
+}
```
tarcieri, line 450: I assume this is related to `components_ref` somehow?

tarcieri:
Per the build failures, the functions that we do end up exposing will need documentation

tarcieri:
I think these were obviated by #1040 which makes many of these methods `pub` by way of the new `ctutils` crate.

I'm going to go ahead and close this, but please file an issue against https://github.com/RustCrypto/utils/issues if there are methods from this PR you need which are still inaccessible.

PR #995, opened 2025-10-23
Expose Uint/Int Functions

ycscaly:
Addresses https://github.com/RustCrypto/crypto-bigint/issues/991

### DIFF
--- a/src/uint/div_limb.rs
+++ b/src/uint/div_limb.rs
@@ -134,13 +134,7 @@ pub(crate) const fn div2by1(u1: Word, u0: Word, reciprocal: &Reciprocal) -> (Wor
 /// In place of `v1` takes its reciprocal, and assumes that `v` was already pre-shifted
 /// so that v1 has its most significant bit set (that is, the reciprocal's `shift` is 0).
 #[inline(always)]
-pub(crate) const fn div3by2(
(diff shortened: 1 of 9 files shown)


src/uint/div_limb.rs
```diff
@@ -134,13 +134,7 @@ pub(crate) const fn div2by1(u1: Word, u0: Word, reciprocal: &Reciprocal) -> (Wor
...
 /// so that v1 has its most significant bit set (that is, the reciprocal's `shift` is 0).
 #[inline(always)]
-pub(crate) const fn div3by2(
-    u2: Word,
-    u1: Word,
-    u0: Word,
-    v1_reciprocal: &Reciprocal,
-    v0: Word,
-) -> Word {
+pub const fn div3by2(u2: Word, u1: Word, u0: Word, v1_reciprocal: &Reciprocal, v0: Word) -> Word {
```
fjarri, line 137: Where do you use this function? It seems pretty specialized to me, with a bunch of assumptions about values coming from other parts of the implementation.

ycscaly: in vartime division. If we get that merged, I wouldn't need to export it

tarcieri:
Closing as stale

PR #998, opened 2025-10-29
Int wrapping multiplication improvements

andrewwhitehead:
This optimizes `Int::wrapping_mul` and `Uint::checked_mul_int` to make use of `Uint::wrapping_mul`, and adds `Int::saturating_mul`.

### DIFF
--- a/benches/int.rs
+++ b/benches/int.rs
@@ -113,7 +113,7 @@ fn bench_concatenating_mul(c: &mut Criterion) {
 }
 
 fn bench_wrapping_mul(c: &mut Criterion) {
-    let mut rng = ChaChaRng::from_os_rng();
+    let mut rng = ChaCha8Rng::from_seed([7u8; 32]);
     let mut group = c.benchmark_group("wrapping ops");
 
     group.bench_function("wrapping_mul, I128xI128", |b| {


src/int.rs
```diff
@@ -56,20 +56,17 @@ impl<const LIMBS: usize> Int<LIMBS> {
...
 
     /// Maximum value this [`Int`] can express.
-    pub const MAX: Self = Self(Uint::MAX.shr(1u32)); // Bit sequence (be): 0111....1111
+    pub const MAX: Self = Self(Uint::MAX.shr_vartime(1u32)); // Bit sequence (be): 0111....1111
 
     /// Bit mask for the sign bit of this [`Int`].
     pub const SIGN_MASK: Self = Self::MIN; // Bit sequence (be): 1000....0000
 
-    /// All-one bit mask.
-    pub const FULL_MASK: Self = Self(Uint::MAX); // Bit sequence (be): 1111...1111
```
andrewwhitehead, line 71: This constant was public, but I can't imagine what it would be needed for, besides the internal usage in calculating `MINUS_ONE` which has been updated.

tarcieri:
@andrewwhitehead looks like this now has a conflict, possibly from #1008. Can you rebase and I can do another pass on this?

src/int.rs
tarcieri: We should probably apply the `unreachable_pub` lint

tarcieri (approved):

PR #1001, opened 2025-11-03
Bump `rand_core` to v0.10.0-rc-2

tarcieri:
Also migrates from `rand_chacha` to `chacha20` for an RNG


tarcieri:
Seems like we have a variety of strategies for seeds for randomized tests

PR #796, opened 2025-03-15
[WIP] cargo-checkct

tarcieri:
Wrapper for BINSEC providing Rust drivers and a way to write tests for constant-time operation:

https://github.com/Ledger-Donjon/cargo-checkct

This PR adds a GitHub workflow which runs `cargo-checkct` on every `push` event

PR #1005, opened 2025-11-16
Add gungraun constant-time benchmarks and checker

xyzzyz:
This is a proof of concept for using [Gungraun](https://gungraun.github.io/gungraun/latest/html/intro.html) for testing constant-timeness of the operations. The purpose of this PR is less so getting it merged, and more so gauging your interest in this work.

Gungraun uses Valgrind tooling to obtain accurate instruction counts for benchmark executions. This allows comparing runs on different inputs to see if the instruction counts are the same.

I added a number of Gungraun benchmarks, along with a simple binary to process the results (with the intention of using it in CI pipeline). Interestingly, I found that while Gungraun calculates the exact same instruction count for add/sub/mul and GCD operations, it finds tiny differences between `mod_pow` runs on different arguments (on the order of 0.005% difference in instruction counts).

Are you interested in including this in `crypto-bigint` test suite? If so, do you have any suggestions for next steps? Some of the ones I see are:
* adding more benchmarks to get higher coverage
* polishing it up and including in CI (for the purposes of constant timeness checking, but we could also get performance regression detection as side benefit)
* investigate exactly why the instruction counts are different for `mod_pow`.

Thanks for your feedback!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,6 +37,15 @@ checksum = "d8b59d472eab27ade8d770dcb11da7201c11234bef9f82ce7aa517be028d462b"
 
 [[package]]
 name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
+dependencies = [
+ "serde",
+]
+
+[[package]]
+name = "bincode"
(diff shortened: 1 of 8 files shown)


tarcieri:
@xyzzyz we're definitely interested in ways to better validate constant-time operation. It's something I've been very interested in working on, but also putting off until we can get to the next release.

I was also looking at `cargo-checkct` for this purpose: #796

Issue #793, opened 2025-03-13
Move away from traits implemented for specific `Uint` sizes?


fjarri:
Some traits are currently only implemented for specific `Uint` sizes, like `Encoding`, `ConcatMixed`/`SplitMixed`, `WideningMul`, and some others. This creates problems when working with big Uints of non-standard sizes, and increases compilation time because of all the macros creating the implementations.

I wonder if there's any desire to try and make these traits implementable generically, or make analogous traits that can be implemented generically, to exist side by side? 

E.g. in my library I would like to use `ConcatMixed`, but can't, since it's only implemented up to `U1024`. So I have a trait `Extendable`, such that
```rust
pub trait Extendable<Wide: Sized>: Sized {
    fn to_wide(&self) -> Wide;
    fn try_from_wide(value: &Wide) -> Option<Self>;
}

impl<const L: usize, const W: usize> Extendable<Uint<W>> for Uint<L> {
    fn to_wide(&self) -> Uint<W> {
        const {
            if W < L {
                panic!("Inconsistent widths in `Extendable::to_wide()`");
            }
        }
        ...
    }
    ...
}
```
That is, the output length has to be provided explicitly, and if it is not consistent with the `Uint` length, a compile-time error is raised. Arguably, it is a clearer compile-time error compared to what you get if you try to use `ConcatMixed` with a type for which it is not implemented. Although, of course, in order to get it, you need to actually use `to_wide()` somewhere in your code, which is a downside.

I have a similar trait for the widening multiplication, where there's a compile-time check that the sum of the lengths of two arguments is equal to the output length. 

For serialization, I have
```rust
pub trait BoxedEncoding: Sized {
    fn to_be_bytes(&self) -> Box<[u8]>;
    fn try_from_be_bytes(bytes: &[u8]) -> Result<Self, String>;
}
```
The problem with having it defined locally is that I have to wrap `Uint`s in a newtype in order to use it. 

Alternatively, it is possible to implement generic non-allocating `Serialize/Deserialize` for `Uint` (without it having to depend on `Encoding`), but it requires the use of `zerocopy` or `bytemuck` to transmute between `&[Limb]` and `&[u8]`.

Any interest of bringing any of that into `crypto-bigint`?

tarcieri:
Ideally I think all traits would be implemented for all sizes, yes. If you can remove the hardcoded size limitations and implement things for all sizes, that'd be great. The const comparison is a nice hack.

[...]

PR #1007, opened 2025-11-19
Add `U8320` and `U12288`

Fethbita:
In https://github.com/RustCrypto/crypto-bigint/issues/228 and https://github.com/RustCrypto/crypto-bigint/pull/229 extra sizes up to 8k were implemented. In some usages, we need the extra sizes to go up to 16k (for example when two 6k numbers are multiplied).

For our use-case `U12288` and `8320` in addition to extra-sizes would be enough, but for completeness, I added all values in 64 bit increments to a new feature `extra-sizes-16k`.

Let me know if you would rather prefer if we only add these two extra-sizes we need to `extra-sizes` feature.

### DIFF
--- a/src/uint/extra_sizes.rs
+++ b/src/uint/extra_sizes.rs
@@ -110,7 +110,7 @@ impl_uint_aliases! {
     (U8064, 8064, "8064-bit"),
     (U8128, 8128, "8128-bit"),
     (U8320, 8320, "8320-bit"),
-    (U12288, 12288, "12288-bit"),
+    (U12288, 12288, "12288-bit")
 }
 
 impl_uint_concat_split_even! {
@@ -160,5 +160,5 @@ impl_uint_concat_split_even! {
     U7936,
     U8064,
     U8320,
-    U12288
+    U12288,
 }


tarcieri:
I don't think it makes sense to just keep adding sizes indefinitely. The type aliases are mostly for convenience. The core implementation is const generic.

What are you actually using them for? Can you just define your own type aliases?

Fethbita:
Yeah, it makes sense. I use `U12288` for `6k*6k` operation and `U8320` for `8k*128` operation. I could define my own type aliases but `impl_uint_aliases` and `impl_uint_concat_split_even` macros are not exported. How would you recommend I do it?

tarcieri:
We could potentially export them.

Otherwise it’s just `type Foo = Uint<{ nlimbs!(12288) }>;`

fjarri:
It is not exactly equivalent since some traits involving size modifications are implemented only for size aliases (`Split`, `Concat`, `RemMixed` etc)

Fethbita:
Indeed, even if I made the `impl_uint_aliases` and `impl_uint_concat_split_even` public, then there are other things that must become public:
```
error[E0603]: module `traits` is private
   --> /tmp/crypto-bigint/src/uint/macros.rs:92:22
    |
 92 |         impl $crate::traits::ConcatMixed<Uint<{ <$name>::LIMBS / 2 }>> for Uint<{ <$name>::LIMBS / 2 }>
    |                      ^^^^^^ private module
    |
note: the module `traits` is defined here
   --> /tmp/crypto-bigint/src/lib.rs:214:1
    |
214 | mod traits;
    | ^^^^^^^^^^
help: consider importing this trait instead
    |
 92 -         impl $crate::traits::ConcatMixed<Uint<{ <$name>::LIMBS / 2 }>> for Uint<{ <$name>::LIMBS / 2 }>
 92 +         impl crypto_bigint::ConcatMixed<Uint<{ <$name>::LIMBS / 2 }>> for Uint<{ <$name>::LIMBS / 2 }>
    |
```
Instead, I changed the PR to only add the 2 types that I need, if that it okay.

tarcieri:
Perhaps we should address #793 as an alternative?

I would prefer to avoid having to start adding a bunch of bespoke sizes for various use cases. We do that in `hybrid-array` by necessity since there's no alternative, and it carries a high maintenance burden, and at least there slows down the compiler whenever we add more impls.

Fethbita:
https://github.com/RustCrypto/crypto-bigint/issues/793 looks like a good suggestion, but I don't think I could implement it here as I am not that familiar with traits and generics.

fjarri:
I suppose I could. Completely forgot I even filed that issue.

tarcieri:
@fjarri that'd be great

fjarri:
Would you rather I replaced the existing traits, or added another set?

tarcieri:
I’ll leave that at your discretion, but if we could minimize the total number of traits that’d be great

fjarri:
Started #1008

tarcieri:
This should be obviated by #1008

Fethbita:
Edit: Created #1095 for this comment

PR #1008, opened 2025-11-21
Use traits implemented for all `Uint`/`Int` types

### LINKED ISSUE #793: Move away from traits implemented for specific `Uint` sizes?

fjarri:
Closes #793 

- `Split` and `Concat` reworked to be implementable for all `Uint`s generically.
- `SplitMixed` and `ConcatMixed` removed (`Split` and `Concat` implement their functionality).
- `RemMixed` removed - `Rem` already allows custom rhs sizes. Moreover, the name `RemMixed`/`rem_mixed()` was misleading because it actually used variable time division inside.
- `Encoding` implemented for all `Uint`s. `Encoding::Repr` for `Uint` is now an `EncodedUint` struct instead of an array.
- Macros deriving `Encoded`, `SplitMixed`, `ConcatMixed`, and `RemMixed` are removed. 
- Relaxed the bound on the error of `Encoding::Repr as TryFrom<&[u8]>` from a concrete type to a `core::error::Error` trait.

Notes:
- The new split/concat errors for incorrect sizes, while being compile time, may not be entirely informative. First, I can't format the error string in the `panic!()` (without another dependency), second, the compiler often does not say which monomorphization caused the error. 
- `Uint::to_le/be_bytes()` are quite similar to `Uint::write_le/be_bytes()`, the former being `const fn`, the latter being faster (since they can operate with `Word`-sized chunks instead of single bytes). 
- the `extra_sizes` feature may be removed since now it only defines `Uint` aliases and doesn't impl any traits. 
- We use `serdect` for constant-time serialization, so we are concerned about serializing secrets... But `Encoding` methods litter the stack with the serialized data (not introduced in this PR, it was the same before). Is that a concern as well?

### DIFF
--- a/src/uint/concat.rs
+++ b/src/uint/concat.rs
@@ -5,12 +5,13 @@ const fn concat<const L: usize, const H: usize, const O: usize>(
     hi: &Uint<H>,
 ) -> Uint<O> {
     const {
-        if L + H != O {
-            panic!(concat![
+        assert!(
+            L + H == O,
+            concat![
                 "The size of the declared type of `Uint` concatenation is not equal to ",
                 "the sum of the sizes of argument types. ",
-            ]);
-        }
+            ]
+        );
     }
 
     let mut result = Uint::<O>::ZERO;
--- a/src/uint/split.rs
+++ b/src/uint/split.rs
@@ -4,12 +4,13 @@ const fn split<const I: usize, const L: usize, const H: usize>(
     value: &Uint<I>,
 ) -> (Uint<L>, Uint<H>) {
     const {
-        if L + H != I {
-            panic!(concat![
+        assert!(
+            L + H == I,
+            concat![
                 "The sum of the sizes of the declared types of `Uint` split is not equal to ",
                 "the size of the input type. ",
-            ]);
-        }
+            ]
+        );
     }
 
     let mut lo = Uint::<L>::ZERO;


src/traits.rs
```diff
@@ -900,33 +866,26 @@ pub trait Invert {
...
 }
 
-/// Widening multiply: returns a value with a number of limbs equal to the sum of the inputs.
-pub trait ConcatenatingMul<Rhs = Self>: Sized {
-    /// Output of the widening multiplication.
-    type Output: Integer;
-
+/// Widening multiply: returns a value with a number of limbs greater than or equal to
+/// the sum of the inputs.
+pub trait ConcatenatingMul<Output, Rhs = Self>: Sized {
```
tarcieri, line 886: I'm a bit unclear why `Output` needs to be moved from an associated type to a generic parameter here. Are there overlapping impls I'm missing somewhere, or is it something else?

tarcieri:
Definitely liking this overall direction

src/traits.rs
fjarri: Because it cannot be derived without const generic expressions, or macro-based implementations for each specific type (which is what this PR removes).

tarcieri: Aha! Makes sense, thanks

fjarri:
@tarcieri I would like to implement Serialize/Deserialize for all Uints as well, there are two options:
- use `bytemuck`, which requires implementing some `unsafe` marker traits, which is currently prohibited by an attribute in the top module;
- add functions to `serdect` that work with byte iterators instead of `&[u8]` (so that I can iterate over an integer's bytes without actually using a temporary buffer)

What do you think is a better way?

Actually, the second option may not be possible with `serde` API... need to check that

*Edit:* yes, I don't think it will be possible. `serde` needs some kind of a `serialize_bytes_iter()` method, and given the glacial pace with which `serde` moves, and the maintainer's distaste for any changes whatsoever, it'll takes months at best to get that in.

tarcieri:
> which is currently prohibited by an attribute in the top module

It's not prohibited (it's deny instead of forbid) and we do have unsafe a few places, however I would be a bit more concerned about adding an additional hard dependency, even if it is a widely used one. We can consider it, though.

> the second option may not be possible with serde API... need to check that

It's possible with serde but not possible with the way serdect works, AFAIK

fjarri:
Sorry, edited your comment by mistake at first...

> It's possible with serde but not possible with the way serdect works, AFAIK

How is it possible? `serialize_seq`/`serialize_tuple` will lead to inefficient representations.

tarcieri:
@fjarri yes, that's what I meant, it's possible with `serde`, but uses those representations that aren't necessarily constant-time on formats like MsgPack, which is why `serdect` migrated away from them

tarcieri:
Another option would be to make the `serde` feature dependent on `hybrid-array` and use `typenum` to compute the size of the output array, which would require making the `ArrayEncoding` impl valid for all sizes too (or at least, all sizes that can be represented by `hybrid-array`)

There are a couple ways to do the linkage between const generics and typenum. One is [`typenum::U`](https://docs.rs/typenum/latest/typenum/generic_const_mappings/type.U.html). The other is [`hybrid_array::ArrayN`](https://docs.rs/hybrid-array/latest/hybrid_array/type.ArrayN.html).

fjarri:
Implemented with `bytemuck`.

tarcieri:
Okay, that's fine for now. If you're done can you mark the PR ready for review?

fjarri:
Ok, but let me check the thing with `Encoding::Repr as TryFrom<&[u8]>` bound before merging.

src/uint/encoding.rs
```diff
@@ -204,58 +201,123 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
-        let mut j = 0;
+/// [`Uint`] encoded as bytes.
+#[derive(Default, Copy, Clone, Debug, PartialEq, Eq)]
+pub struct EncodedUint<const LIMBS: usize>(Uint<LIMBS>);
 
-        while j < Limb::BYTES {
-            ret[i * Limb::BYTES + j] = limb_bytes[j];
-            j += 1;
+impl<const LIMBS: usize> EncodedUint<LIMBS> {
+    const fn as_le_bytes(value: &Uint<LIMBS>) -> Self {
```
tarcieri: The `as_*` function prefix is for borrowed->borrowed conversions per: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

```suggestion
    const fn to_le_bytes(value: &Uint<LIMBS>) -> Self {
```

src/uint/encoding.rs
```diff
@@ -204,58 +201,123 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+                dst_bytes[j] = src_bytes[j];
+                j += 1;
+            }
+
+            i += 1;
         }
+        Self(buffer)
+    }
+
+    const fn as_be_bytes(value: &Uint<LIMBS>) -> Self {
```
tarcieri: Likewise here:

```suggestion
    const fn to_be_bytes(value: &Uint<LIMBS>) -> Self {
```

tarcieri:
@fjarri yeah, I remember you expressing wanting to have the concrete error type for the `TryFrom` bound: #261

src/uint/encoding.rs
fjarri: Oh yes, I just put a placeholder there and forgot to fix it. `to_le_bytes` doesn't fit well either, because this method is a constructor. Still thinking of a better name.

tarcieri: `encode_to_be`/`encode_to_le` perhaps?

fjarri: Changed to `new_be()`/`new_le()`

fjarri:
I replaced the strict type with a bound on `core::error::Error`, I think this will be enough.

src/uint/encoding.rs
```diff
@@ -204,58 +203,140 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+            }
+
+            i += 1;
+        }
+        Self(buffer)
     }
+}
 
-    ret
+impl<const LIMBS: usize> AsRef<[u8]> for EncodedUint<LIMBS> {
```
tarcieri, line 292: Can you add a `Deref` impl as well?

fjarri: A deref to what? I only implemented the bounds from `Encoding`

tarcieri: `Index` would also be helpful

tarcieri: `Deref<Target = [u8]>` so it acts more transparently like a byte slice.

(I'm just going by what breaks when I try this change against various repos)

fjarri: Should I add a corresponding bound to `Encoding` then?

tarcieri: Sure.

(While I'm at it, this breaks `crypto-primes`, though I assume you're okay doing the updates for that)

fjarri: Where are you seeing the need for `Deref`, by the way? The original `Repr` was an `u8` array which does not implement `Deref`. So I can't even add this bound to `Encoding::Repr` because it conflicts with the implementation for `Limb`.

tarcieri: `[u8; N]` -> `[u8]` uses [special deep magic](https://doc.rust-lang.org/std/ops/trait.CoerceUnsized.html).

I was seeing this issue where the resulting encoded value was being passed to an API expecting a slice. `sm2` is one example, I think `ed448-goldilocks` was another.

It's fine if you don't add the bound but I'd prefer to keep the `Deref` impl.

fjarri: > I was seeing this issue where the resulting encoded value was being passed to an API expecting a slice.

The deep magic quoted seems to be nightly-only at the moment. Can they just use `AsRef`?

tarcieri: I linked that page more as a reference for something the compiler already does, which is an unsizing coercion. The API may be nightly-only, but the behavior is very much baked into the language and stable.

This page probably does a better job of explaining what those coercions are:

https://doc.rust-lang.org/std/marker/trait.Unsize.html

The point being that the compiler will automatically coerce `[u8; N]` to `[u8]` where appropriate via such an unsizing coercion. To get equivalent behavior with a user-defined smart pointer type like we're implementing, you need to use a `Deref` coercion instead.

fjarri: No, I can impl `Deref` for `EncodedUint`, but the implementation of `Encoding` for `Limb` has `Repr = [u8; 8]`, so I cannot add a `Deref` bound to the trait. 

I am always wary of adding Deref impls unless they lead to significant code noise reduction, and it doesn't seem to be the case here. Plus, I am questioning whether those crates should be relying on Deref at all when it's not even in the trait bounds. Pushed a commit with it, but I want to register my objection to it.

tarcieri:
@fjarri it seems like you're using `bytemuck` to do effectively two things:

1. call the equivalent of e.g. [`u64::to_be`](https://doc.rust-lang.org/std/primitive.u64.html#method.to_be) / [`u64::from_be`](https://doc.rust-lang.org/std/primitive.u64.html#method.from_be) (and so on for `u32` and `_le`)
2. perform a pointer cast

I guess I'm fine with that to get the feature implemented, but it seems like it should be fairly straightforward to remove and replace with something that calls the equivalent core functionality to do the same.

fjarri:
Perhaps, but that would require an `unsafe` block. Should I do that?

src/uint/encoding.rs
tarcieri: I would consider it to be a smart pointer type that's trying to store a `[u8; N]` in a `[Word; LIMBS]`, which is effectively a workaround for not being able to use `[u8; N]` because we can't calculate `N` from `LIMBS`.

I would prefer for it to have ergonomics that make it more like `[u8; N]` where possible, especially for real-world use cases.

tarcieri:
Yes, that's fine, there are other `unsafe` pointer casts already, e.g. `Uint::as_(mut_)words`

src/uint/concat.rs
```diff
@@ -1,50 +1,71 @@
...
+    lo: &Uint<L>,
+    hi: &Uint<H>,
+) -> Uint<O> {
+    const {
+        if L + H != O {
+            panic!(concat![
+                "The size of the declared type of `Uint` concatenation is not equal to ",
+                "the sum of the sizes of argument types. ",
+            ]);
+        }
```
tarcieri:
```suggestion
        assert!(L + H == O,
            concat![
                "The size of the declared type of `Uint` concatenation is not equal to ",
                "the sum of the sizes of argument types. ",
            ]
        );
```

fjarri: Oh I remember this not working when I filed the original issue

fjarri: Fixed

fjarri:
Checking out the changes required in `crypto-primes`, and it seems like this PR makes writing certain generic functions impossible. Namely, we need to extend a given generic `Uint` by 128 bits, perform some operations, and then drop the high 128 bits. Formerly this could be done, provided the `Uint` passed by the user had the required `ConcatMixed` defined.

One can argue that it just makes the algorithm deficiency immediately obvious to the library writer instead of the end user realizing their chosen `Uint` size doesn't impl the required trait. But still I wonder if this PR will break some downstream code, with the required changes going beyond some simple fixes.

(In my case, something like `Uint::div_wide()` that takes lo and hi parts of the dividend would help)

*Edit:* also, as I noted in the description, debugging the compile time failures is kind of hard since the compiler doesn't provide a "backtrace". Although it is probably much simpler in the new code where you can tell which change led to the error.

tarcieri:
@fjarri oof, I did ask about `crypto-primes`. So you have a workaround, or need to revert?

fjarri:
I'll think about it, but even if I do find a solution, other people might struggle with something else. The question is, is that because this PR makes such generic code "fail early", so to speak, or because the PR is fundamentally flawed. I'm leaning towards reverting and giving it more consideration.

Honestly, these lacunas in the compiler functionality are the most frustrating things about Rust.

Fethbita:
> Checking out the changes required in `crypto-primes`, and it seems like this PR makes writing certain generic functions impossible. Namely, we need to extend a given generic `Uint` by 128 bits, perform some operations, and then drop the high 128 bits. Formerly this could be done, provided the `Uint` passed by the user had the required `ConcatMixed` defined.

I tried to find where this usage is in `crypto-primes` but couldn't. Could you point out where it is used? 

I would also like to use the same operation on my code (generic `U4096` -> `U4224` -> operations) but could not figure out how to achieve it.

fjarri:
@Fethbita https://github.com/entropyxyz/crypto-primes/blob/master/src/hazmat/primecount.rs#L76

Actually, now that I look at it it just doubles the width of integers, which I guess is easier from the typing point of view, but it presents the same problems with this PR.

tarcieri:
@fjarri so what's the verdict? should we revert, or have you found a workaround?

can you maybe put together a playground with a minimal example of the problem and I can ask around for some solutions?

Fethbita:
> Actually, now that I look at it it just doubles the width of integers, which I guess is easier from the typing point of view, but it presents the same problems with this PR.

I saw that as well. What do you mean by `but it presents the same problems with this PR`? Currently without this PR, it is possible to double the integer size with the same method used just like the link you shared, but this is also possible after the PR, isn't it? I don't think increasing the width of an integer by 128 was possible before, and it is also not possible after (and have it connected with the traits). Am I wrong?

fjarri:
@Fethbita , before the PR you could use the `ConcatMixed` trait. The problem is, it was defined for a small number of uints (up to 1024 I think), because of quadratic expansion of generated code. 

> but this is also possible after the PR, isn't it?

Well, you'd have to have the second generic parameter (`RHS_LIMBS`), but since now it's not determined by a trait bound, the user would have to specify it explicitly, which is quite annoying. 

> so what's the verdict? should we revert, or have you found a workaround?

I would vote for revert. I can work around this stuff, but I expect a lot of frustration from other downstream users. The thing is, the approach in this PR works when you have concrete types for inputs and outputs of concat/split; but if you want to temporarily extend and then shrink back an integer (like in the primecount link above), this does not work - the compiler cannot derive the extended type. 

Having said that, I will think about fixing the primecount function so that it doesn't rely on `Concat` because it just means that its potential users might suddenly hit a problem when applying it to some Uint without an alias.

tarcieri:
Opened #1015 to revert

andrewwhitehead:
Can the encoding changes be separated out from the rest? They don't seem to be dependent.

fjarri:
Good idea, see #1016

fjarri:
Btw, the removal of `RemMixed` may deserve its own PR as well.

tarcieri:
I was also wondering if `ConcatenatingMul` could replace `ConcatMixed` independently of everything else

andrewwhitehead:
For adding a fixed number of limbs, it could be more useful to expose the `ExtendedUint` used in bingcd and add functionality there as needed. `Uint` could have `extended_mul` and `extended_square` methods returning this type.

PR #1010, opened 2025-11-24
BREAKING: Make random_mod platform-independent

### LINKED ISSUE #1009: random_mod is not platform-independent

mrdomino:
Replaces the fancy platform-word-based logic with a simple call to `random_bits_core` with rejection sampling. Each retry has p > 0.5 (worst case) of selecting a number inside the range; compare against [OpenBSD][0]:

```c
/*
 * This could theoretically loop forever but each retry has
 * p > 0.5 (worst case, usually far better) of selecting a
 * number inside the range we need, so it should rarely need
 * to re-roll.
 */
for (;;) {
	r = arc4random();
	if (r >= min)
		break;
}
```

**NB**. This change will generally result in different numbers being produced by `Uint::random_mod`, as evidenced by the test change.

Fixes #1009

[0]: https://github.com/openbsd/src/blob/0c90e2b526b26d9f50dde7f7712cec107cda3326/lib/libc/crypt/arc4random_uniform.c#L43-L53

### DIFF
--- a/src/uint/rand.rs
+++ b/src/uint/rand.rs
@@ -267,12 +267,12 @@ mod tests {
     fn random_mod_platform_independence() {
         let mut rng = get_four_sequential_rng();
 
-        let modulus = NonZero::new(U256::from_u32(7776)).unwrap();
-        let mut vals = [U256::ZERO, U256::ZERO, U256::ZERO, U256::ZERO];
+        let modulus = NonZero::new(U256::from_u32(8192)).unwrap();
+        let mut vals = [U256::ZERO, U256::ZERO, U256::ZERO, U256::ZERO, U256::ZERO];
         for val in &mut vals {
             random_mod_core(&mut rng, val, &modulus, modulus.bits_vartime()).unwrap();
         }
-        let expected = [7020, 5991, 55, 1375];
+        let expected = [55, 3378, 2172, 1657, 5323];
         for (want, got) in expected.into_iter().zip(vals.into_iter()) {
             assert_eq!(got, U256::from_u32(want));
         }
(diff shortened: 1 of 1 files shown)


src/uint/rand.rs
```diff
@@ -288,6 +262,26 @@ mod tests {
         );
     }
 
+    /// Make sure random_mod output is consistent across platforms
+    #[test]
+    fn random_mod_platform_independence() {
+        let mut rng = get_four_sequential_rng();
+
+        let modulus = NonZero::new(U256::from_u8(7)).unwrap();
```
mrdomino: This is indeed a silly test case; however it does show the platform-dependent behavior of the old code.

I would be fine with rewriting it to run against a longer modulus or for more iterations though.

mrdomino:
#1011 shows failing test results. ~Surprisingly (to me) i686 fails but wasm32 passes… (the behavior I was seeing elsewhere was variance between amd64 / arm64 on the one hand and wasm32 on the other.)~ (looks like wasm32 is just built, not tested.)

src/uint/rand.rs
```diff
@@ -288,6 +262,26 @@ mod tests {
...
+        let mut val = U256::ZERO;
+        let _ = random_mod_core(&mut rng, &mut val, &modulus, modulus.bits_vartime());
```
erik-3milabs: we should probably also check that the produced value is actually the same.

mrdomino: Added.

mrdomino: Test now exercises the rejection sampling logic.

tarcieri (approved):

tarcieri:
Opened #1017 to revert this. It seems it never actually passed CI (my bad)

PR #1014, opened 2025-11-25
Add `as_montgomery_mut` for `BoxedMontyForm`

ElijahVlasov:
This PR adds a mutable access method for the underlying Montgomery form `BoxedUint` in `BoxedMontyForm` type.

### DIFF
--- a/src/modular/boxed_monty_form.rs
+++ b/src/modular/boxed_monty_form.rs
@@ -380,4 +380,18 @@ mod tests {
         assert_eq!(zero.div_by_2(), zero);
         assert_eq!(one.div_by_2().mul(&two), one);
     }
+
+    #[test]
+    fn as_montgomery_mut() {
+        let modulus = Odd::new(BoxedUint::from(9u8)).unwrap();
+        let params = BoxedMontyParams::new(modulus);
+        let one = BoxedMontyForm::one(params.clone());
+        let two = one.add(&one);
+        let four = two.mul(&two);
+        let mut x = two.clone();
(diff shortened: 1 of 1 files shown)


src/modular/boxed_monty_form.rs
```diff
@@ -242,6 +242,12 @@ impl BoxedMontyForm {
         &self.montgomery_form
     }
 
+    /// Mutably access the [`BoxedMontyForm`] value in Montgomery form.
+    pub fn as_montgomery_mut(&mut self) -> &mut BoxedUint {
+        debug_assert!(&self.montgomery_form < self.params.modulus());
```
tarcieri: The potential for this assertion to be tripped is more when the borrow completes than before, as the type's invariants otherwise prevent this. Not sure it's useful, and not present in the other similar methods

ElijahVlasov: True. Removed the check

tarcieri:
@ElijahVlasov can you add a small test? Codecov isn't happy.

ElijahVlasov:
> @ElijahVlasov can you add a small test? Codecov isn't happy.

Would it be fine to add it in the `tests` module below?

https://github.com/ElijahVlasov/crypto-bigint/blob/f282fec141844fecf2a1258250bc4cccabe3ee82/src/modular/boxed_monty_form.rs#L360

tarcieri:
Yes that's fine

tarcieri (approved):

PR #1015, opened 2025-11-25
Revert "Use traits implemented for all `Uint`/`Int` types"

tarcieri:
This reverts commit f268e8db4ff8c1db553b325a3916c41fe1f90868 (#1008)

See discussion on #1008 for context


Fethbita:
@tarcieri, could you re-open #793 since this was reverted?

tarcieri:
We should probably reopen an issue, but I don't think #793 is the right one, unless @fjarri has another idea for a design. Otherwise I'd say we tried that and it didn't work out, at least given the current limitations of the Rust type system.

Fethbita:
Makes sense, maybe https://github.com/RustCrypto/crypto-bigint/issues/228?

tarcieri:
@Fethbita I'm curious if we can find a solution for what you want to do which doesn't involve having to add a bunch of bespoke sizes

Fethbita:
@tarcieri  I am not sure, it would still be very helpful to me to have `U8320` and `U12288` types but without adding those, I am not sure how to solve it.

PR #990, opened 2025-10-22
Implement `Encoding` for `Int` Aliases

ycscaly:
This PR adds a missing functionality for `Ixxx` types by implementing the `Encoding` trait.

Work originally done by @erik-3milabs

PR #1016, opened 2025-11-26
Implement `Encoding` for all `Uint`s

fjarri:
A part of #1008 that only deals with `Encoding`:
- `Encoding` implemented for all `Uint`s. `Encoding::Repr` for `Uint` is now an `EncodedUint` struct instead of an array.
- Macros deriving `Encoding` are removed. 
- Relaxed the bound on the error of `Encoding::Repr as TryFrom<&[u8]>` from a concrete type to a `core::error::Error` trait.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -167,6 +167,8 @@
 #[macro_use]
 extern crate alloc;
 
+pub use uint::encoding::{EncodedUint, TryFromSliceError};
+
 #[cfg(feature = "rand_core")]
 pub use rand_core;
 #[cfg(feature = "rlp")]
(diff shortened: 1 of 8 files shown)


fjarri:
Note that this can be extended to replace #990 (possibly in a follow-up), and also `Encoding` can be implemented for `BoxedUint`.

fjarri:
What is up with the cross-compile tests? I think there's the same problem in `master`?

tarcieri:
Yeah, very strange

fjarri:
Could it be caused by #1010? I notice they timeouted there too

tarcieri:
Huh, you're right, guess I didn't notice at the time I merged?

Maybe we should enable some branch protections

tarcieri:
@fjarri all right, #1017 reverted #1010, can you try rebasing?

PR #1017, opened 2025-11-26
Revert "BREAKING: Make random_mod platform-independent"

tarcieri:
This reverts commit 62b90b8b39263d4970853e2ecfada62646bc8428 (#1010)

We're noticing CI failures on cross which seem to be related to this
change, e.g.

https://github.com/RustCrypto/crypto-bigint/actions/runs/19647957432/job/56267760132

It looks like it's potentially getting stuck in an infinite loop.

cc @mrdomino @fjarri

### DIFF
--- a/src/uint/boxed/rand.rs
+++ b/src/uint/boxed/rand.rs
@@ -28,7 +28,7 @@ impl RandomBits for BoxedUint {
         }
 
         let mut ret = BoxedUint::zero_with_precision(bits_precision);
-        random_bits_core(rng, &mut ret.limbs, bit_length).map_err(RandomBitsError::RandCore)?;
+        random_bits_core(rng, &mut ret.limbs, bit_length)?;
         Ok(ret)
     }
 }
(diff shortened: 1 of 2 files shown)


tarcieri:
Yeah not sure exactly what's happening

fjarri:
I guess without that removed masking code in `random_mod_core()` the rejection takes too long on 32-bit platforms.

mrdomino:
Not sure that’s it; the i686 tests were passing, and in a different repo I confirmed that that algorithm worked against wasm32.

When I get a chance I’ll give a go at trying to reproduce the cross environment and see if I can figure out what’s going on. Anyone else who’d like to take a swing at it, be my guest.

tarcieri:
Notably the cross test that was seemingly wedged was actually on `aarch64-unknown-linux-gnu`, so a 64-bit target, but I have no idea why a problem would arise there but not locally on my M1/M2 Macs

PR #1018, opened 2025-11-26
Reapply "BREAKING: Make random_mod platform-independent" (#1017)

mrdomino:
This reverts commit 5784b132976ff893c42877558e25e6916511ab74.

PR #1020, opened 2025-11-27
Reapply "BREAKING: Make random_mod platform-independent"

mrdomino:
This reverts #1017, reapplying #1010. See also: https://github.com/RustCrypto/crypto-bigint/pull/1018#issuecomment-3584312683.

It turns out that the previous approach ran afoul of a likely LLVM (or possibly rustc) bug.

This time, I tried removing `random_mod_core` entirely and instead writing `random_mod` and `try_random_mod` as rejection loops over `random_bits` and `try_random_bits`. It is a little unfortunate that we write the rejection loop three times, but somehow, this does not (seem to) run afoul of the bug.

### DIFF
--- a/src/uint/boxed/rand.rs
+++ b/src/uint/boxed/rand.rs
@@ -1,10 +1,7 @@
 //! Random number generator support.
 
 use super::BoxedUint;
-use crate::{
-    NonZero, RandomBits, RandomBitsError, RandomMod,
-    uint::rand::{random_bits_core, random_mod_core},
-};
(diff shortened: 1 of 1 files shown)


mrdomino:
@tarcieri WDYT? I could go either way on this personally; it does seem a bit of a shame to lose `random_mod_core`. (I’d initially forgotten about `BoxedUint` when writing this.)

mrdomino:
(fyi #1009)

I don't think I like this approach. With that bug lurking and the BoxedUint loop untested, this could still hang in some cases; I don't know why it doesn't.

PR #1021, opened 2025-11-27
BREAKING: Make `random_mod_core` platform-independent (take 2)

### LINKED ISSUE #1009: random_mod is not platform-independent

mrdomino:
This introduces a fairly robust platform-independence test for `random_mod_core`. The test exercises the rejection sampling and also the way multi-word `Uint`s are constructed, which changed in this PR, as you can see in commit d2f4b89:
```diff
-            U256::from_be_hex("C54302F2EB1E2F69C3B919AE0D16DF2259CD1A8A9B8EA8E0862878227D4B40A3")
+            U256::from_be_hex("C3B919AE0D16DF2259CD1A8A9B8EA8E0862878227D4B40A3C54302F2EB1E2F69")
```

The difference, under most circumstances, is just that word order will be rotated; the old entropy gathering approach independently rejection-sampled the high word first, then built up the rest, so the resulting `Uint` was sort of like `(MSW, LSW, LSW+1, …, MSW-1)` over the raw entropy stream, except in cases where the high-word rejection sample triggered, in which case it was a trickier transform. The new interpretation is simply little-endian, both bytewise and wordwise, aligned to 64-bit word boundaries.

I elected to make this change for two reasons:
1. It is more aesthetically pleasing to me personally.
2. More importantly, with this change we actually _are_ changing the way we sample entropy in `random_mod`, particularly in rare cases of multiple rejection samples over the most significant word. Rather than make this a rare, subtle, difficult-to-detect change that will bite people unexpectedly, I elected to make this a big, obvious, easy-to-detect change that people will usually notice.

To work around the apparent compiler bug that required #1010 to be reverted (See #1018 for exploration of the bug), I also elected to make the comparison loop variable-time rather than using `ct_lt`, since the latter is what seems to trigger the bug. My defense is that `random_mod` is already variable-time; it leaks RNG state via timing information every time it rejection-samples. (It might be a good idea to rename it to `random_mod_vartime`, since AFAICT leaking this timing information is unavoidable in the general case.)

By making this change, we are able to carry through the prior copy-elision optimization from master, resulting in an implementation that performs about at parity with master in the benchmarks I ran.

One other note: I elected to take 64-bit rather than 32-bit words as primary here, i.e., 32-bit platforms need to adjust by sampling and discarding an extra word in odd cases rather than 64-bit platforms needing to adjust by only sampling half a word in those cases. Both are doable, but operating in terms of 64-bit word sizes frees us from the constraint of our RNG needing to be "4-byte sequential."

Fixes #1009

### DIFF
--- a/src/uint/rand.rs
+++ b/src/uint/rand.rs
@@ -147,7 +147,7 @@ where
             // Read entropy in 64-bit blocks, even on 32-bit platforms.
             let _ = rng.try_next_u32()?;
         }
-        // Do a manual, variable-time comparison loop here to avoid a copiler bug that causes a
+        // Do a manual, variable-time comparison loop here to avoid a compiler bug that causes a
         // hang on linux-aarch64 under `--release` with `Uint` of 5 or more limbs using `ct_lt`.
         let x = x.as_ref();
         let n = n.as_ref().as_ref();


mrdomino:
Benchmarks:
```
bounded random/random_mod, U1024
                        time:   [78.558 ns 83.597 ns 88.778 ns]
                        change: [+30.994% +37.066% +43.381%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, small bound: Collecting 100 samples in estimated 5.0001 s (145M iter
bounded random/random_mod, U1024, small bound
                        time:   [26.914 ns 29.058 ns 31.129 ns]
                        change: [+46.284% +56.005% +66.055%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, 512 bit bound low: Collecting 100 samples in estimated 5.0001 s (110
bounded random/random_mod, U1024, 512 bit bound low
                        time:   [57.944 ns 61.234 ns 64.603 ns]
                        change: [+25.267% +30.131% +35.484%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, 512 bit bound hi: Collecting 100 samples in estimated 5.0001 s (63M
bounded random/random_mod, U1024, 512 bit bound hi
                        time:   [76.614 ns 81.337 ns 86.333 ns]
                        change: [+24.949% +31.362% +38.310%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, tiny high limb: Collecting 100 samples in estimated 5.0002 s (100M i
bounded random/random_mod, U1024, tiny high limb
                        time:   [49.912 ns 49.936 ns 49.963 ns]
                        change: [+28.153% +28.293% +28.431%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high severe
```

Personally I would gladly take this hit for the simpler interpretation of the entropy stream.

mrdomino:
Unfortunately, my attempt at optimizing this ran afoul of the infinite hang bug. Current benchmarks show quite a large regression, probably because of the extra copies that are saving us from LLVM applying the bad optimization:
```
bounded random/random_mod, U1024
                        time:   [137.73 ns 146.24 ns 154.77 ns]
                        change: [+133.55% +144.24% +156.02%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, small bound: Collecting 100 samples in estimated 5.0002 s (119M iter
bounded random/random_mod, U1024, small bound
                        time:   [51.103 ns 53.731 ns 56.439 ns]
                        change: [+182.25% +195.11% +207.77%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, 512 bit bound low: Collecting 100 samples in estimated 5.0003 s (67M
bounded random/random_mod, U1024, 512 bit bound low
                        time:   [73.848 ns 78.540 ns 83.636 ns]
                        change: [+68.881% +78.635% +87.389%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, 512 bit bound hi: Collecting 100 samples in estimated 5.0001 s (35M
bounded random/random_mod, U1024, 512 bit bound hi
                        time:   [139.34 ns 148.02 ns 157.15 ns]
                        change: [+122.37% +132.42% +143.98%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, tiny high limb: Collecting 100 samples in estimated 5.0003 s (62M it
bounded random/random_mod, U1024, tiny high limb
                        time:   [71.228 ns 71.322 ns 71.436 ns]
                        change: [+83.316% +83.869% +84.388%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
```

mrdomino:
Current benchmarks show equivocal performance compared with master, slower in some cases, faster in some cases:
```
bounded random/random_mod, U1024
                        time:   [72.961 ns 76.993 ns 81.109 ns]
                        change: [+20.725% +25.815% +31.323%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, small bound: Collecting 100 samples in estimated 5.0001 s (251M iter
bounded random/random_mod, U1024, small bound
                        time:   [14.832 ns 15.687 ns 16.576 ns]
                        change: [−16.001% −12.053% −7.9623%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking bounded random/random_mod, U1024, 512 bit bound low: Collecting 100 samples in estimated 5.0002 s (139
bounded random/random_mod, U1024, 512 bit bound low
                        time:   [42.299 ns 44.756 ns 47.282 ns]
                        change: [−10.546% −6.7606% −2.5299%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking bounded random/random_mod, U1024, 512 bit bound hi: Collecting 100 samples in estimated 5.0001 s (68M
bounded random/random_mod, U1024, 512 bit bound hi
                        time:   [71.057 ns 74.627 ns 78.443 ns]
                        change: [+14.196% +18.859% +24.401%] (p = 0.00 < 0.05)
                        Performance has regressed.
Benchmarking bounded random/random_mod, U1024, tiny high limb: Collecting 100 samples in estimated 5.0001 s (124M i
bounded random/random_mod, U1024, tiny high limb
                        time:   [40.052 ns 40.076 ns 40.104 ns]
                        change: [+2.3526% +2.6806% +3.0255%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) low severe
  3 (3.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
```

mrdomino:
@tarcieri any thoughts on this?

I could see the switch from `ct_lt` being controversial here; and arguably merits renaming this to `random_mod_vartime`.

tarcieri:
Sorry I haven't looked at this, it was a holiday weekend here in the US, and I got a bit lost following your other PR with supposed miscompilation? (did you ever get that into some sort of outside-of-CI reproducible state? If so we should report it upstream)

I feel like this is fundamentally trying to something quite similar to #285, but I find the implementation there a bit more comprehensible.

cc @fjarri

mrdomino:
No worries; miscompilation bug is here:
https://github.com/rust-lang/rust/issues/149522

E.g.:
https://github.com/mrdomino/subtle-repro/actions/runs/19838188507

mrdomino:
> I feel like this is fundamentally trying to something quite similar to #285, but I find the implementation there a bit more comprehensible.

Yes; this implementation is kinda wonky for two reasons:
1. The desire for `random_mod` benchmarks to be fast, which prevents me just going with 7e0166b85dca252c81301269f1d7e14a60178255.
2. The compiler bug, which prevents me using `ct_lt` with the optimized version.

mrdomino:
Yeah actually re-reading #285, I prefer it myself - difference (in terms of values produced / rng state) is that this is (64-bit) word-aligned whereas that is byte-aligned; this wastes more entropy but may provide more opportunities for optimization, whereas that wastes less entropy (but still some; we are not going down to the bit level after all.)

fjarri:
And `Encoding` is now available for all `Uint`, so it can be made simpler

mrdomino:
@fjarri a take 3, based on `Encoding`: #1026.

tarcieri:
Closing in favor of #1026

PR #1026, opened 2025-12-02
BREAKING: Write `random_mod` in terms of new `random_bits`

### LINKED ISSUE #1009: random_mod is not platform-independent

mrdomino:
## Breaking changes

- `Encoding::Repr` is no longer required to implement `Copy`, so consumers of `Encoding::Repr` will need to explicitly call `clone`.

- The numbers produced by both `random_bits` and `random_mod` will generally be different, and calling these functions will leave the RNG in a different state, than before.

## Fixes

- Adds a mitigation for rust-lang/rust#149522, modifying `Word::borrowing_sub` to use `overflowing_sub` instead of `WideWord` casts.

## Summary

This essentially applies https://github.com/RustCrypto/crypto-bigint/pull/285 to `random_bits` as well as `random_mod`. Both functions behave the same way now, with the only difference being that `random_mod` adds rejection sampling; otherwise both will produce the same numbers over the same entropy stream.

Questions of platform dependence are now easy; we do not define these algorithms in terms of sequences of machine words but of bytes. Randomly sampled `Uint`s are now always constructed little-endian over the entropy stream. This does not preclude future machine-specific optimizations, but given how perilous the landscape has been (e.g. #1018), I’ve elected to err in the direction of parsimony rather than performance for this change.

This leverages the existing work making `Uint` implement `Encoding`. It additionally needs `Encoding` on `BoxedUint` to make `RandomMod` and `RandomBits` work there; this is implemented, but requires dropping the `Copy` constraint on `Encoding::Repr`.

Fixes #1009

[0]: https://github.com/rust-lang/rust/issues/149522

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -641,7 +641,6 @@ pub trait Encoding: Sized {
     /// Byte array representation.
     type Repr: AsRef<[u8]>
         + AsMut<[u8]>
-        + Copy
         + Clone
         + Sized
         + for<'a> TryFrom<&'a [u8], Error: core::error::Error>;
--- a/src/uint/boxed/encoding.rs
+++ b/src/uint/boxed/encoding.rs
@@ -1,7 +1,7 @@
 //! Const-friendly decoding operations for [`BoxedUint`].
 
 use super::BoxedUint;
-use crate::{DecodeError, Limb, Word, uint::encoding};
+use crate::{DecodeError, Encoding, Limb, Word, uint::encoding};
 use alloc::{boxed::Box, string::String, vec::Vec};
 use subtle::{Choice, CtOption};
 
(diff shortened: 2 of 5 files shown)


tarcieri:
> Encoding::Repr is no longer required to implement Copy, so consumers of Encoding::Repr will need to explicitly call clone.

That's probably a good change if we want to be able to impl `Encoding` for `BoxedUint` where it would be nice to use `Vec<u8>` for that value

mrdomino:
@tarcieri PTAL - I added a mitigation for the compiler bug that allows us to keep using `ct_lt` for the comparison.

NB. this mitigation could also be applied to any of the other attempts, including the original reverted PR, at your preference; it may also make sense to merge it separately, in which case feel free to cherry-pick the commit onto a new branch or I can do the same.

mrdomino:
#1029 has just the cherry-picked mitigation commit.

tarcieri:
@mrdomino can you rebase now that #1029 is merged?

mrdomino:
Done. Two clean commits, one that adds the test (passing on 64-bit, failing on 32-bit), one that implements the change (and shows the diffs against the test.)

PR #1027, opened 2025-12-04
Add native linux- and darwin-aarch64 tests

mrdomino:
Removes the `test-cross` target for `aarch64-unknown-linux-gnu`, preferring to run this test against GitHub’s native Linux arm runners. Should give a considerable speedup from the combination of not needing to be virtualized and not needing to set up the cross environment.

Also adds a test case against Darwin on `macos-latest` to get another aarch64 platform, and avoids running an empty `deps` action if there is none.

### DIFF
--- a/.github/workflows/crypto-bigint.yml
+++ b/.github/workflows/crypto-bigint.yml
@@ -58,7 +58,6 @@ jobs:
       - run: cargo build --all-features --benches
 
   test:
-    runs-on: ubuntu-latest
     strategy:
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


mrdomino:
Yeah, that's probably still dominated by the powerpc cross test. I wonder if there's a faster way to get confirmation that it works against big-endian systems.

tarcieri:
Some of the longest tests time-wise are actually the `i686` tests, because a considerable amount of time is spent installing `gcc-multilib`, which is a general problem across @RustCrypto.

It sure would be nice if it were pre-installed on the GitHub Actions runners

PR #1028, opened 2025-12-05
Add LCM for Uint

Fethbita:
I only added the LCM for `Uint` for now, I think it should not be difficult to add it to other types as well. I have never used the Int type so if that is different, might take some time.

### DIFF
--- a/src/uint/lcm.rs
+++ b/src/uint/lcm.rs
@@ -4,7 +4,7 @@ use crate::{ConcatMixed, NonZero, Uint};
 
 impl<const LIMBS: usize> Uint<LIMBS> {
     /// Compute the least common multiple of `self` and `rhs`.
-    pub const fn lcm_uint<const WIDE_LIMBS: usize>(&self, rhs: &Self) -> Uint<WIDE_LIMBS>
+    pub const fn lcm<const WIDE_LIMBS: usize>(&self, rhs: &Self) -> Uint<WIDE_LIMBS>
     where
         Self: ConcatMixed<Uint<LIMBS>, MixedOutput = Uint<WIDE_LIMBS>>,
     {
@@ -31,7 +31,7 @@ mod tests {
         ) where
             Uint<LIMBS>: ConcatMixed<Uint<LIMBS>, MixedOutput = Uint<WIDE_LIMBS>>,
         {
-            assert_eq!(lhs.lcm_uint(&rhs), target);
+            assert_eq!(lhs.lcm(&rhs), target);
         }
 
         fn run_tests<const LIMBS: usize, const WIDE_LIMBS: usize>()
--- a/tests/uint.rs
+++ b/tests/uint.rs
@@ -4,7 +4,7 @@ mod common;
 
 use common::to_biguint;
 use crypto_bigint::{
-    Encoding, Gcd, Limb, NonZero, Odd, U256, U4096, U8192, Uint, Word,
+    Encoding, Gcd, Limb, NonZero, Odd, U256, U512, U4096, U8192, Uint, Word,
     modular::{MontyForm, MontyParams},
 };
 use num_bigint::BigUint;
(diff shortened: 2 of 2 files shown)


src/uint/lcm.rs
```diff
@@ -0,0 +1,75 @@
...
+            lhs: Uint<LIMBS>,
+            rhs: Uint<LIMBS>,
+            target: Uint<WIDE_LIMBS>,
+        ) where
+            Uint<LIMBS>: ConcatMixed<Uint<LIMBS>, MixedOutput = Uint<WIDE_LIMBS>>,
+        {
+            assert_eq!(lhs.lcm_uint(&rhs), target);
+        }
+
+        fn run_tests<const LIMBS: usize, const WIDE_LIMBS: usize>()
```
fjarri, line 37: What's our policy on proptests? Seems like it would be useful to have one here.

tarcieri: I'm definitely a fan of proptests

Fethbita: By proptest, do you mean this [proptest](https://github.com/proptest-rs/proptest)? I have never used it before, so what kind of tests would you like here? Generate 2 primes (it seems [difficult](https://proptest-rs.github.io/proptest/proptest/tutorial/filtering.html) to get it to generate primes) and calculate that the LCM output equals to multiple of the inputs? Or would it be used only for detecting whether the function crashes with any kind of input?

fjarri: I think there are two types of tests possible here. First, with primes; but you don't need filtering for those, just use random bytes as a seed for a random prime. Second, edge cases, for which you don't need primes, just random numbers. `num_bigint`'s `lcm` can be used as the reference.

tarcieri: @Fethbita you can look here for some example protests: https://github.com/RustCrypto/crypto-bigint/blob/master/tests/uint.rs

I guess I would be okay with not having them for the initial PR, but it would definitely be good to have them as a followup, whoever authors it. It's something I can look at.

Truly random primes seem like they might need a circular dependency on `crypto-primes` which is doable (it can be added as a `dev-dependency`, I think) but a little tricky.

Another option is to write the test with a canned set of primes and select from them randomly, though if the solution space is small enough it could just be brute forced every test run and kind of defeats the point of proptest.

src/uint/lcm.rs
```diff
@@ -0,0 +1,75 @@
+//! This module implements Least common multiple (LCM) for [`Uint`].
+
+use crate::{ConcatMixed, NonZero, Uint};
+
+impl<const LIMBS: usize> Uint<LIMBS> {
+    /// Compute the least common multiple of `self` and `rhs`.
+    pub const fn lcm_uint<const WIDE_LIMBS: usize>(&self, rhs: &Self) -> Uint<WIDE_LIMBS>
```
tarcieri: I'm not sure I like this `_uint` naming scheme, which I guess was introduced in #923 

It feels to me like `_uint` should be the "default" for `Self`/`Uint`, so this is just `fn lcm`

cc @andrewwhitehead

src/uint/lcm.rs
Fethbita: I added a single test for comparing with num_bigint, using random values.

src/uint/lcm.rs
Fethbita: Right, removed the `_uint`.

andrewwhitehead: It's kinda the same issue here as in the other PR. If there's an LCM trait, then the inherent method can clobber the trait method, making combinations like Uint/Int more difficult to use in non-generic contexts.

tarcieri: @andrewwhitehead that's fine so long as they have the same type signature. You can always call the inherent method with UFCS.

If they have different type signatures, that's bad.

tarcieri (approved):

PR #1030, opened 2025-12-12
Rename `random_mod` ↦ `random_mod_vartime`

mrdomino:
Rejection sampling means that `random_mod` is inherently variable-time, and the current name violates the contract in the `README`:

> All functions contained in the crate are designed to execute in
> constant time unless explicitly specified otherwise (via a `*_vartime`
> name suffix).

This renames `random_mod` to `random_mod_vartime`, adding in deprecated aliases at the old names for backwards-compatibility. It preserves the `ConstantTimeLess` functionality for good measure and to not needlessly increase the amount of timing information leaked.

Of course, `random_mod` leaks a pretty small amount of timing information; by my reckoning, at worst an average 2 bits per call, and approaching 0 as the modulus approaches `2^n-1`. So this is a bit of an edge case; but I think the `vartime` naming here reduces surprise.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -466,8 +466,8 @@ pub trait RandomMod: Sized + Zero {
     /// example, it implements `CryptoRng`), then this is guaranteed not to
     /// leak anything about the output value aside from it being less than
     /// `modulus`.
-    fn random_mod<R: RngCore + ?Sized>(rng: &mut R, modulus: &NonZero<Self>) -> Self {
-        let Ok(out) = Self::try_random_mod(rng, modulus);
+    fn random_mod_vartime<R: RngCore + ?Sized>(rng: &mut R, modulus: &NonZero<Self>) -> Self {
+        let Ok(out) = Self::try_random_mod_vartime(rng, modulus);
         out
     }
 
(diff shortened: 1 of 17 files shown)


src/traits.rs
```diff
@@ -480,10 +480,41 @@ pub trait RandomMod: Sized + Zero {
...
+    /// example, it implements `CryptoRng`), then this is guaranteed not to
+    /// leak anything about the output value aside from it being less than
+    /// `modulus`.
+    #[deprecated(since = "0.7.0", note = "please use `try_random_mod_vartime` instead")]
+    fn try_random_mod<R: TryRngCore + ?Sized>(
+        rng: &mut R,
+        modulus: &NonZero<Self>,
+    ) -> Result<Self, R::Error> {
+        Self::try_random_mod_vartime(rng, modulus)
+    }
```
mrdomino, line 517: **NB**. these are the deprecated aliases added (which might otherwise get lost in the noise of the mechanical rename diff.)

mrdomino:
Rereading the argument about cryptographic security of the RNG, the case for the old name being okay seems stronger than I initially thought. I do think it’s a bit of a funny edge case; I could go either way on this change, feel free to reject if it doesn’t seem worthwhile.

tarcieri (approved):
It should be possible to have a fallible version that runs in constant time, takes a select (possibly const generic) number of samples, selects the first one that fits the criteria in constant-time among all of the samples, and returns an error if none of them fit the criteria.

For now though, I agree, this should be named `*_vartime`.

tarcieri:
Not sure what's going on with codecov. I think it's mad you're adding methods which aren't being tested, even if they're deprecated.

PR #1040, opened 2025-12-26
Replace `ConstCtOption` with `ctutils::CtOption`

tarcieri:
Analogous to what #1035 did for `ConstChoice`, this extracts the necessary methods onto `ctutils::CtOption` then replaces the type wholesale using the implementation in `ctutils`.

There were a few tricky parts, namely we had type-specific implementations of `expect` and `unwrap_or`.

It is actually possible to write a generic `const fn` implementation of `CtOption::expect`, avoiding the problems with destructors and the unstable `Destruct` trait, by bounding on `T: Copy` and returning a copy of the inner value. This is implemented as `CtOption::expect_copied` so we can still support `expect` for non-`Copy` types.

Unfortunately there's no analogous generic `const fn` solution for `CtOption::unwrap_or`, as this needs a constant-time selector function and even function pointers can't be used from a `const fn`. So this does the only thing we can do to replace it with some level of abstraction: a `ctutils::unwrap_or!` macro.

See also: RustCrypto/utils#1274 which extracts some needed methods onto `ctutils::CtOption`.

cc @andrewwhitehead


tarcieri:
As a bit of a plan of where to go from here: I'd like to get rid of `ConstChoice`/`ConstCtOption` and entirely replace them with `ctutils::{Choice, CtOption}`.

For now we can impl both the `subtle` traits and the new `ctutils` traits.

I'd like to replace anything which isn't a `subtle` trait impl which is using `subtle::{Choice, CtOption}` with `ctutils::{Choice, CtOption}`. Then `subtle` can potentially be made an optional dependency.

tarcieri:
This will also make it possible to get rid of the `ConstantTimeSelect` trait defined in `crypto-bigint`, which is there to work around the `Copy` bound on `subtle::ConditionallySelectable`

tarcieri:
Benchmarks are, as usual, all over the place. I've rerun them a few times to see if there's anything that sticks out as consistently better or worse, and it just seems like especially for operations that are measured in `ns` there's just too much variability. Fortunately most of that is in the 3%-5% range (faster or slower).

I've seen a few things that feel like flukes (e.g. `I256::wrapping_mul` 20% faster, `I4096::concatenating_mul`  8% slower), but I think I'm looking at noise rather than signal there.

All that said, I'm not seeing anything particularly concerning.

tarcieri:
Given that, I'm going to go ahead and merge, and press forward trying to get rid of the `Choice`/`ConstChoice` and `CtOption`/`ConstCtOption` distinction

PR #1043, opened 2025-12-27
Replace `ConstantTimeSelect` with `ctutils::CtSelect`

tarcieri:
The `ConstantTimeSelect` trait was a workaround for having a trait-based API that works for `Boxed*` and non-boxed types, since we can't impl `subtle::ConditionallySelectable` on `Boxed*` types since it has a `Copy` bound.

`ctutils::CtSelect` accomplishes the same goals by having only a `Sized` bound (since we need to be able to return selected values).

This commit completely replaces the `ConstantTimeSelect` trait with `ctutils::CtSelect`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -193,7 +193,7 @@ pub use crate::{
 };
 
 // TODO(tarcieri): get rid of `Const*` prefix
-pub use ctutils::{Choice as ConstChoice, CtOption as ConstCtOption};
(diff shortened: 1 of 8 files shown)


tarcieri:
As far as boxed ops go, since that's mainly what this PR impacts, I'm actually seeing a bunch of performance improvements for the most part, but all within a range where they may be spurious. So far there's only one 13% regression (in `neg` somehow? probably spurious)

PR #1048, opened 2025-12-28
Replace `ConstantTimeEq` with `CtEq` in trait bounds

tarcieri:
Changes all of the traits in `traits.rs`, most notably the `Integer`, `Zero`, and `One` traits, to use `ctutils::CtEq` instead of `subtle::ConstantTimeEq`, as part of a larger overall migration to `ctutils` (see also #1040, #1043)


tarcieri:
Codecov doesn't like the uncovered lines but the tests are otherwise passing.

I plan on doing some separate passes soon to increase testing coverage.

PR #1052, opened 2025-12-28
Use `ConstCtOption` for checked operations

tarcieri:
Uses it both as the return type for ops described by `Checked*` traits as well as the inner type for the `Checked` wrapper.

This seemingly uncovered some issues with the division implementation, including the possibility that checked division was somehow short-circuited when using the previous implementation of the `NonZero` type, as with this change it's now invoking a function that panics with a zero divisor, when instead it still needs to always call the same function and get a result without panicking, even if the divisor is zero.

To solve the general problem, `NonZero::new` now also bounds on `One` and `CtSelect` (in addition to `Zero`), and automatically uses `One::one_like` as a placeholder value in the `ConstCtOption` it returns in the event the provided value is equal to zero. That way we never construct an invalid `NonZero` containing an inner `0` value.

### DIFF
--- a/src/checked.rs
+++ b/src/checked.rs
@@ -1,29 +1,30 @@
 //! Checked arithmetic.
 
-use crate::{CheckedAdd, CheckedDiv, CheckedMul, CheckedSub, ConstChoice, CtEq, CtSelect};
+use crate::{
+    CheckedAdd, CheckedDiv, CheckedMul, CheckedSub, ConstChoice, ConstCtOption, CtEq, CtSelect,
+};
 use core::ops::{Add, Div, Mul, Sub};
-use subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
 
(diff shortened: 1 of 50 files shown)


tarcieri:
@andrewwhitehead @fjarri I'm a little confused how the tests were passing before, unless the `CtOption` returned from `NonZero::new` somehow wasn't invoking its combinators correctly. With this change (but without substituting 1 for `NonZero(0)` values) it was blowing up inside `UintRef::div_rem`, which is advertised as panicking if given a zero divisor, so that's to be expected, but it seems like it may not have even been getting called under the previous implementation?

Sidebar: it also seems like `Odd` should get the same treatment, i.e. when the `None` equivalent of `CtOption` is returned from `Odd:new`, it should contain `Odd(One::one_like(...))` as a placeholder, so we never instantiate an invalid `Odd`.

PR #1071, opened 2026-01-04
Fix `Default` impl for `Odd`

### LINKED ISSUE #1070: Default impl for `Odd` is the default of `T` (zero)
Currently `struct Odd<T>(T)` derives `Default`, meaning that it will return a zero integer, which is obviously not odd :)
the following code passes:
```rust
use crypto_bigint::{NonZero, Odd, U256};

fn main() {
    let odd = Odd::<U256>::default();
    assert!(odd.is_zero().to_bool());
    let non_zero: &NonZero<U256> = odd.as_nz_ref();
    assert!(non_zero.is_zero().to_bool());
}

```

I think it should either be removed, or modified to return `ONE` just like the `NonZero` wrapper does: https://github.com/RustCrypto/crypto-bigint/blob/613a471aa778b03cd00f49bf9670be25789a0a94/src/non_zero.rs#L329

tarcieri:
It was previously derived, which would use the `Default` impl of the underlying `T` which, in almost every case, will return zero, violating the invariant of the type.

This adds a custom `Default` impl similar to the one on `NonZero` which is bounded on `One` trait, and returns `Odd(T::one())`.

Fixes #1070

### DIFF
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -322,11 +322,11 @@ where
 
 impl<T> Default for NonZero<T>
 where
-    T: Constants,
+    T: One,
 {
     #[inline]
     fn default() -> Self {
-        Self(T::ONE)
+        Self(T::one())
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/non_zero.rs
```diff
@@ -322,11 +322,11 @@ where
 
 impl<T> Default for NonZero<T>
 where
-    T: Constants,
+    T: One,
```
tarcieri, line 325: Went ahead and relaxed this bound considerably too

PR #1073, opened 2026-01-04
Rename `Uint::gcd_uint(_vartime)` => `::gcd(_vartime)`

tarcieri:
In this case `self` and `rhs` are both `Uint`, so there's no need for a `*_uint` to disambiguate the argument type.

This overlaps with the trait impl but has the same signature aside from being a `const fn`, which is a pattern we use throughout the library.

Using UFCS-ish syntax it's possible to disambiguate, e.g. `Uint::gcd(a, b)` will prefer the inherent impl, whereas `Gcd::gcd(a, b)` will use the trait.

### DIFF
--- a/benches/uint.rs
+++ b/benches/uint.rs
@@ -4,7 +4,7 @@ use criterion::{
     BatchSize, BenchmarkGroup, BenchmarkId, Criterion, criterion_group, criterion_main,
 };
 use crypto_bigint::{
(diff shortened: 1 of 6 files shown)


tarcieri:
cc @andrewwhitehead

andrewwhitehead:
Sure, looks good to me.

tarcieri:
Notably this got rid of a bunch of now unused `Gcd` trait imports, since it's using the inherent method instead

Issue #1082, opened 2026-01-06
Square root trait and implementation cleanups


andrewwhitehead:
Some potential cleanups for the traits and implementation:

- [x] Add a `CheckedSquareRoot` trait which could be implemented by `Int` (returning none for negative inputs). This could be a supertrait of SquareRoot.
- [x] Rename the `SquareRoot` trait for clarity, perhaps `UintSquareRoot` or `FloorSquareRoot`.
- ~~[ ] Implement the square root trait(s) for Montgomery forms (this makes `FloorSquareRoot` a less desirable name)~~.
- [x] Rename `Uint::sqrt/sqrt_vartime` to `isqrt` or `sqrt_unchecked` to reflect the inexact nature of the result (and corresponding trait updates to match).
- [ ] Implement square root traits and methods for `OddUint` for consistency. This would require changing the return type, but it seems reasonable to add an Output type to the trait(s).
- [ ] Support nth roots based on algorithm 1.14 from the same source (Brent & Zimmermann, Modern Computer Arithmetic, v0.5.9). Should this be a separate trait?

andrewwhitehead:
I think it makes sense to make `CheckedSquareRoot` the primary trait (as in #1083). The other option here would be to maintain `SquareRoot` but make the result fallible. Montgomery forms could implement this, although the method is also provided by `ff::Field`.

`Uint` could implement a `FloorSquareRoot` trait (my current preferred naming) with `floor_sqrt`, `floor_sqrt_vartime` methods. It doesn't seem like Montgomery forms have a need for a method that doesn't return an Option/CtOption. `IntegerSquareRoot` with `isqrt/isqrt_vartime` would be an option somewhat aligned with the standard library, but it seems less precise (implies the result could be negative? and doesn't specify the rounding).

I have a vartime implementation of the nth root calculation. It's a bit unclear how to implement that in constant time because a) does it need to be constant time in the exponent? and b) we don't know the number of rounds to perform, although I think it might be the same.

PR #1083, opened 2026-01-06
Add CheckedSquareRoot trait, related cleanups

andrewwhitehead:
Related to #1082 

Adds a `CheckedSquareRoot` trait and renames the `SquareRoot` to `FloorSquareRoot`, extending it with an Output type. Associated methods are renamed, `sqrt -> floor_sqrt` where appropriate.

Implements `CheckedSquareRoot` for `Uint`, `NonZeroUint` and `Int`.

### DIFF
--- a/src/uint/sqrt.rs
+++ b/src/uint/sqrt.rs
@@ -6,12 +6,30 @@ impl<const LIMBS: usize> Uint<LIMBS> {
     /// Computes `floor(√(self))` in constant time.
     ///
     /// Callers can check if `self` is a square by squaring the result.
+    #[deprecated(since = "0.7.0", note = "please use `floor_sqrt` instead")]
+    pub const fn sqrt(&self) -> Self {
+        self.floor_sqrt()
+    }
+
+    /// Computes `floor(√(self))` in constant time.
+    ///
+    /// Callers can check if `self` is a square by squaring the result.
     pub const fn floor_sqrt(&self) -> Self {
         let self_is_nz = self.is_nonzero();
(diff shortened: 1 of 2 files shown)


tarcieri:
I think my personal preference would be for the default `sqrt` to be checked, which seems like the safest option, and either have `isqrt` or `sqrt_unchecked` for the alternative

andrewwhitehead:
@tarcieri Yes that makes sense, I figured naming would be adjusted later.

tarcieri:
I think the naming is important because of the existing `SquareRoot` trait, which should get declared checked (or unchecked if you really want to go that way), then its existing impls need to be audited and any which aren't checked (or unchecked) as appropriate need to get moved to the new trait

andrewwhitehead:
@tarcieri I agree the naming is important, but I think these changes are mostly independent of that. I'll add comments to the open issue on the naming.

src/uint/sqrt.rs
```diff
@@ -1,87 +1,75 @@
...
+use crate::{CheckedSquareRoot, CtEq, CtOption, FloorSquareRoot, Limb, NonZero, Uint};
 
 impl<const LIMBS: usize> Uint<LIMBS> {
-    /// Computes √(`self`) in constant time.
+    /// Computes `floor(√(self))` in constant time.
     ///
-    /// Callers can check if `self` is a square by squaring the result
-    pub const fn sqrt(&self) -> Self {
+    /// Callers can check if `self` is a square by squaring the result.
+    pub const fn floor_sqrt(&self) -> Self {
```
tarcieri, line 17: If we go this route it definitely needs to be with a legacy old method name and a deprecation

PR #1086, opened 2026-01-12
`NonZero` and `Odd` encoding tests

tarcieri:
...and a few others. This actually caught a bug!

Adds tests for the following:
- `NonZero::{from_be_bytes, from_le_bytes}`
- `NonZero::{from_be_hex, from_le_hex}`
- `NonZero::one`
- `Odd::{from_be_hex, from_le_hex}`

Additionally sprinkles `#[track_caller]` (which was helpful for debugging the aforementioned bug) and `#[inline]` in a few places that make sense.

### DIFF
--- a/src/odd.rs
+++ b/src/odd.rs
@@ -105,6 +105,7 @@ impl<const LIMBS: usize> Odd<Uint<LIMBS>> {
     /// Create a new [`Odd<Uint<LIMBS>>`] from the provided big endian hex string.
     ///
     /// Panics if the hex is malformed or not zero-padded accordingly for the size, or if the value is even.
+    #[track_caller]
     pub const fn from_be_hex(hex: &str) -> Self {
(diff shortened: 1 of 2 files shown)


src/odd.rs
```diff
@@ -114,8 +115,9 @@ impl<const LIMBS: usize> Odd<Uint<LIMBS>> {
     /// Create a new [`Odd<Uint<LIMBS>>`] from the provided little endian hex string.
     ///
     /// Panics if the hex is malformed or not zero-padded accordingly for the size, or if the value is even.
+    #[track_caller]
     pub const fn from_le_hex(hex: &str) -> Self {
-        let uint = Uint::<LIMBS>::from_be_hex(hex);
+        let uint = Uint::<LIMBS>::from_le_hex(hex);
```
tarcieri, line 120: Here's the bug. Oops.

PR #1087, opened 2026-01-12
Have `BoxedMontyForm::new` auto-`Clone` its params

tarcieri:
Changes the `params` argument to `&BoxedMontyParams` and automatically invokes `Clone` on it, since cloning the inner `Arc` is cheap.

We were otherwise leaving this to the caller to do which is annoying, especially since it can be automatic.

However, changing this to a reference effectively means we also need to change the `Monty::new` trait method to take a reference, which means we also need to change `MontyForm::new` to keep everything consistent. That's not really a problem as `MontyParams` are `Copy`.

This does wind up being quite a bit of a breaking change, but should be worth it to get rid of the need for callers to clone params.

### DIFF
--- a/benches/boxed_monty.rs
+++ b/benches/boxed_monty.rs
@@ -26,11 +26,11 @@ fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
             || {
                 let a = BoxedMontyForm::new(
                     BoxedUint::random_mod_vartime(&mut rng, params.modulus().as_nz_ref()),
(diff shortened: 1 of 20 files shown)


tarcieri:
Heads up @fjarri, this is a breaking change I suspect will impact `crypto-primes`.

cc @andrewwhitehead

tarcieri:
Definitely like this a *lot* better.

I also have a PR ready to open on `crypto-primes`, where it also seems cleaner to me, IMO.

PR #1090, opened 2026-01-12
Add `BoxedUint::concatenating_add`

tarcieri:
Addition method which returns a result concatenated with one additional overflow limb which would be returned if `carrying_add` were called with the same operands.

### DIFF
--- a/src/uint/boxed/add.rs
+++ b/src/uint/boxed/add.rs
@@ -1,9 +1,10 @@
 //! [`BoxedUint`] addition operations.
 
 use crate::{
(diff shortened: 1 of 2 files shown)


tarcieri:
A bit torn on this name since we're moving from `widening_mul` to `concatenating_mul`.

I guess I could call it `concatenating_add` where I guess it's concatenating an extra limb onto the result.

tarcieri:
Well, I guess I convinced myself, and renamed it to `concatenating_add` for consistency with `concatenating_mul`.

(FWIW I need this for computing SRP exponents)

tarcieri:
I'm wondering if this should perhaps be the default behavior for the `Add` trait or if that's too weird.

Not going to do it for the time being.

PR #1092, opened 2026-01-13
Use `core` convention for mixed-sign ops

tarcieri:
The convention is as follows:

- `{unsigned}::*_signed` for ops that take `{signed}` RHS operand
- `{signed}::*_unsigned` for ops that take `{unsigned}` RHS operand

Previously we used `*_int` and `*_uint` respectively.

The more we align with core, the more operations that people expect to work will "just work".

cc @andrewwhitehead


tarcieri:
Confirmed this change has no effect on `crypto-primes`... yay! 🎉

Issue #1046, opened 2025-12-27
Vartime methods should return `Option` instead of `CtOption`


tarcieri:
We have several methods that are advertised as variable-time which are returning `CtOption` instead of `Option`.

This is a bit suboptimal: `CtOption` eagerly evaluates its combinators, whereas `Option` uses lazy evaluation. It would generally be more efficient to use `Option`.

Perhaps a reason why this was the case originally was trouble converting from `ConstCtOption` to `Option` inside a `const fn` in the past, but there's now `ctutils::CtOption::into_option_copied` which can be used as a `const fn`, in cases where there's some shared code between the constant-time and non-constant-time versions of a function.

cc @andrewwhitehead @fjarri

andrewwhitehead:
It does depend on why the method is vartime, in some cases it might still be constant time for `self` and it would make sense to return a `CtOption`.

tarcieri:
Aah yes, that's true

andrewwhitehead:
It might be cleaner and more ergonomic to return `Checked<T>` from the constant-time methods. With that, I think you could mostly hide `CtOption` from the public API. `Checked` could use a few helper methods in that case, like `into_option` and `expect`/`expect_copied` and maybe `unwrap`. For the vartime cases it seems like `Option<T>` might still be the best option, unless another `CheckedVartime` type was added.

PR #1097, opened 2026-01-13
Make vartime inversion return an `Option` instead of `CtOption`

andrewwhitehead:
Related to #1046

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -904,11 +904,11 @@ pub trait Invert {
     type Output;
 
     /// Computes the inverse.
-    fn invert(&self) -> Self::Output;
+    fn invert(&self) -> CtOption<Self::Output>;
 
     /// Computes the inverse in variable-time.
-    fn invert_vartime(&self) -> Self::Output {
-        self.invert()
+    fn invert_vartime(&self) -> Option<Self::Output> {
+        self.invert().into_option()
     }
 }
 
--- a/benches/boxed_monty.rs
+++ b/benches/boxed_monty.rs
@@ -96,6 +96,19 @@ fn bench_montgomery_ops<M: Measurement>(group: &mut BenchmarkGroup<'_, M>) {
         )
     });
 
+    group.bench_function(format!("invert_vartime, {UINT_BITS}-bit"), |b| {
+        b.iter_batched(
+            || {
+                BoxedMontyForm::new(
+                    BoxedUint::random_mod_vartime(&mut rng, params.modulus().as_nz_ref()),
+                    &params,
+                )
+            },
+            |x| black_box(x).invert_vartime(),
+            BatchSize::SmallInput,
+        )
+    });
+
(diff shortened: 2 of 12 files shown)


src/traits.rs
```diff
@@ -904,11 +904,11 @@ pub trait Invert {
...
     /// Computes the inverse.
-    fn invert(&self) -> Self::Output;
+    fn invert(&self) -> CtOption<Self::Output>;
 
     /// Computes the inverse in variable-time.
-    fn invert_vartime(&self) -> Self::Output {
-        self.invert()
+    fn invert_vartime(&self) -> Option<Self::Output> {
+        self.invert().into_option()
     }
```
tarcieri, line 912: Noticed this change a little too late.

So we unified the `Invert` traits used by `crypto-bigint` and `elliptic-curve`, and in the latter the reason this trait doesn't explicitly specify `CtOption` as the return type but rather relies on the `Output` type to specify that is so we can express infallible inversions for non-zero elements of prime fields. I guess `crypto-bigint` doesn't make use of this trick itself, but this change prevents it from being possible.

There's also the problem that `elliptic-curve` is still using `subtle` due to its dependency on `ff`/`group` traits, and. I need to [either upstream my work from `ctutils` into `subtle` or convince the authors of `ff`/`group` to migrate from `subtle` to `ctutils`](https://github.com/zkcrypto/ff/issues/148), but until then I need to use `Invert` with `subtle::CtOption` instead of `ctutils::CtOption`.

So I'm going to need to revert this PR and cut another release until we can figure out what to do here. Perhaps split `Invert` and `InvertVartime` into separate traits so they can use different `Output` types?

andrewwhitehead:
@tarcieri  That's unfortunate. I don't feel like infallible inversion is the same operation, though. Maybe this one should be renamed `CheckedInvert`.

andrewwhitehead:
Having separate `CheckedInvert` and `Invert` traits would also line up nicely with the new `CheckedSquareRoot`/`FloorSquareRoot` traits.

tarcieri:
@andrewwhitehead we can do that I guess, though I can't use `CheckedInvert` in `elliptic-curve` then due to the `subtle`/`ctutils` incompatibility 😢

tarcieri:
This seems like a different case than `CheckedSquareRoot`/`FloorSquareRoot` though... it makes sense for a single type to impl both of those.

But for inversions, there's only going to be one way to do them for a given type: either the type has an invariant that means the inversion will always succeed, or there's going to be a failure case you have to handle. No type is ever going to impl both traits.

andrewwhitehead:
Anything that can implement `Invert` can implement `CheckedInvert` though, which might provide more opportunities for generic implementations... `FloorSquareRoot` currently requires `CheckedSquareRoot` as well.

tarcieri:
What's a case where it's actually useful to impl both traits?

PR #1104, opened 2026-01-17
Consolidate `ct` modules

tarcieri:
Consolidates modules containing all constant-time functionality for each numeric type, including: `Int`, `Uint`, `BoxedUint`, `ConstMontyForm`, `MontyForm`, and `BoxedMontyForm`.

It's the new home for any `ctutils::Ct*` impls or `subtle` trait impls.

This makes auditing the constant-time operations easier, and is also in anticipation of having to add additional `CtAssign` impls to all types when we bump `ctutils`.


tarcieri:
I think we may have even had it factored this way at one point in the past 😅 

Without this structure it's a bit unclear where to put the `CtAssign` impls besides a new `assign` module and when I was thinking about that I really wished all the constant-time impls were in a single module.

PR #1117, opened 2026-01-19
`NonZero<UintRef>` improvements

tarcieri:
- Switch from `NonZero<&UintRef>` to `&NonZero<UintRef>` (also `mut`)
- Introduce `UintRef::{as_nz_unchecked, as_mut_nz_unchecked}` as `unsafe fn` wrappers for the pointer cast to `NonZero`
- Add `NonZeroUint::as(_mut)_nz` and `NonZeroBoxedUint::as(_mut)_nz` as an infallible way of obtaining `&NonZero<UintRef>`

### DIFF
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -2,7 +2,7 @@
 
 use crate::{
     Bounded, Choice, ConstOne, Constants, CtAssign, CtEq, CtOption, CtSelect, Encoding, Int, Limb,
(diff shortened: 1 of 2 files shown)


tarcieri:
@andrewwhitehead perhaps this is enough for `UintRef` to have `NonZero` divisors and therefore not panic?

andrewwhitehead:
Nice, I started adding something like that recently too.

PR #1118, opened 2026-01-19
Add `pow_vartime` for `MontyForm`, `ConstMontyForm`

andrewwhitehead:
This could be further optimized for small exponents where it doesn't make sense to use a precomputed table. That's probably true for the non-vartime version as well.

I think this also enables a vartime multi-exponentiation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,13 +82,13 @@ checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "chacha20"
-version = "0.10.0-rc.6"
+version = "0.10.0-rc.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f895fb33c1ad22da4bc79d37c0bddff8aee2ba4575705345eb73b8ffbc386074"
+checksum = "26f659b53b7f4e776e0cb370508f21d62a22349009195b1089de5f9b24afba2f"
 dependencies = [
  "cfg-if",
  "cpufeatures",
- "rand_core 0.10.0-rc-3",
+ "rand_core 0.10.0-rc-4",
 ]
 
 [[package]]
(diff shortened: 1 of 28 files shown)


tarcieri:
Re: the non-vartime version, I was going to look at wiring up the generic AMM one today

andrewwhitehead:
@tarcieri I started doing that as well, but it wouldn't be `const` and `AmmMultiplier` isn't currently implemented for those types. It might be useful to have a generic `square_repeat` (sqn) method that only reduced the value after repeated rounds which could be used here. Last I checked there seemed to be little to no benefit to using the AMM multiplication for these types though.

tarcieri:
> it wouldn't be const and AmmMultiplier isn't currently implemented for those types.

Yeah wiring up `AmmMultiplier` is the missing piece really. `const` doesn't matter for my use cases (`dsa`, `rsa`, `srp`).

> It might be useful to have a generic square_repeat (sqn) method that only reduced the value after repeated rounds which could be used here. 

That would be generally useful, even a vartime version:

https://github.com/RustCrypto/elliptic-curves/blob/8006f06/primefield/src/monty.rs#L387-L401

> Last I checked there seemed to be little to no benefit to using the AMM multiplication for these types though.

I'd be curious to see what the benchmarks have to say, particularly in a real-world context (e.g. `srp` could easily be changed from `BoxedMontyForm` to `ConstMontyForm`, and already has `ConstMontyParams` for all the groups). I'm not really familiar with the other implementation, but AFAIK AMM should have the fewest reductions?

andrewwhitehead:
I'm adding this mainly to support the modular sqrt which is currently also `const`.

Each AMM multiplication step does a conditional subtraction of the modulus (when there is an overflow), while the full multiplication step does a conditional subtraction followed by a conditional addition. When there's a small number of limbs this doesn't have much impact. It's worth benchmarking though.

tarcieri:
All the use cases I mentioned have a large number of limbs

tarcieri (approved):

tarcieri:
Tried to merge the conflicts so I can rebase another branch I have which is going to make a lot of cosmetic changes (#1129)

PR #1121, opened 2026-01-19
Remove `UintRef::div*` from the public API

tarcieri:
It would be nice to at least attempt to use `NonZeroUintRef` as the divisor before we make these public, as it could eliminate a potential panic condition.

This changes them from `pub` to `pub(crate)` and leaves a TODO comment to attempt the `NonZeroUintRef` change first before making them public again.

### DIFF
--- a/src/uint/ref_type.rs
+++ b/src/uint/ref_type.rs
@@ -124,7 +124,6 @@ impl UintRef {
     }
 
     /// Conditionally assign all of the limbs to the maximum.
-    #[cfg(feature = "alloc")]
     #[inline]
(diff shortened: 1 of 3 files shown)


andrewwhitehead:
@tarcieri  I don't think it's possible to use `NonZeroUintRef`s in the division methods. In `div_rem` for example, it leaves the quotient in `rhs` but that value could be zero. In fact it seems problematic that you can get a mutable reference to the value in a `NonZero` or `Odd`.

tarcieri:
Yeah crap, you're right, I added some methods to do that but those are of course problematic because with a mutable reference you can violate the type's invariant. I'll do a followup to remove them

PR #1124, opened 2026-01-19
Revert "Make vartime inversion return an `Option` instead of `CtOption`"

tarcieri:
This reverts commit b9ba08f54e9ad95f6f353ed114ce6e25ae6da621 (#1097).

This changed `Invert::invert` trait method to use `ctutils::CtOption` explicitly.

`elliptic-curve` uses this `Invert` trait to express its inversions, including the ability to fully customize `Invert::Output` to express infallible inversions of `NonZeroScalar` which simply return `Self`.

It's also still using `subtle::CtOption` to handle the fallible cases as we're still using `subtle` in the public API until a decision is made upstream in `ff`/`group`:

https://github.com/zkcrypto/ff/issues/148

I also think we should still potentially leave the door open to supporting such infallible inversions in `crypto-bigint`, for e.g. `NonZero<ConstMontyForm<MOD, _>>` where `MOD: ConstMontyParams + PrimeField` or something like that.

### DIFF
--- a/benches/uint.rs
+++ b/benches/uint.rs
@@ -801,7 +801,7 @@ fn bench_invert_mod(c: &mut Criterion) {
                 loop {
                     let x = U256::random_from_rng(&mut rng);
                     let inv_x = x.invert_odd_mod_vartime(&m);
-                    if inv_x.is_some() {
+                    if inv_x.is_some().into() {
                         break (x, m);
                     }
                 }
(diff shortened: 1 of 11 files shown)


tarcieri:
As I expressed on #1097, maybe we can split out an `InvertVartime` trait that can have its own separate `Output` type from `Invert`. Vartime inversions of non-zero prime fields could also be infallible and *not* return an `Option`, or we can use `Option` by convention (and in bounds) for the fallible cases.

cc @andrewwhitehead

PR #1131, opened 2026-01-20
Impl `AmmMultiplier` for `DynMontyMultiplier`

tarcieri:
- Extracts the `almost_montgomery_mul` function into `modular::mul` and changes its arguments to `UintRef`
- Adds the `AmmMultiplier` impl to `DynMontyMultiplier` which allocates the product buffer on the stack and calls into `almost_montgomery_mul`
- Wires everything up as `(Const)MontyForm::pow_amm(_bounded_exp)` which is an additive change that leaves the existing modpow implementations (which are `const fn`, work on mixed operand sizes, and support multi-exp) as they are
- Adds proptests that everything works for both `ConstMontyForm` and `MontyForm` (including proptests for the old `ConstMontyForm::pow`)


tarcieri:
cd @andrewwhitehead 

This really wasn't so bad. The algorithm is already implemented in-place and there really wasn't that much left to decouple from `BoxedUint`

src/modular/boxed_monty_form/pow.rs
```diff
@@ -17,10 +20,16 @@ impl BoxedMontyForm {
     /// NOTE: `exponent_bits` may be leaked in the time pattern.
     #[must_use]
     pub fn pow_bounded_exp(&self, exponent: &BoxedUint, exponent_bits: u32) -> Self {
-        let z =
+        let mut z =
             pow_montgomery_form_amm(&self.montgomery_form, exponent, exponent_bits, &self.params);
 
-        Self::from_amm(z, self.params.clone())
+        // Perform the final reduction on the output
+        almost_montgomery_reduce(z.as_mut_uint_ref(), self.params.modulus().as_uint_ref());
```
tarcieri: Perhaps `pow_montgomery_form_amm` should just call `almost_montgomery_reduce` itself?

I think ideally we keep AMM internalized to functions working with it directly.

src/traits.rs
```diff
@@ -1137,7 +1140,14 @@ pub trait Monty:
     type Multiplier<'a>: Debug + Clone + MontyMultiplier<'a, Monty = Self>;
 
     /// The precomputed data needed for this representation.
-    type Params: 'static + Clone + Debug + Eq + Sized + Send + Sync;
+    type Params: 'static
+        + AsRef<GenericMontyParams<Self::Integer>>
```
tarcieri, line 1144: A bit of a hack to get around the indirection introduced by `BoxedMontyParams` and its `Arc`, but it does get us a single generic struct we can always use to access the params

src/modular/boxed_monty_form/pow.rs
tarcieri: Went ahead and did this, though I might extract it into its own PR

andrewwhitehead:
Looks good to me. I have some minor related changes for pow/pow_vartime, one that was dropped in the last merge, but that shouldn't conflict.

tarcieri:
Yeah, the two implementations should be nicely parallel/separate and out of each other's way.

Still need to add benchmarks to see if it actually is a better option for larger numbers of limbs, but if it turns out that's not the case we can always remove it.

PR #1133, opened 2026-01-20
Have `pow_montgomery_form_amm` reduce its output

tarcieri:
Extracts an `almost_montgomery_reduce` function which accepts a `&mut UintRef` as input and conditionally subtracts the modulus twice, as explained in extensive comments.

This required porting `conditional_borrowing_sub_assign` to `UintRef`, as well as adding a `CtLt` impl to `UintRef`. `Uint::lt` and `BoxedUint::conditional_borrowing_sub_assign` now both delegate these functions to `UintRef`.

This also adds a `AsRef<GenericMontyParams<Self::Integer>>` bound to `Monty::Params` which can be used to obtain the underlying struct, which is used to query the params for the modulus in `pow_montgomery_form_amm`

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -10,7 +10,10 @@ pub use num_traits::{
     WrappingSub,
 };
 
-use crate::{Choice, CtOption, Limb, NonZero, Odd, Reciprocal, UintRef, modular::Retrieve};
+use crate::{
+    Choice, CtOption, Limb, NonZero, Odd, Reciprocal, UintRef,
+    modular::{GenericMontyParams, Retrieve},
+};
 use core::fmt::{self, Debug};
 
 #[cfg(feature = "rand_core")]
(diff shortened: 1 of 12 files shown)


src/traits.rs
```diff
@@ -1137,7 +1140,14 @@ pub trait Monty:
     type Multiplier<'a>: Debug + Clone + MontyMultiplier<'a, Monty = Self>;
 
     /// The precomputed data needed for this representation.
-    type Params: 'static + Clone + Debug + Eq + Sized + Send + Sync;
+    type Params: 'static
+        + AsRef<GenericMontyParams<Self::Integer>>
```
tarcieri, line 1144: Perhaps `Monty::params` could just be changed to be:

```rust
fn params(&self) -> &GenericMontyParams<Self::Integer>;
```

...though that seems like it could use its own PR

tarcieri: It seems like actually changing it is complicated by e.g. `Monty::new_params_vartime` which does still need to make `BoxedMontyParams` for `BoxedMontyForm`, so I guess this little hack may not be too bad considering we still need the trait to be able to express that and abstract its way around the `Arc`

tarcieri: Well I went ahead and gave it a try: https://github.com/RustCrypto/crypto-bigint/pull/1134

tarcieri:
Going to go ahead and merge this then we can decide on #1134

PR #1134, opened 2026-01-20
Have `Monty::params` return `GenericMontyParams`

tarcieri:
Previously it wasn't possible to actually retrieve any of the parameters in a generic context. Returning a struct instead of an associated type makes that possible.

It's a bit confusing because there's still a `Monty::Params` type which is different: this is needed to gloss over the distinction between `BoxedMontyParams` which contains an
`Arc<GenericMontyParams<BoxedUint>>`.

To do that, it adds an `AsRef<GenericMontyParams<Self::Integer>>` bound to represent that bit of indirection, and also a
`From<GenericMontyParams<U>>` bound to make generic conversions possible.

A generic impl of `AsRef<GenericMontyParams<U>>` is also added to `GenericMontyParams<U>>` which suffices for the `MontyParams` type alias.

Now `Monty::params` returns a concrete generic type rather than an associated type, which means we can put any shared functionality we want in `GenericMontyParams`.

cc @andrewwhitehead

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -10,7 +10,10 @@ pub use num_traits::{
     WrappingSub,
 };
 
-use crate::{Choice, CtOption, Limb, NonZero, Odd, Reciprocal, UintRef, modular::Retrieve};
+use crate::{
+    Choice, CtOption, Limb, NonZero, Odd, Reciprocal, UintRef,
+    modular::{GenericMontyParams, Retrieve},
+};
 use core::fmt::{self, Debug};
 
 #[cfg(feature = "rand_core")]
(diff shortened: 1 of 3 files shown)


tarcieri:
@fjarri FYI, this would be a breaking change, and I suspect it might impact `crypto-primes`. But we need to make changes for `rand_core` anyway

src/traits.rs
```diff
@@ -1158,7 +1169,7 @@ pub trait Monty:
 
     /// Returns the parameter struct used to initialize this object.
     #[must_use]
-    fn params(&self) -> &Self::Params;
+    fn params(&self) -> &GenericMontyParams<Self::Integer>;
```
tarcieri, line 1172: Note this would not be ideal if someone were ever to use it to construct `BoxedMontyForm` via the `From` impl above as it would lose the `Arc` and they would need to clone it

tarcieri:
Not sure this is worth it. This kind of conflates the two parameters at the same API level, whereas as-is it's clearer what you're dealing with and how to get to the other.

PR #1139, opened 2026-01-20
Add modular square root calculation for prime modulus `ConstMontyForm`

andrewwhitehead:
This implements a new macro, `const_prime_monty_params!` as an alternative to `const_monty_params!` for a prime modulus. The resulting `ConstMontyForm` values support a `sqrt` method, calculating the modular square root (if it exists). Support for `MontyForm` is possible, but it would not be very ergonomic as the `PrimeParams` would need to be passed around separately.


andrewwhitehead:
This is really only in draft because I thought there might be debate about the macro naming or other aspects, I can mark it for review

tarcieri:
Would be nice to get it into a release, especially as it could be useful for `primefield` to replace the implementation there

PR #1140, opened 2026-01-21
Have `Monty::zero` and `Monty::one` borrow params

tarcieri:
This is more or less a followup to #1087 which had `BoxedMontyForm::new` auto-clone its params (since it's an `Arc` and that's cheap).

This does the same for these trait methods, where previously to use them you always had to clone the params yourself.


tarcieri:
I'm not going to merge this right away as it's a breaking change and we *just* got a new `crypto-primes` release out earlier

andrewwhitehead:
For consistency, `from_montgomery` could also accept a params reference

PR #1247, opened 2026-04-02
impl ConstantTimeGreater and ConstantTimeLess for ConstMontyForm

joeymich:
Implements `ctutils::CtGt`, `ctutils::CtLt`, `subtle::ConstantTimeGreater`, and `subtle::ConstantTimeLess`  for `ConstMontyForm`.

Addresses the following TODOs in `elliptic-curves` ([L722](https://github.com/RustCrypto/elliptic-curves/blob/cd6425facdd9bb2a26dd097e9270b9eef2e064ac/primefield/src/monty.rs#L722), [L732](https://github.com/RustCrypto/elliptic-curves/blob/cd6425facdd9bb2a26dd097e9270b9eef2e064ac/primefield/src/monty.rs#L732), [L764](https://github.com/RustCrypto/elliptic-curves/blob/cd6425facdd9bb2a26dd097e9270b9eef2e064ac/primefield/src/monty.rs#L764), [L774](https://github.com/RustCrypto/elliptic-curves/blob/cd6425facdd9bb2a26dd097e9270b9eef2e064ac/primefield/src/monty.rs#L774))

### DIFF
--- a/src/modular/const_monty_form/ct.rs
+++ b/src/modular/const_monty_form/ct.rs
@@ -39,7 +39,7 @@ where
     MOD: ConstMontyParams<LIMBS>,
 {
     fn ct_gt(&self, other: &Self) -> Choice {
-        self.montgomery_form.ct_gt(&other.montgomery_form)
+        self.retrieve().ct_gt(&other.retrieve())
     }
 }
 
@@ -48,7 +48,7 @@ where
     MOD: ConstMontyParams<LIMBS>,
 {
     fn ct_lt(&self, other: &Self) -> Choice {
-        self.montgomery_form.ct_lt(&other.montgomery_form)
+        self.retrieve().ct_lt(&other.retrieve())
     }
 }
 
--- a/src/modular/const_monty_form/ct.rs
+++ b/src/modular/const_monty_form/ct.rs
@@ -2,7 +2,7 @@
 
 use super::{ConstMontyForm, ConstMontyParams};
 use crate::{Choice, CtAssign, CtEq};
-use ctutils::{CtAssignSlice, CtEqSlice, CtSelectUsingCtAssign};
+use ctutils::{CtAssignSlice, CtEqSlice, CtGt, CtLt, CtSelectUsingCtAssign};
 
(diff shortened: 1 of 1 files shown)


src/modular/const_monty_form/ct.rs
```diff
@@ -34,6 +34,24 @@ impl<MOD, const LIMBS: usize> CtEqSlice for ConstMontyForm<MOD, LIMBS> where
 {
 }
 
+impl<MOD, const LIMBS: usize> CtGt for ConstMontyForm<MOD, LIMBS>
+where
+    MOD: ConstMontyParams<LIMBS>,
+{
+    fn ct_gt(&self, other: &Self) -> Choice {
+        self.montgomery_form.ct_gt(&other.montgomery_form)
```
tarcieri: This operation really needs to tell us about the ordering in canonical form since that's what we actually care about. The Montgomery form ordering will not reflect that.

As a trivial example, if we pick `N=7` and `R=10`, and have values `a=2` and `b=3`, in Montgomery form `a=6` (`2*10 mod 7`) and `b=2` (`3*10 mod 7`), in which case this would be the opposite ordering of what we want.

I would suggest for an initial implementation to just convert to canonical form using `retrieve()` before comparing.

fjarri: To be fair, there are cases where one needs just some reasonable ordering, for example if you want to store these numbers in a `BTreeMap`/`BTreeSet`. But it certainly can be confusing.

tarcieri: I don't think it makes sense to implement it in a way that doesn't match to the canonical form ordering, but I also don't actually have a use case for these comparisons.

joeymich: Ah, good catch, I overlooked that. I'll update the PR to compare canonical forms using `retrieve()`.

If the user needs `Ord` for some collection, we could leave it up to them to define a newtype. I see we only impl `Ord` for the base `Uint` types and wrappers.

tarcieri (approved):

PR #1256, opened 2026-04-27
Add support for exact division

andrewwhitehead:
This adds `div_exact` and `div_exact_vartime` methods for `Uint`, `BoxedUint`, and `Limb`. There is currently no trait support.

For cases where it applies, exact division can be much faster than div/rem. In my tests this is about 40% faster for a 4096-bit `BoxedUint`.

The algorithm is similar to `DivideByWord` in [Modern Computer Arithmetic](https://arxiv.org/pdf/1004.4710) but I don't have a reference for the adjustments to support a multi-word divisor.

### DIFF
--- a/src/primitives.rs
+++ b/src/primitives.rs
@@ -94,6 +94,43 @@ pub(crate) const fn u32_bits(n: u32) -> u32 {
     u32::BITS - n.leading_zeros()
 }
 
+cpubits::cpubits! {
+    32 => {
+        /// Returns the multiplicative inverse of the argument modulo 2^32.
+        ///
+        /// For correct results, the input `value` must be odd.
+        #[must_use]
+        pub(crate) const fn u32_invert_odd(value: u32) -> u32 {
+            debug_assert!(value & 1 == 1, "value must be odd");
+            let x = value.wrapping_mul(3) ^ 2;
+            let y = 1u32.wrapping_sub(x.wrapping_mul(value));
+            let (x, y) = (x.wrapping_mul(y.wrapping_add(1)), y.wrapping_mul(y));
+            let (x, y) = (x.wrapping_mul(y.wrapping_add(1)), y.wrapping_mul(y));
+            x.wrapping_mul(y.wrapping_add(1))
+        }
+    }
+}
+
+/// Returns the multiplicative inverse of the argument modulo 2^64. The implementation is based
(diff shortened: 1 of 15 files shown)


andrewwhitehead:
This just needs to be rebased on #1257 which is a separate issue.

src/primitives.rs
```diff
@@ -94,6 +94,43 @@ pub(crate) const fn u32_bits(n: u32) -> u32 {
...
+
+/// Returns the multiplicative inverse of the argument modulo 2^64. The implementation is based
+/// on the Hurchalla's method for computing the multiplicative inverse modulo a power of two, and
+/// is essentially an optimized Newton iteration.
+///
+/// For correct results, the input `value` must be odd.
+///
+/// For better understanding the implementation, the following paper is recommended:
+/// J. Hurchalla, "An Improved Integer Multiplicative Inverse (modulo 2^w)",
+/// <https://arxiv.org/pdf/2204.limbs4342.pdf>
```
tarcieri: FYI, this 404s. Looks like a find/replace error?

tarcieri:
```suggestion
/// <https://arxiv.org/abs/2204.04342>
```

src/uint/ref_type/div.rs
```diff
@@ -601,4 +599,156 @@ impl UintRef {
...
+    /// performing loops without updates to the quotient for larger divisors when vartime operation
+    /// is not specified.
+    #[allow(clippy::cast_possible_truncation)]
+    #[inline(always)]
+    const fn div_exact_odd_with_inverse(
+        x: &mut UintRef,
+        y: &Odd<UintRef>,
+        y_inv: Limb,
+        ywords: u32,
+        vartime: Choice,
```
tarcieri: Is there a specific reason this isn't e.g. a const generic parameter? It seems like you could just as easily do `if VARTIME & done.to_bool_vartime()` below.

src/primitives.rs
andrewwhitehead: Fixed

src/uint/ref_type/div.rs
andrewwhitehead: That was just for consistency with a couple other methods in the same file. I switched to a generic parameter (they're private or pub(crate) methods) and it made a small improvement in performance. The methods are inlined anyway so it doesn't really matter if they're expanded more than once.

tarcieri (approved):

PR #1260, opened 2026-05-03
Return Option from vartime inversion and pow methods

cong-or:
#1096 converted the vartime bit shift methods from `CtOption` to `Option`. This does the same for vartime inversion and pow.                                                                
                                                                                                                                                                                              
  Since these methods are already variable-time, wrapping results in `CtOption` doesn't add any constant-time guarantees — it just forces eager evaluation in combinators where `Option` would
   be lazy.
                                                                                                                                                                                              
  The `Invert` trait impls still return `CtOption` — #1097 tried changing that and got reverted in #1124 because `elliptic-curve` needs `Invert::Output` to stay flexible. So the trait impls
  convert back from `Option` via match.

### DIFF
--- a/benches/uint.rs
+++ b/benches/uint.rs
@@ -891,7 +891,7 @@ fn bench_invert_mod(c: &mut Criterion) {
                 loop {
                     let x = U256::random_from_rng(&mut rng);
                     let inv_x = x.invert_odd_mod_vartime(&m);
-                    if inv_x.is_some().into() {
+                    if inv_x.is_some() {
                         break (x, m);
                     }
                 }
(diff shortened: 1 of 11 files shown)


tarcieri:
I'm a little confused that `cargo semver-checks` isn't complaining, but these are breaking changes and we're in the stable part of a release cycle.

cong-or:
> I'm a little confused that `cargo semver-checks` isn't complaining, but these are breaking changes and we're in the stable part of a release cycle.

@tarcieri No worries — happy to rebase this when you're ready to collect breaking changes for the next release.

Issue #182, opened 2023-02-09
Any chance of supporting 8-bit platforms?


xphoniex:
I'm trying to compile [k256](https://github.com/RustCrypto/elliptic-curves/tree/master/k256) for ATmega328, a popular 8-bit chip on arduino boards, but I'm getting:

```rust
error: this crate builds on 32-bit and 64-bit platforms only
  --> /.cargo/registry/src/github.com-1ecc6299db9ec823/crypto-bigint-0.4.9/src/limb.rs:31:1
   |
31 | compile_error!("this crate builds on 32-bit and 64-bit platforms only");
```

The other alternatives to k256, secp256k1 and libsecp256k1 both need `alloc`, which I have not been able to get working yet on that platform. Any thoughts?

tarcieri:
With the forthcoming `cpubits` crate we may be able to select the 32-bit backend for 8-bit platforms, if that works: https://github.com/RustCrypto/utils/issues/824

xphoniex:
sidenote: I tried compiling for 32-bit ARM and it still failed because of `getrandom` crate:
```rust
error[E0463]: can't find crate for `std`
 --> /home/usr/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.8/src/error_impls.rs:9:1
  |
9 | extern crate std;
  | ^^^^^^^^^^^^^^^^^ can't find crate
  |
  = note: the `thumbv6m-none-eabi` target may not support the standard library

error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv6m-none-eabi` target may not support the standard library
  = note: `std` is required by `once_cell` because it does not declare `#![no_std]`

error: target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets
   --> /home/usr/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.8/src/lib.rs:268:9
    |
268 | /         compile_error!("target is not supported, for more information see: \
269 | |                         https://docs.rs/getrandom/#unsupported-targets");
    | |________________________________________________________________________^

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
   --> /home/usr/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.8/src/lib.rs:290:5
    |
290 |     imp::getrandom_inner(dest)
    |     ^^^ use of undeclared crate or module `imp`
```

[...]

PR #1265, opened 2026-05-12
Extract `bitlen` module with bytes/limbs conversions

tarcieri:
Adds a crate-internal module with the following infallible functions that always panic in the event of overflow/truncation rather than silently permitting it:

- `bitlen::from_bytes`
- `bitlen::from_limbs`
- `bitlen::to_bytes`
- `bitlen::to_limbs`

These wrap up the casting strategy for handling `usize` <=> `u32` conversions and checked arithmetic needed when converting from a number of bytes or a number of limbs represented as a `usize` to a number of bits represented as a `u32` and back.

Eliminates several `allow(clippy::cast_possible_truncation)` usages by handling the relevant concerns in a single place.

This does not appear to have a negative impact on benchmarks.

### DIFF
--- a/src/bitlen.rs
+++ b/src/bitlen.rs
@@ -1,0 +1,72 @@
+//! Bit calculations.
+//!
+//! These provide a single place to handle bits-to-bytes conversions and the `u32`/`usize`
+//! discrepancy that can only be solved by panics or casting.
+
+use crate::Limb;
+
+/// Calculate a `u32` representing a number of bits from the given number of bytes as a `usize`.
+#[cfg_attr(not(feature = "alloc"), allow(dead_code))]
+pub(crate) const fn from_bytes(nbytes: usize) -> u32 {
+    checked_mul(nbytes, 8)
+}
+
+/// Calculate a `u32` representing a number of bits in an integer with the given number of limbs.
+pub(crate) const fn from_limbs(nlimbs: usize) -> u32 {
+    checked_mul(nlimbs, Limb::BITS)
+}
+
+/// Calculate a `usize` representing the number of bytes required to store the given number of bits
+/// represented as a `u32` (i.e. rounding up).
+#[cfg_attr(not(feature = "alloc"), allow(dead_code))]
+pub(crate) const fn to_bytes(nbits: u32) -> usize {
+    u32_to_usize(nbits.div_ceil(8))
+}
+
+/// Calculate a `usize` representing the number of limbs required to store the given number of bits
+/// represented as a `u32` (i.e. rounding up).
+pub(crate) const fn to_limbs(nbits: u32) -> usize {
+    u32_to_usize(nbits.div_ceil(Limb::BITS))
+}
+
+const fn u32_to_usize(n: u32) -> usize {
+    const {
+        // Ensure cast from `u32` to `usize` won't overflow
+        assert!(usize::BITS >= u32::BITS, "usize too small");
+    }
+
+    n as usize
+}
+
+#[allow(
+    clippy::cast_possible_truncation,
+    reason = "assertion ensures panic instead of truncation"
+)]
+const fn usize_to_u32(n: usize) -> u32 {
+    assert!(n < u32_to_usize(u32::MAX), "u32 overflow");
+    n as u32
+}
+
+const fn checked_mul(size: usize, n: u32) -> u32 {
+    usize_to_u32(size).checked_mul(n).expect("u32 overflow")
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn from_bytes() {
+        assert_eq!(super::from_bytes(0), 0);
+        assert_eq!(super::from_bytes(1), 8);
+        assert_eq!(super::from_bytes(2), 16);
+    }
+
+    #[test]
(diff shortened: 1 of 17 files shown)


tarcieri:
cc @andrewwhitehead

src/bitlen.rs
```diff
@@ -0,0 +1,72 @@
...
+    u32_to_usize(nbits.div_ceil(8))
+}
+
+/// Calculate a `usize` representing the number of limbs required to store the given number of bits
+/// represented as a `u32` (i.e. rounding up).
+pub(crate) const fn to_limbs(nbits: u32) -> usize {
+    u32_to_usize(nbits.div_ceil(Limb::BITS))
+}
+
+const fn u32_to_usize(n: u32) -> usize {
```
tarcieri, line 32: This is perhaps a bit more generally useful but a private function for now

src/bitlen.rs
```diff
@@ -0,0 +1,72 @@
...
+    const {
+        // Ensure cast from `u32` to `usize` won't overflow
+        assert!(usize::BITS >= u32::BITS, "usize too small");
```
tarcieri, line 35: This won't work on 16-bit platforms, but we already have an assert that prevents building on them anyway.  See also: #182

tarcieri: I'll follow that up by saying that on 16-bit platforms we have the opposite problem of the one this is trying to solve: we can infallibly convert `usize` to `u32` on such platforms where this particular function is needed to convert `u32::MAX` to a `usize` for the check that a given `usize` won't overflow `u32`.

src/uint/boxed.rs
```diff
@@ -32,7 +32,8 @@ mod rand;
 
 use crate::{
     Choice, CtAssign, CtEq, CtOption, Integer, Limb, NonZero, Odd, One, Resize, UintRef, Unsigned,
-    UnsignedWithMontyForm, Word, Zero, modular::BoxedMontyForm, traits::sealed::Sealed,
+    UnsignedWithMontyForm, Word, Zero, bits::to_limbs as bits_to_limbs, modular::BoxedMontyForm,
```
tarcieri: The one slightly annoying part about a `bits` module is we already have quite a few `bits` modules like `uint::bits` and `uint::boxed::bits` which it clashes with.

Perhaps I could call it `bitlen` to make it a bit more unambiguous?

tarcieri:
Running the benchmarking suite I'm not seeing anything concerning, although this one jumped out at me, not sure what's going on here:

```
Benchmarking Boxed Montgomery arithmetic/lincomb_vartime, BoxedUint*BoxedUint+BoxedUint*BoxedUint: Collecting 100 samples in estimated 5.0635 s (258k iterat
Boxed Montgomery arithmetic/lincomb_vartime, BoxedUint*BoxedUint+BoxedUint*BoxedUint
                        time:   [13.183 µs 13.231 µs 13.280 µs]
                        change: [−21.284% −20.732% −20.253%] (p = 0.00 < 0.05)
                        Performance has improved.
```

src/uint/boxed.rs
tarcieri: Renamed to `bitlen` which seems clearer

src/bitlen.rs
```diff
@@ -0,0 +1,72 @@
...
+/// Calculate a `u32` representing a number of bits in an integer with the given number of limbs.
+pub(crate) const fn from_limbs(nlimbs: usize) -> u32 {
+    checked_mul(nlimbs, Limb::BITS)
+}
+
+/// Calculate a `usize` representing the number of bytes required to store the given number of bits
+/// represented as a `u32` (i.e. rounding up).
+#[cfg_attr(not(feature = "alloc"), allow(dead_code))]
+pub(crate) const fn to_bytes(nbits: u32) -> usize {
+    u32_to_usize(nbits.div_ceil(8))
```
andrewwhitehead, line 23: I wonder if the clippy lint (integer_division_remainder_used) should be detecting uses of `div_ceil` as well. This could be done with addition and a shift but then you have to worry about overflow.. in this case I guess something like `(nbits >> 3) + (((nbits & 7) + 7) >> 3)`

tarcieri: Good point about the use of division here. 

There are definitely some gaps in the `integer_division_remainder_used` lint: https://github.com/rust-lang/rust-clippy/issues/16490

src/uint/boxed/invert_mod.rs
```diff
@@ -149,22 +149,20 @@ impl Odd<BoxedUint> {
...
         let mut inv = BoxedUint::from_words_with_precision(*inv_64.as_words(), bits);
 
         if k_limbs <= U64::LIMBS {
             // trim to k_limbs
             inv.as_mut_uint_ref().trailing_mut(k_limbs).fill(Limb::ZERO);
         } else {
-            // expand to k_limbs
-            #[allow(clippy::cast_possible_truncation)]
-            let mut scratch = BoxedUint::zero_with_precision(k_limbs as u32 * 2 * Limb::BITS);
+            // expand to k_limbss
```
dishmaker, line 164:
```suggestion
            // expand to k_limbs
```

PR #1266, opened 2026-05-12
Add `BoxedUint::{from_be_slice_truncated, from_le_slice_truncated}`

tarcieri:
Adds infallible constructors for `BoxedUint` that handle the error case for the current `from_be_slice`/`from_le_slice` by truncating the input.

My real goal is to be able to add some slice constructors to the `Encoding` trait that can actually provide equivalent semantics between `Uint` and `BoxedUint` and also ensure construction of modulus-sized `BoxedUint`s.

A truncating behavior like this eliminates the error case and allows for an infallible constructor. As it were, it's largely mimicking the `bits2int` function described in RFC6979 § 2.3.2 which is what I'd like to implement using `Uint` for ECDSA an `BoxedUint` for DSA.

An infallible constructor lets us compose both `from_*e_slice_vartime` and `from_*e_slice` in terms of `from_*e_slice_truncated`, whee the former is also infallible and picks its capacity/precision on-the-fly based on the input data (e.g. we need that to decode the public modulus and figure out what precision we're working with).

### DIFF
--- a/src/uint/boxed/encoding.rs
+++ b/src/uint/boxed/encoding.rs
@@ -13,8 +13,8 @@ impl BoxedUint {
     /// The `bits_precision` argument represents the precision of the resulting integer, which is
     /// fixed as this type is not arbitrary-precision.
     ///
-    /// The new [`BoxedUint`] will be created with `bits_precision`
-    /// rounded up to a multiple of [`Limb::BITS`].
+    /// The new [`BoxedUint`] will be created with `bits_precision` rounded up to a multiple of
+    /// [`Limb::BITS`].
     ///
     /// # Errors
     /// - Returns [`DecodeError::InputSize`] if the length of `bytes` is larger than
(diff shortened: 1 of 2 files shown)


tarcieri:
cc @andrewwhitehead 

I'm going to try to flesh this out more with the trait changes I'd like and see if I can actually use it to implement the full RFC6979.

Edit: obviously needs tests as well.

src/uint/boxed/encoding.rs
```diff
@@ -39,6 +35,27 @@ impl BoxedUint {
...
     }
 
+    /// Create a new [`BoxedUint`] from the provided big endian bytes, zero padding if necessary,
+    /// and truncating to the least significant bytes in the event the given amount of data exceeds
+    /// `bits_precision`.
+    #[must_use]
+    pub fn from_be_slice_truncated(mut bytes: &[u8], bits_precision: u32) -> Self {
+        let bytes_precision = bitlen::to_bytes(bits_precision);
+
+        // TODO(tarcieri): mask bits in the most significant byte if necessary
```
tarcieri: Implementing this will break the current `DecodeError::Precision` checks in `from_be_slice`/`from_le_slice`, so those need to be updated first.

I'll also note it's a bit weird those methods return `DecodeError::InputSize` if the input slice is too long, and `DecodeError::Precision` if it contains bits that exceed the given precision, but it shouldn't be too hard to preserve that behavior though it might need to be an eager check (those bits shouldn't be set in the first place and if they are it's an error, so we don't need to worry about constant-time).

src/uint/boxed/encoding.rs
```diff
@@ -163,9 +187,8 @@ impl BoxedUint {
     /// # Panics
     /// - if hex string is not the expected size
     #[must_use]
-    #[allow(clippy::integer_division_remainder_used, reason = "public parameter")]
     pub fn from_be_hex(hex: &str, bits_precision: u32) -> CtOption<Self> {
-        let nlimbs = (bits_precision / Limb::BITS) as usize;
+        let nlimbs = bitlen::to_limbs(bits_precision);
```
andrewwhitehead, line 194: Hmm, seems like it currently cuts off the top limb if bits_precision isn't a multiple of Limb::BITS?

tarcieri: I should probably pull this into its own PR and add some tests.

tarcieri: Pulled this into its own PR with a test: https://github.com/RustCrypto/crypto-bigint/pull/1276

The answer is "yes, but" in that a subsequent assert detects the hex string is short and panics.

src/uint/boxed/encoding.rs
tarcieri: The original behavior of `from_*e_slice` is now preserved except for the case where the number of bits is not a multiple of the limb size, which previously returned `DecodeError::Precision` but now succeeds with masking. I guess such a change might technically be considered breaking but I don't think it will have any practical impact.

If it turns out to be a problem we can add an additional check.

PR #1270, opened 2026-06-04
Fix BoxedUint bitor-assign with wider rhs

tob-joe:
## Summary

`BoxedUint` supports mixed-precision bitwise OR by returning a result sized to the wider operand. The `BitOrAssign<&BoxedUint>` implementation did not follow that behavior: it only ORed the overlapping limb prefix. If rhs had more limbs than lhs, the high rhs limbs were dropped.

## Fix

Grow the lhs limb buffer only when rhs is wider, padding new limbs with zero, then OR rhs limbs into lhs in place. Equal-width and narrower-rhs assignments do not allocate. Wider-rhs assignment replaces the boxed slice because `BoxedUint` stores limbs as a fixed-length `Box<[Limb]>`.

## Tests

Added `bitor_assign_preserves_wider_rhs_limbs`, which OR-assigns a one-limb lhs with a two-limb max rhs and checks that the assignment result matches non-assignment bitwise OR.

-----

### DIFF
--- a/src/uint/boxed/bit_or.rs
+++ b/src/uint/boxed/bit_or.rs
@@ -68,9 +68,7 @@ impl BitOrAssign for BoxedUint {
 impl BitOrAssign<&BoxedUint> for BoxedUint {
     fn bitor_assign(&mut self, other: &Self) {
         if other.limbs.len() > self.limbs.len() {
-            let mut limbs = core::mem::take(&mut self.limbs).into_vec();
-            limbs.resize(other.limbs.len(), Limb::ZERO);
-            self.limbs = limbs.into_boxed_slice();
+            self.resize_in_place_unchecked(other.bits_precision());
         }
 
         for (a, b) in self.limbs.iter_mut().zip(other.limbs.iter()) {
--- a/src/uint/boxed/bit_or.rs
+++ b/src/uint/boxed/bit_or.rs
@@ -1,13 +1,13 @@
 //! [`BoxedUint`] bitwise OR operations.
 
-use crate::{BitOr, BitOrAssign, BoxedUint, CtOption};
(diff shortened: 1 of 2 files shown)


src/uint/boxed/bit_or.rs
```diff
@@ -67,8 +67,30 @@ impl BitOrAssign for BoxedUint {
 
 impl BitOrAssign<&BoxedUint> for BoxedUint {
     fn bitor_assign(&mut self, other: &Self) {
+        if other.limbs.len() > self.limbs.len() {
+            let mut limbs = core::mem::take(&mut self.limbs).into_vec();
+            limbs.resize(other.limbs.len(), Limb::ZERO);
+            self.limbs = limbs.into_boxed_slice();
```
tarcieri: Hmm, we should probably have at least an internal function ala `BoxedUint::resize_in_place`

tob-joe: Addressed in `47c51aa` by adding a crate-private `BoxedUint::resize_in_place_unchecked` helper and using it from `BitOrAssign<&BoxedUint>` instead of open-coding the limb-buffer resize. The helper preserves the existing `resize_unchecked` behavior, including unchecked truncation on shrink and canonicalizing zero requested precision via `Self::from(limbs)`. I also changed consuming `resize_unchecked` to delegate to the helper.

Checks run:

- `cargo test --features alloc bitor_assign_preserves_wider_rhs_limbs -- --nocapture`
- `cargo check --features alloc`

tarcieri (approved):

src/uint/boxed/bit_or.rs
tob-joe: agreed!

PR #1271, opened 2026-06-04
Handle wide power-of-two divisors in exact division

tob-joe:
## Summary

`UintRef::div_exact` could panic when the divisor was wider than the dividend and had exactly as many trailing zeros as the dividend precision. In that case, exact division should report inexact division rather than indexing or shifting past the dividend.

## Fix

`ensure_trailing_zeros` now handles zero-count requests that cover the full dividend width by checking whether the whole dividend is zero, instead of indexing the limb at `nlimbs`.

After the trailing-zero exactness check, `div_exact` now uses an unbounded right shift so shifting by the full dividend precision produces zero instead of panicking.

## Tests

Added a regression test covering exact division of `U128::ONE` by a `U256` power-of-two divisor at bit position `U128::BITS`.

Verified with:

```text
cargo test --all-features --test uint div_exact_wide_power_of_two_divisor_is_inexact -- --exact
```

-----

### DIFF
--- a/src/uint/ref_type/div.rs
+++ b/src/uint/ref_type/div.rs
@@ -620,7 +620,7 @@ impl UintRef {
 
         // Check that the dividend evenly divides by 2^tz, and shift it to match the divisor
         let div2s_exact = self.ensure_trailing_zeros(tz).and(excess_z.not());
-        self.shr_assign(tz);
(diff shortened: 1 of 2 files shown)


andrewwhitehead (approved):
LGTM

tarcieri (approved):

PR #1272, opened 2026-06-04
Enforce UintRef slice length checks in release builds

tob-joe:
## Summary

`UintRef::copy_from_slice` documents that it panics when `self.nlimbs() != limbs.len()`, but the check used `debug_assert!`, so optimized builds did not enforce that contract:

https://github.com/RustCrypto/crypto-bigint/blob/4c6f87d35dce6bdfc60113f64ab5dfe396db89ef/src/uint/ref_type/slice.rs#L25-L33

`UintRef::conditional_copy_from_slice` has the same runtime length check pattern and its `# Panics` docs also state a limb-count mismatch panic, though the current wording says `rhs.nlimbs()` despite taking a limb slice:

https://github.com/RustCrypto/crypto-bigint/blob/4c6f87d35dce6bdfc60113f64ab5dfe396db89ef/src/uint/ref_type/slice.rs#L41-L48

## Fix

Use `assert!` for both length checks so debug and release builds preserve the documented panic behavior.

## Tests

Added regression tests for mismatched lengths in both slice-copy APIs and verified them in release mode:

```text
cargo test --release --all-features uint::ref_type::slice::tests -- --nocapture
```

-----

### DIFF
--- a/src/uint/ref_type/slice.rs
+++ b/src/uint/ref_type/slice.rs
@@ -30,7 +30,7 @@ impl UintRef {
     #[track_caller]
     pub const fn copy_from_slice(&mut self, limbs: &[Limb]) {
         // TODO core::slice::copy_from_slice should eventually be const
-        debug_assert!(self.limbs.len() == limbs.len(), "length mismatch");
+        assert!(self.limbs.len() == limbs.len(), "length mismatch");
         let mut i = 0;
         while i < self.limbs.len() {
             self.limbs[i] = limbs[i];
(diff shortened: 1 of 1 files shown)


andrewwhitehead:
More specifically the two methods are ignoring the high limbs of `limbs` instead of panicking in release mode. I think an earlier version might have relied on an out of bounds error instead? As part of some other work I'd like to adjust both of these methods to accept a slice shorter than `self`, but that's a later PR.

andrewwhitehead (approved):

tarcieri (approved):

PR #1274, opened 2026-06-05
Reject oversized fixed-width DER integers

tob-joe:
## Summary

Fixed-width `Uint` DER decoding converts an ASN.1 INTEGER into a fixed-size byte array:

https://github.com/RustCrypto/crypto-bigint/blob/4c6f87d35dce6bdfc60113f64ab5dfe396db89ef/src/uint/encoding/der.rs#L20-L30

Before this change, an oversized canonical positive INTEGER could compute an offset with `saturating_sub(0)` and then panic while copying too many bytes into the fixed-width array.

## Fix

Check the decoded INTEGER byte length before copying. If it is wider than the fixed-width `Uint` target, return `Tag::Integer.length_error()` instead of unwinding.

## Tests

Added a regression test that decodes a 17-byte positive INTEGER as `U128` and verifies the operation returns an error without panicking.

Verified with:

```text
cargo test --all-features uint::encoding::der::test::fixed_uint_rejects_oversized_integer_without_panic -- --exact
```

-----

### DIFF
--- a/src/uint/encoding/der.rs
+++ b/src/uint/encoding/der.rs
@@ -25,13 +25,16 @@ where
 
     fn try_from(bytes: Asn1UintRef<'a>) -> der::Result<Uint<LIMBS>> {
         let mut array = Array::default();
-        let len: usize = bytes.len().try_into()?;
-        if len > array.len() {
-            return Err(Tag::Integer.length_error().into());
-        }
+        let len = usize::try_from(bytes.len())?;
+        let offset = array
+            .len()
+            .checked_sub(len)
+            .ok_or_else(|| Tag::Integer.length_error())?;
+        let out = array
+            .get_mut(offset..)
+            .ok_or_else(|| Tag::Integer.length_error())?;
 
-        let offset = array.len() - len;
-        array[offset..].copy_from_slice(bytes.as_bytes());
+        out.copy_from_slice(bytes.as_bytes());
         Ok(Uint::from_be_byte_array(array))
     }
 }
--- a/src/uint/encoding/der.rs
+++ b/src/uint/encoding/der.rs
@@ -25,7 +25,12 @@ where
 
     fn try_from(bytes: Asn1UintRef<'a>) -> der::Result<Uint<LIMBS>> {
         let mut array = Array::default();
-        let offset = array.len().saturating_sub(bytes.len().try_into()?);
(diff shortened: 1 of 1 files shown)


src/uint/encoding/der.rs
```diff
@@ -226,4 +232,16 @@ pub mod test {
...
     }
+
+    #[test]
+    fn fixed_uint_rejects_oversized_integer_without_panic() {
+        use der::Decode;
+
+        let mut der = vec![0x02, 17, 0x01];
+        der.extend_from_slice(&[0u8; 16]);
+
+        let result = std::panic::catch_unwind(|| U128::from_der(&der));
```
andrewwhitehead: A panic is already a test failure, we don't need to capture it.

tob-joe: ah yeah good catch!

src/uint/encoding/der.rs
```diff
@@ -25,7 +25,12 @@ where
 
     fn try_from(bytes: Asn1UintRef<'a>) -> der::Result<Uint<LIMBS>> {
         let mut array = Array::default();
-        let offset = array.len().saturating_sub(bytes.len().try_into()?);
+        let len: usize = bytes.len().try_into()?;
```
tarcieri: I'd consider this a little more idiomatic:

```suggestion
        let len = usize::try_from(bytes.len())?;
```

src/uint/encoding/der.rs
```diff
@@ -25,7 +25,12 @@ where
...
+        if len > array.len() {
+            return Err(Tag::Integer.length_error().into());
+        }
+
+        let offset = array.len() - len;
         array[offset..].copy_from_slice(bytes.as_bytes());
```
tarcieri: It seems like you could use `get_mut` here, along with checked arithmetic, and by doing so avoid the need for a manual length check.

tob-joe:
Addressed the DER decode review feedback in `2a9fb1b`:

- switched the DER INTEGER length conversion to `usize::try_from(bytes.len())?`
- replaced the manual length check with checked offset arithmetic plus `get_mut`

Checks run:

- `cargo test --all-features uint::encoding::der::test::fixed_uint_rejects_oversized_integer_without_panic -- --exact`
- `git diff --check -- src/uint/encoding/der.rs`

tarcieri (approved):

PR #1278, opened 2026-06-08
Update MontyParams CtEq to match PartialEq

tob-joe:
MontyParams derives PartialEq/Eq over all fields, but the runtime parameter CtEq implementations omitted the cached mod_leading_zeros field. That allowed internally constructible parameter values to compare unequal with == while CtEq returned true.

This updates generic MontyParams and BoxedMontyParams CtEq to include mod_leading_zeros, and adds regression tests for values that differ only in that field.

Testing:
- cargo test --all-features mod_leading_zeros -- --nocapture

-----


andrewwhitehead (approved):
LGTM

tarcieri (approved):

PR #1282, opened 2026-06-08
Handle power-of-two moduli in mul_mod_special

### LINKED ISSUE #1281: mul_mod_special panics for c = 0 in one-limb implementations
`mul_mod_special` is documented as computing multiplication modulo the special modulus `p = MAX + 1 - c`, i.e. `2^bits - c`.

When `c = 0`, that modulus is `2^bits`, so the expected result is ordinary fixed-width wrapping multiplication:

```text
(a * b) mod 2^bits == a.wrapping_mul(b)
```

The one-limb fixed and boxed implementations instead reach a reduction path that constructs `NonZero<Limb>` from `0`:

```rust
NonZero::<Limb>::new_unwrap(Limb::ZERO.wrapping_sub(c))
```

With `c = 0`, this panics with `invalid value: zero`. This affects `Uint<1>::mul_mod_special` and one-limb `BoxedUint::mul_mod_special`. Wider inputs already naturally behave like wrapping multiplication for this `c = 0` power-of-two modulus case.

Expected behavior: `mul_mod_special(..., Limb::ZERO)` should return the same value as `wrapping_mul` for the same precision.

tob-joe:
Closes #1281.

`mul_mod_special` computes multiplication modulo `MAX + 1 - c`, i.e. `2^bits - c`. When `c = 0`, that modulus is `2^bits`, so reduction is ordinary fixed-width wrapping multiplication.

The one-limb fixed and boxed implementations reached a limb reduction path which constructs `NonZero<Limb>` from `0` when `c = 0`, causing a panic. This adds an early `wrapping_mul` return for the power-of-two modulus case and regression tests for both fixed and boxed integers.

Testing:

- `cargo test --all-features mul_mod_special_zero_c_is_wrapping_multiplication -- --nocapture`
- `cargo test --all-features uint::mul_mod::tests -- --nocapture`
- `cargo test --all-features uint::boxed::mul_mod::tests -- --nocapture`
- `git diff --check`

-----

### DIFF
--- a/src/uint/boxed/mul_mod.rs
+++ b/src/uint/boxed/mul_mod.rs
@@ -22,6 +22,10 @@ impl BoxedUint {
     pub fn mul_mod_special(&self, rhs: &Self, c: Limb) -> Self {
         debug_assert_eq!(self.bits_precision(), rhs.bits_precision());
 
+        if c.is_zero().into() {
+            return self.wrapping_mul(rhs);
+        }
+
         if self.nlimbs() == 1 {
             let reduced = mul_rem(
                 self.limbs[0],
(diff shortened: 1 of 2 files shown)


andrewwhitehead:
These methods are not marked as vartime, I don't think it's appropriate to be short circuiting based on the value of the modulus. They should probably accept a `NonZero<Limb>` instead, although that would be a breaking change.

PR #1286, opened 2026-06-08
Implement Encoding for Int serde support

tob-joe:
Implement `Encoding` for `Int<LIMBS>`.

`Int` already has `Serialize` and `Deserialize` impls behind the `serde` feature, but those impls require `Int<LIMBS>: Encoding`. Since `Int` did not implement `Encoding`, the serde impls were not reachable for signed fixed-width integer aliases like `I256`.

This implements `Encoding` for `Int<LIMBS>` by delegating to the wrapped `Uint<LIMBS>` representation. That preserves the existing two's-complement byte layout used by signed integers.

The tests add a dedicated serde trait coverage matrix for the crate's explicit serde impl surfaces, including representative fixed-width unsigned and signed integers, invariant wrappers, arithmetic wrappers, boxed integers under `alloc`, and `ConstMontyForm`. The tests also exercise `I256` deserialization through `serdect` without adding a concrete serializer dev-dependency.

-----

### DIFF
--- a/src/int/encoding.rs
+++ b/src/int/encoding.rs
@@ -1,6 +1,6 @@
 //! Const-friendly decoding/encoding operations for [`Int`].
 
-use crate::{Int, Uint};
+use crate::{EncodedUint, Encoding, Int, Uint};
 
 impl<const LIMBS: usize> Int<LIMBS> {
     /// Create a new [`Int`] from the provided big endian hex string.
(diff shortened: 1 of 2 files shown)


andrewwhitehead:
This looks good to me. It does appear that `from_be_bytes`/`to_be_bytes` aren't being tested for `Int` at the moment. I was a little surprised that the serde test doesn't increase coverage but I see it's just testing that the trait is implemented, maybe we should also be attempting a round trip?

tarcieri (approved):

tarcieri:
> maybe we should also be attempting a round trip?

That sounds good, though I'm going to land this

PR #1287, opened 2026-06-08
Preserve `NonZero` and `Odd` invariants in `Zeroize` impls

tob-joe:
Preserve `NonZero` and `Odd` wrapper invariants after `zeroize()`.

The current `Zeroize` impls zeroize the wrapped value directly. For `NonZero<T>` and `Odd<T>`, that can leave a live wrapper containing zero or an even value after a safe method call, violating the wrapper's type invariant.

This changes the impls to zeroize the inner storage and then restore a one-like sentinel value. That keeps the wrapper valid after zeroization, mirrors the `zeroize` crate's behavior for core `NonZero*` integer types, and preserves boxed integer precision through `One::one_like`.

The tests cover stack and boxed `NonZero`/`Odd` values, including boxed precision preservation after zeroization.

-----

### DIFF
--- a/src/non_zero.rs
+++ b/src/non_zero.rs
@@ -662,9 +662,10 @@ where
 }
 
 #[cfg(feature = "zeroize")]
-impl<T: zeroize::Zeroize + Zero> zeroize::Zeroize for NonZero<T> {
+impl<T: zeroize::Zeroize + One> zeroize::Zeroize for NonZero<T> {
     fn zeroize(&mut self) {
         self.0.zeroize();
+        self.0 = T::one_like(&self.0);
     }
 }
 
@@ -673,6 +674,12 @@ mod tests {
     use super::NonZero;
     use crate::{I128, One, U128};
     use hex_literal::hex;
+
+    #[cfg(all(feature = "alloc", feature = "zeroize"))]
+    use crate::BoxedUint;
+
+    #[cfg(feature = "zeroize")]
+    use zeroize::Zeroize;
 
     #[test]
     fn default() {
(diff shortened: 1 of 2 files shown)


src/non_zero.rs
```diff
@@ -662,9 +662,10 @@ where
 }
 
 #[cfg(feature = "zeroize")]
-impl<T: zeroize::Zeroize + Zero> zeroize::Zeroize for NonZero<T> {
+impl<T: zeroize::Zeroize + One> zeroize::Zeroize for NonZero<T> {
     fn zeroize(&mut self) {
         self.0.zeroize();
+        self.0 = T::one_like(&self.0);
```
andrewwhitehead, line 668: In the case of a `NonZero<BoxedUint>`, this actually leaves the original value in memory and creates a new allocation (unless the compiler optimizes it away, but we can't assume that). It might be best to zeroize the internal value, then use `One::set_one` to restore an acceptable state. Either way I think the modified bounds on the implementation make this a breaking change that would need to go into a future minor version.

tarcieri: Hmm, maybe we can either actually make a proper breaking v0.8 release (I'd like to seal `Encoding` for #1266 as well), or I can make some extra-last-minute tiny breaking changes to v0.7 and yank the previous releases.

Within @RustCrypto we haven't cut any releases of anything that would preclude a v0.8.

tarcieri: I'm going to go ahead and merge this one. I'm planning on yanking the previous releases due to #1305 anyway, and while this is a "breaking" change, in the extremely unlikely event that someone relying on the behavior that allows zero here is constructing values which violate the invariant and has a buggy program.

(sidebar: seems like `cargo-semver-checks` doesn't consider this a breaking change, which is probably an oversight)

PR #1291, opened 2026-06-10
Fix BoxedUint GCD with mixed-precision zero operands

tob-joe:
## Summary

`BoxedUint::gcd` does not correctly handle mixed-precision zero operands.

The constant-time path only substitutes a nonzero placeholder for the left operand before calling the nonzero GCD helper. That causes two problems:

- `gcd(0_wide, x_narrow)` can panic when the final constant-time correction assigns a narrow value into a wider result.
- `gcd(x_wide, 0_narrow)` can return the wrong value when `x` has more trailing zero bits than the zero right operand has precision.

The second case occurs because `gcd_nz` computes a common power-of-two factor using `min(lhs.trailing_zeros(), rhs.trailing_zeros())`. For a zero `BoxedUint`, `trailing_zeros()` is capped by that value's precision, so a narrow zero right operand can cause the result to lose part of the power-of-two factor.

## Fix

This changes the `BoxedUint` GCD wrapper to handle zero operands symmetrically.

For the variable-time path, the wrapper now returns the nonzero operand directly when either side is zero, and returns zero for `gcd(0, 0)`.

For the constant-time path, the wrapper now substitutes one for both zero operands before calling `gcd_nz`, then conditionally corrects the result with the original operands. This avoids passing a zero right operand into `gcd_nz` while keeping the output precision a public function of the input precisions.

## Regression Tests

The new tests cover:

- `gcd(0_wide, x_narrow)`
- `gcd(x_narrow, 0_wide)`
- `gcd(x_wide, 0_narrow)` where `x` has more trailing zero bits than the zero operand has precision
- the same right-zero wrong-result case for `gcd_vartime`
- mixed-precision nonzero operands
- mixed-precision `gcd(0, 0)`

## Verification

```text
cargo test --all-features uint::boxed::gcd::tests::gcd_ -- --nocapture
cargo fmt --all -- --check
cargo clippy --all-targets --all-features -- -D warnings
git diff --check
```

-----

### DIFF
--- a/src/modular/safegcd/boxed.rs
+++ b/src/modular/safegcd/boxed.rs
@@ -129,17 +129,23 @@ fn invert_odd_mod_precomp<const VARTIME: bool>(
 /// Calculate the greatest common denominator of `f` and `g`.
 pub fn gcd<const VARTIME: bool>(f: &BoxedUint, g: &BoxedUint) -> BoxedUint {
     if VARTIME {
-        if let Some(f_nz) = f.as_nz_vartime() {
-            gcd_nz::<VARTIME>(f_nz, g).get()
-        } else {
-            // gcd of (0, g) is g
-            g.clone()
+        match (f.as_nz_vartime(), g.as_nz_vartime()) {
+            (Some(f_nz), Some(_)) => gcd_nz::<VARTIME>(f_nz, g).get(),
+            (Some(_), None) => f.clone(),
+            (None, Some(_)) => g.clone(),
+            (None, None) => {
+                BoxedUint::zero_with_precision(u32_max(f.bits_precision(), g.bits_precision()))
+            }
         }
     } else {
         let (f_nz, f_is_nonzero) = f.to_nz_or_one();
-        // gcd of (0, g) is g
-        let mut r = gcd_nz::<VARTIME>(&f_nz, g).get();
-        r.ct_assign(g, !f_is_nonzero);
+        let (g_nz, g_is_nonzero) = g.to_nz_or_one();
+        let mut r = gcd_nz::<VARTIME>(&f_nz, g_nz.as_ref()).get();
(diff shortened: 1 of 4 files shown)


andrewwhitehead:
I think we should be explicit in the size of the `BoxedUint` of the result, as we are (generally) for other operations, and always return the maximum bit precision between the two operands in this case. For the vartime implementation, this should just mean returning `ref.resize(max_precision)` instead of using `clone`. This could use documenting as well.

src/modular/safegcd/boxed.rs
```diff
@@ -163,7 +162,10 @@ pub fn gcd_nz<const VARTIME: bool>(f: &NonZero<BoxedUint>, g: &BoxedUint) -> Non
     // 4) 2^k•gcd(f, g) = 2^k•gcd(a, 2^j•b)
 
     let i = f.as_ref().trailing_zeros();
-    let k = u32_min(i, g.trailing_zeros());
+    let j = g
+        .is_nonzero()
+        .select_u32(f.bits_precision(), g.trailing_zeros());
+    let k = u32_min(i, j);
```
andrewwhitehead, line 168: This seems to be the real issue. Previously, if `g` was zero and smaller than `f`, then `k` would max out at the number of bits in `g`. This could lead to the wrong GCD being computed.

tob-joe: oh yeah, apologies for that being unclear! Currently if z holds 0, `gcd(z,x)` panics if `x.precision() < z.precision()`, while `gcd(x,z) =  x >> max(0, x.trailing_zeros() - z.precision())`

andrewwhitehead: That's from two commits I added to your PR BTW, it's an alternative fix really.

andrewwhitehead (approved):

PR #1310, opened 2026-06-25
Fixes for exact division edge cases

andrewwhitehead:
This adds fixes for a couple types of inputs that were not producing the correct result:

- Zero dividends should always divide exactly, even when the divisor has more limbs
- Inexact division could be undetected when the dividend had one or more zero-valued lower limbs

The suite of proptests for BoxedUint are updated to explicitly create inputs with a single bit set, only high bits set, and only low bits set (which also detects these issues). Some tests are expanded to test differently-sized inputs where this was not being exercised before.

### DIFF
--- a/tests/boxed_uint.rs
+++ b/tests/boxed_uint.rs
@@ -8,8 +8,8 @@ mod common;
 use common::to_biguint;
 use core::ops::Range;
 use crypto_bigint::{
-    BitOps, BoxedUint, CheckedAdd, Choice, ConcatenatingMul, ConcatenatingSquare, Gcd, Integer,
-    Limb, NonZero, Odd, Resize, Word,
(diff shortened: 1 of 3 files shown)


tests/boxed_uint.rs
```diff
@@ -150,6 +178,7 @@ proptest! {
     }
 
     #[test]
+    // TODO: expand to f in uint(), g in uint()
```
andrewwhitehead, line 187: Changing this highlights the issue fixed by #1291

andrewwhitehead:
This does make the test runs significantly slower at the moment, I'll look at reducing the frequency of tests for larger BoxedUints.

tarcieri (approved):

PR #1329, opened 2026-07-28
Add wide division: div_rem_wide and wrapping_div_wide

### LINKED ISSUE #1315: `Uint::wrapping_div_wide(_exact)`
As we have `rem_wide`, I'd like to request `wrapping_div_wide` with a type signature:

```rs
fn wrapping_div_wide<const HI_LIMBS: usize, const RHS_LIMBS: usize>((lo, hi): (Self, Uint<HI_LIMBS>), rhs: &NonZero<Uint<RHS_LIMBS>>) -> Self;
```

(or approximate)

For my specific use case, I know the division to be exact, hence why I'd be fine with a `wrapping_div_wide_exact` method instead of a `wrapping_div_wide` method. I did poke at adding this myself but it appeared non-trivial. Modifying `div_exact` to also intake a `x_hi: &mut UintRef`, indexing `x_lo` when `xi < x_lo.nlimbs()` and `x_hi` otherwise was fine, but it still poked a bit too much for me to be comfortable without first making an issue to discuss this.

My practical use-case is I have such a need within my code and currently do `lo.concat(hi).div_exact(rhs)`, but this requires the types implement `Concat`. As my segments are ~5k bits, the only types which reliably implement `Concat` are `U6144, U8192`, at which point I'm way over the size I'd like to be operating with. Removing the `Concat` bound would dramatically increase which types are eligible for use within my constraints.

This is horribly niche and I would not be offended at all if this was immediately closed as out-of-current-scope-availability/unreasonable.

cong-or:
## Title

Add wide division: `div_rem_wide` and `wrapping_div_wide`

Closes #1315.

## What this adds

Division for a **double-width dividend** — a number twice as wide as the `Uint` type, passed as a pair of halves `(lo, hi)` meaning `lo + hi * 2^BITS`, divided by a normal-width divisor.

Both constant-time and variable-time forms, mirroring the existing division methods on `Uint`:

```rust
// constant-time
Uint::div_rem_wide((lo, hi), rhs)            // -> (quotient, remainder)
Uint::wrapping_div_wide((lo, hi), rhs)       // -> quotient only
Uint::div_wide_exact((lo, hi), rhs)          // -> Some(quotient) if it divides evenly, else None

// variable-time in the divisor (constant-time for a fixed divisor)
Uint::div_rem_wide_vartime((lo, hi), rhs)
Uint::wrapping_div_wide_vartime((lo, hi), rhs)
Uint::div_wide_exact_vartime((lo, hi), rhs)
```

A quick example:

```rust
// dividend = 3 * 2^256 + 5, divided by 3
let (quo, rem) = U256::div_rem_wide((U256::from(5u64), U256::from(3u64)), &three);

// true quotient is 2^256 + 1, which doesn't fit in 256 bits, so it wraps to 1
assert_eq!(quo, U256::ONE);
assert_eq!(rem, U256::from(2u64));
```

## Why

`rem_wide` already lets you take the remainder of a double-width value without widening the type. This fills the obvious gap: getting the quotient of that same double-width value.

The workaround today is to glue the two halves together with `concat()` and divide the result. That has two downsides:

1. `concat()` pushes you up to the next `Uint` size, which only exists for certain sizes and means allocating a bigger type than you actually need (e.g. a ~5000-bit value forces you all the way to `U8192`).
2. It only works for types where the `Concat` trait is implemented.

`div_rem_wide` sidesteps both—it divides the `(lo, hi)` pair directly, at any limb count, with no wider type required. It's the natural quotient-returning companion to `rem_wide`.

Because the true quotient of a double-width dividend can itself be wider than the type, `wrapping_div_wide` keeps the low half (reduces mod `2^BITS`), hence the `wrapping_` name, consistent with the existing `wrapping_div`. `div_wide_exact` is the wide version of `div_exact`, for when you know the division comes out clean (in which case the quotient always fits).

## How it works

No new division algorithm—it reuses the existing engine.

The core is modeled directly on `rem_wide`/`rem_wide_large_shifted`: the same Knuth long division with the `div3by2` fast path and the same normalization. The one addition is that instead of throwing the quotient digits away, each digit (produced most-significant-first) is shifted into a small accumulator, yielding the low half of the quotient. Single-limb divisors take the simpler `div2by1` path, exactly like `div_rem`.

The variable-time form mirrors `rem_wide_vartime` (trimming the divisor and stopping early), sharing the same core loop behind a `const VARTIME` flag.

The implementation is a handful of small functions that parallel the existing `rem_wide` family, so it should read familiarly alongside the surrounding code.

## Testing

The new code is checked against a deliberately simple, trusted reference: glue the halves with `concat()`, run the existing (well-tested) `div_rem`, and compare the quotient and remainder—for both the constant-time and variable-time methods.

This runs over:

- Randomized inputs at `U64`, `U128`, `U192`, and `U256` (a few thousand cases per run).
- Full-width, single-word, and mid-width divisors, chosen so every internal branch is exercised (including the "divisor is narrower than the dividend" case and the variable-time early exit).
- Hand-written edge cases: zero halves, divide-by-one, all-ones, and exact divisions.

All 561 existing library tests still pass, and `clippy` and `rustfmt` are clean.

## One thing I'd like your call on

- **Scope.** The methods use a uniform width for `lo`, `hi`, and `rhs`, matching `rem_wide`. If you'd also like mixed-width generics (a wider `hi` / different-width `rhs`, as sketched in the issue), I'm happy to follow up with that separately.

### DIFF
--- a/src/uint/div.rs
+++ b/src/uint/div.rs
@@ -2,8 +2,8 @@
 
 use super::div_limb::Reciprocal;
 use crate::{
-    CheckedDiv, CtOption, Div, DivAssign, DivRemLimb, DivVartime, Limb, NonZero, Rem, RemAssign,
-    RemLimb, RemMixed, ToUnsigned, Uint, UintRef, Unsigned, Wrapping,
+    CheckedDiv, Choice, CtOption, Div, DivAssign, DivRemLimb, DivVartime, Limb, NonZero, Rem,
+    RemAssign, RemLimb, RemMixed, ToUnsigned, Uint, UintRef, Unsigned, Wrapping,
 };
 
 impl<const LIMBS: usize> Uint<LIMBS> {
@@ -169,11 +169,14 @@ impl<const LIMBS: usize> Uint<LIMBS> {
     }
 
     /// Computes `(lo + hi * 2^Self::BITS) / rhs` for a double-width dividend, returning the
-    /// wrapped quotient and the remainder.
+    /// wrapped quotient, the remainder, and a [`Choice`] that is truthy when the quotient fit in
+    /// `Self` without truncation.
     ///
     /// The quotient of such a dividend may exceed `Self::BITS`; only its low `Self::BITS` bits are
-    /// returned (i.e. the quotient is reduced modulo `2^Self::BITS`). This is the quotient-tracking
-    /// counterpart of [`Uint::rem_wide`], and avoids widening the operands via
+    /// returned (i.e. the quotient is reduced modulo `2^Self::BITS`), which is why the name is
+    /// prefixed with `wrapping`. The returned [`Choice`] is truthy exactly when no wrapping
+    /// occurred, i.e. when the high half of the dividend is less than `rhs`. This is the
+    /// quotient-tracking counterpart of [`Uint::rem_wide`], and avoids widening the operands via
     /// [`Concat`][`crate::Concat`], so it is available for any limb count.
     ///
     /// ### Usage:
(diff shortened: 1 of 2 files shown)


src/uint/ref_type/div.rs
```diff
@@ -204,6 +204,352 @@ impl UintRef {
...
+    /// be at least as large as `rhs`. `x_lower_upper` is left in an indeterminate state.
+    ///
+    /// The true quotient may be up to twice the width of `rhs`; only its low `quo.nlimbs()` limbs
+    /// are retained (i.e. the quotient is reduced modulo `2^(quo.nlimbs() * Limb::BITS)`). This is
+    /// the quotient-tracking counterpart of [`UintRef::rem_wide`].
+    ///
+    /// # Panics
+    /// If the divisor is zero.
+    #[inline(always)]
+    pub(crate) const fn div_rem_wide(
```
andrewwhitehead: It looks like this should be named `wrapping_div_rem_wide` for instance to make it clear that the quotient can be truncated. Can we maybe detect truncation and return a Choice indicating whether it's exact?

src/uint/ref_type/div.rs
```diff
@@ -204,6 +204,352 @@ impl UintRef {
...
+    }
+
+    /// Conditionally shift the limbs one position toward the most-significant end, dropping the
+    /// top limb and inserting `limb` at the least-significant position. A no-op when `shift` is
+    /// falsy.
+    ///
+    /// Used as a quotient accumulator: feeding quotient limbs most-significant first retains the
+    /// low `self.nlimbs()` limbs of the full-width quotient.
+    #[inline(always)]
+    const fn shift_in_limb(&mut self, limb: Limb, shift: Choice) {
```
andrewwhitehead, line 394: This seems to be duplicating `conditional_shl_assign_by_limbs_vartime`?

src/uint/div.rs
```diff
@@ -998,4 +1122,140 @@ mod tests {
...
+    fn div_rem_wide_vs_concat() {
+        let mut rng = ChaCha8Rng::from_seed([9u8; 32]);
+        for _ in 0..300 {
+            // Single-word path (U64 operands).
+            let lo64 = U64::random_from_rng(&mut rng);
+            let hi64 = U64::random_from_rng(&mut rng);
+            check::<{ U64::LIMBS }, { U128::LIMBS }>(
+                lo64,
+                hi64,
+                nz(U64::random_from_rng(&mut rng)),
```
andrewwhitehead: I think this should be able to use `NonZero::<U64>::random_from_rng` making the `nz` method unnecessary

src/uint/div.rs
```diff
@@ -998,4 +1122,140 @@ mod tests {
         };
         assert_eq!(a.divide_x_by_y(), U1024::from(2707385u64));
     }
+
+    /// Check the wide-division methods (constant-time and variable-time) for a dividend
+    /// `lo + hi * 2^(L * Limb::BITS)` against the trusted `div_rem` reference, which divides the
+    /// same value widened into `Uint<W>` (with `W == 2 * L`).
+    fn check<const L: usize, const W: usize>(lo: Uint<L>, hi: Uint<L>, den: Uint<L>) {
```
andrewwhitehead: This should probably be named `check_wide_division` or something less ambiguous

src/uint/div.rs
```diff
@@ -168,6 +168,126 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+    /// let quo = U256::div_wide_exact((U256::from(15u64), U256::ZERO), &rhs).unwrap();
+    /// assert_eq!(quo, U256::from(5u64));
+    ///
+    /// // 16 is not divisible by 3
+    /// let not_exact = U256::div_wide_exact((U256::from(16u64), U256::ZERO), &rhs);
+    /// assert!(bool::from(not_exact.is_none()));
+    /// ```
+    #[inline]
+    #[must_use]
+    pub const fn div_wide_exact(lower_upper: (Self, Self), rhs: &NonZero<Self>) -> CtOption<Self> {
```
andrewwhitehead, line 255: The nice thing about `div_exact` is that it is much faster than regular division, although this implementation could be switched later. I'm not sure I understand what is guaranteeing that the quotient actually fits since wrapping doesn't seem to be flagged?

src/uint/div.rs
```diff
@@ -168,6 +168,126 @@ impl<const LIMBS: usize> Uint<LIMBS> {
...
+    /// let rhs = NonZero::new(U256::from(3u64)).unwrap();
+    /// let (quo, rem) = U256::div_rem_wide((lo, hi), &rhs);
+    ///
+    /// // the true quotient 2^256 + 1 doesn't fit in 256 bits, so it wraps down to 1
+    /// assert_eq!(quo, U256::ONE);
+    /// assert_eq!(rem, U256::from(2u64));
+    /// ```
+    #[inline]
+    #[must_use]
+    pub const fn div_rem_wide(lower_upper: (Self, Self), rhs: &NonZero<Self>) -> (Self, Self) {
```
andrewwhitehead: I think this would also need to be named `wrapping_div_rem_wide` or similar

cong-or:
@andrewwhitehead  Thanks for the review! Pushed 0ac5d1c with everything addressed.

Names. Renamed div_rem_wide → wrapping_div_rem_wide (and the vartime version) to make it clear the quotient may be truncated. They now also return a fits flag. The check is just hi < rhs, so it's simple, constant-time, and done before any state is modified.

div_wide_exact. Good catch. A zero remainder wasn't sufficient: the quotient can still overflow (e.g. 2^128 / 1), and the old code returned the wrong value in that case. It now only returns Some when the remainder is zero and the quotient fits. I also fixed the docs and added a regression test. I left it built on the full division for now, since the faster path can be swapped in later.

shift_in_limb. Reworked it to call conditional_shl_assign_by_limbs_vartime and then insert the new limb, so the shifting logic isn't duplicated.

Tests. Renamed check → check_wide_division, dropped the nz helper in favor of NonZero::random_from_rng, and updated the checker to verify the new fits flag.

One question: I currently have wrapping_div_rem_wide returning (quo, rem, fits). If you'd rather keep the API smaller, I'm happy to leave it as (quo, rem) and only expose the flag through div_wide_exact.

PR #1331, opened 2026-08-04
CI: simplify i686 ("x86") apt dependencies

tarcieri:
This attempts to reduce the total number of dependencies we're using to do i686 builds from the whole of `gcc-multilib` to just what's actually needed for rustc to work.


tarcieri:
Closing in favor of https://github.com/RustCrypto/actions/pull/66

PR #1338, opened 2026-09-01
impl Zeroize for BoxedMontyParams

baloo:
The intent is to add support for zeroize on RSA's precomputed p_params and q_params.

BoxedMontyParams got placed in an Arc in https://github.com/RustCrypto/crypto-bigint/pull/1143 but I'm not sure I follow the reason why.

To implement Zeroize, this changes remove the Arc. The content is still placed on the heap (There is an underlying BoxedUint).

### DIFF
--- a/src/modular/monty_params.rs
+++ b/src/modular/monty_params.rs
@@ -216,13 +216,15 @@ impl<const LIMBS: usize> FixedMontyParams<LIMBS> {
 pub(crate) mod boxed {
     use super::MontyParams;
     use crate::{Limb, Odd, U64, Word};
-    use alloc::sync::Arc;
     use core::fmt::{self, Debug};
+
+    #[cfg(feature = "zeroize")]
+    use zeroize::Zeroize;
 
     /// Parameters to efficiently go to/from the Montgomery form for an odd modulus whose size and value
     /// are both chosen at runtime.
     #[derive(Clone, Eq, PartialEq)]
-    pub struct BoxedMontyParams(Arc<MontyParams<crate::uint::boxed::BoxedUint>>);
+    pub struct BoxedMontyParams(MontyParams<crate::uint::boxed::BoxedUint>);
 
     impl BoxedMontyParams {
         /// Instantiates a new set of [`BoxedMontyParams`] representing the given `modulus`.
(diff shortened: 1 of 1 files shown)


baloo:
Let's make that a draft until we can figure out the reasoning behind Arc.

tarcieri:
`Arc` prevents the params from being copied every single time a new value is made, incurring a heap allocation for each one. It would incur copies and significantly increase the amount of memory required.

That said we do generally try to make use of in-place operations to avoid heap allocations whenever possible.

See also previous discussion of this in #708

tarcieri:
A potential alternative here could be to expose the equivalent of [`Arc::into_inner`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_inner) and/or [`Arc::get_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut) for the params, namely `BoxedMontyParams` is a wrapper for the generic `MontyParams`.

Someone wanting to write a `Zeroize` impl as in RSA could do e.g. `BoxedMontyParams::get_mut` and obtain an `Option<&mut MontyParams<_>>`, and if they have exclusive access can now use the `Zeroize` impl on `MontyParams`.

src/modular/monty_params.rs
```diff
@@ -216,13 +216,15 @@ impl<const LIMBS: usize> FixedMontyParams<LIMBS> {
...
     use core::fmt::{self, Debug};
 
+    #[cfg(feature = "zeroize")]
+    use zeroize::Zeroize;
+
     /// Parameters to efficiently go to/from the Montgomery form for an odd modulus whose size and value
     /// are both chosen at runtime.
     #[derive(Clone, Eq, PartialEq)]
-    pub struct BoxedMontyParams(Arc<MontyParams<crate::uint::boxed::BoxedUint>>);
+    pub struct BoxedMontyParams(MontyParams<crate::uint::boxed::BoxedUint>);
```
tarcieri, line 227: This defeats the point of having a newtype at all. If we're not using the `Arc` then this should just be:


```suggestion
    pub type BoxedMontyParams = MontyParams<crate::uint::boxed::BoxedUint>;
```

PR #1342, opened 2026-09-10
Implement `Zeroize` for `EncodedUint`

jplatte:
Noticed that this was missing when trying to upgrade crypto-bigint in [aws-sigv4](https://github.com/smithy-lang/smithy-rs/tree/main/aws/rust-runtime/aws-sigv4).


tarcieri (approved):

tarcieri:
Note: I've logged the unrelated proptest failure here: #1343

jplatte:
Thanks for the quick merge! Is there going to be a patch release soon / would it help for me to do a release prep PR collecting changes since the last release into the changelog and bumping the version?

tarcieri:
#1343 seems like an important enough issue it would be good to get resolved before the next release, and also "release prep" PRs by third parties are generally unhelpful for us and often create more work than just letting us do the release ourselves (e.g. now we have to review that PR and decide if it's good enough and if it's not generally close it and do our own release PR, which wastes more time than just letting us do the latter)

jplatte:
Okay, I will just wait then. Thanks again!

Issue #1343, opened 2026-09-11
`pow` proptest failure


tarcieri:
This occurred in CI: https://github.com/RustCrypto/crypto-bigint/actions/runs/34535751379/job/103067274916?pr=1342

```
test pow ... FAILED
failures:

---- pow stdout ----
proptest: FileFailurePersistence::SourceParallel set, but failed to find lib.rs or main.rs

thread 'pow' (1392) panicked at src/modular/reduction.rs:107:5:
assertion failed: Uint::lt(montgomery_form, modulus.as_ref()).to_bool_vartime()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

[...]

proptest: FileFailurePersistence::SourceParallel set, but failed to find lib.rs or main.rs
proptest: Saving this and future failures in /project/tests/fixed_monty_form.proptest-regressions
proptest: If this test was run on a CI system, you may wish to add the following line to your copy of the file. (You may need to create it.)
cc a47746340844e2269f95d60a7e8976fdc8d3df792f6183fbb263e02ecec70d88

thread 'pow' (1392) panicked at tests/fixed_monty_form.rs:161:1:
Test failed: assertion failed: Uint::lt(montgomery_form, modulus.as_ref()).to_bool_vartime().
minimal failing input: x = Uint(0x90F28438C0B05658F62AFDD85E7F4D88E43399F6F79C44D007B3366B20D456B0), y = Uint(0x23FBE7F0BCC28955C6BDF2A16FEAE4BF55C6B90C67715E6FE1E2E18DAB23EC60), n = MontyParams {
    modulus: Odd(
        Uint(0x72419BE35D7D03FB5B969236F15E2AA93AE856FABA8273EAB1AA1D4614367739),
    ),
    one: Uint(0x1B7CC8394505F80948D2DB921D43AAAD8A2F520A8AFB182A9CABC573D793118E),
    r2: Uint(0x4F88F0B21B3DBC04603366D66B072895C349364CE3A7C36879ADCC2CC4767548),
    mod_inv: Uint(0xE940554B0A2D4709),
    mod_leading_zeros: 1,
}
	successes: 71
	local rejects: 0
	global rejects: 0



failures:
    pow

test result: FAILED. 14 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.51s
```

cc @andrewwhitehead

andrewwhitehead:
[...]

PR #1344, opened 2026-09-11
[WIP] FixedMontyForm::pow(_vartime) regression test for #1343

tarcieri:


tarcieri:
@andrewwhitehead would you mind opening a separate PR with the fix? Feel free to reuse my test if you'd like

PR #1346, opened 2026-09-19
safegcd: compute the iteration count in u64

dorn:
**At a glance.** The fix is this:

```diff
 const fn iterations(bits: u32) -> u32 {
-    (45907 * bits + 30179) / 19929
+    let iterations = (45907 * bits as u64 + 30179) / 19929;
+    assert!(iterations <= u32::MAX as u64, "precision too large for safegcd");
+    iterations as u32
 }
```

The rest of the diff is tests and the changelog entry.

**Why.** `iterations` evaluated `(45907 * bits + 30179) / 19929` in `u32`. The sum exceeds `u32::MAX` from `bits = 93558`; a release build wraps it silently, so at 98304 bits (1536 limbs) the reduction loop runs 10,934 divsteps instead of 226,447. The reduction never finishes, and:

- `BoxedUint::invert_mod` / `invert_odd_mod` return `None` for values that do have an inverse (the final `f` is not 1);
- `gcd` / `gcd_odd` return a partially reduced, wrong value.

The count itself (about 2.3 x bits) only exceeds `u32::MAX` above 1,864,517,463 bits of precision, a 233 MB operand; that case now panics instead of wrapping.

**How we hit it.** `rsa` 0.10.0-rc builds the Euler totient with `concatenating_mul`, giving a 65536-bit key a 98304-bit totient container; `e.invert_mod(&totient)` spuriously failed and the key-generation loop retried forever. The failure mode is a spurious `None` / wrong gcd, never a wrong inverse, so no security impact.

**Tests.** Exact iteration counts on both sides of the overflow point and at the largest precision that fits, plus the panic beyond it; inversion (odd and even modulus) and gcd at 93,568 bits, the smallest limb-aligned precision affected, verified against `num-bigint`. All large-precision assertions fail on the previous arithmetic; they add ~0.45 s to the suite and need `alloc`.


andrewwhitehead:
I think it's sufficient to compute it as u64 and then reduce it after dividing:

```
const fn iterations(bits: u32) -> u32 {
    ((45907 * bits as u64 + 30179) / 19929) as u32
}
```

With an allow for the truncation. Not sure if there's a const checked conversion from u64 to u32?

dorn:
Done in 6c615d1: `u64` intermediate, narrowed back to `u32` with a const `assert!` -- there is no const `TryFrom` on stable, but a `const fn` can check by hand -- so the loops keep their `u32` counters. The assertion only fires above 1,864,517,463 bits of precision (a 233 MB operand), where the old arithmetic wrapped; `iterations_panics_beyond_u32` covers it. The diff is now the function body plus tests, and the description is updated to match.

dorn:
Thanks — done in 6c615d1: u64 intermediate, narrowed back to u32 behind a
const assert!
(no const TryFrom on stable, but a const fn can check by hand).
Fires only above ~1.86e9 bits of precision, where it used to wrap.

-dorn

andrewwhitehead (approved):

andrewwhitehead:
It sounds like for efficiency you might also be able to use checked_mul or a checked resize for the totient. Unfortunately I think `resize_checked` is only defined for Uint right now.

dorn:
Yes, that is the other half of it.

On our side rsa's compute_private_exponent_euler_totient now starts the
product from p - 1 instead of one_with_precision,
which drops the container to 2x the prime precision (65536 bits for
RSA-65536) with the same concatenating_mul.

Happy to send that to RustCrypto/RSA as its own PR if you'd like it there.
Just let me know.

Regards,

-dorn

